PR #1, opened 2021-04-15
Add Docker configuration and document steps to get started [DS-99]

danbentley:
I've included basic docker configuration, tool to handle dependencies (Poetry) and a barebones Wagtail project.

There's a lot here, and a lot of that needs documenting (`.env` files, Poetry, directory structure etc).  I feel it's worth just documenting the steps to get the project running for now. Following the steps in the README will build and run an empty Wagtail instance.

Let me know if the steps make sense, whether there's anything you'd like me to elaborate on or remove.


README.md
```diff
@@ -0,0 +1,72 @@
...
+on both the `db` and `elasticsearch` containers in order for the `web`
+container to communicate with those services.
+
+### Getting started
+
+#### Docker
+
+##### Build containers
+
+`docker-compose build`
```
Rijidij, line 50: A `.env` file is required for this to work.
Can you provide an example file?

danbentley: Good spot!

I hadn't pushed that last commit (a743a76)

Rijidij (approved):

AshTNA (changes requested):
Thanks @danbentley - these are very clear and they worked first time, and definitely match the level we like in the team!

One thing I'd maybe add is to mention that `docker-compose up` takes over your terminal instance, so maybe it should be `docker-compose up -d` to detach it from the current terminal instance. If not, instead a step after that could be added to instruct the user to open a new terminal. I'm mindful of the fact that some less experienced devs with docker/servers may think that their terminal has frozen on this step when in fact it's just the server log, waiting for requests!

Another suggestion would be maybe to add next to the admin dashboard link that the username to login is `sysadmin` and the password is the one you set in the final command. I'm thinking that some may assume that their email address they entered when making a superuser is the username and not `sysadmin`. 

I'm happy to make these changes if you're busy. Thanks :+1:

danbentley:
Thanks Ash. I'm glad it worked first time!

Excellent points!

I run docker containers in the foreground with streams of text scrolling in my terminal but I know most people don't (and for good reason!)

I've updated the command with the `--detach` flag.

Also, great suggestion on documenting which credentials should be used to log in. 

Technically speaking, both the user name and email address are accepted by the login form. Though logging in with "`sysadmin`" is a good habit to develop, especially once we have content on staging/prod.

AshTNA (approved):
Thanks @danbentley!

gtvj (approved):
Thanks @danbentley - the instructions were really clear and it all worked perfectly for me too. I'll merge this now 👍

PR #3, opened 2021-04-22
DS 141

gtvj:
Hi @danbentley, 

This PR:

* Introduces SASS
* Introduces a layout and partials for the header and footer

I hope it all seems OK from a Wagtail perspective but let us know if anything might benefit from being changed. 

Thanks 👍


danbentley:
Thanks Ash, this looks great!

Do you guys use [`nvm`](https://github.com/nvm-sh/nvm) or other tools to manage node versions?

In the past, we've had issues where projects no longer build because the project requires an older version of node. This can especially be apparent when pushing code to production/staging.

To help resolve such issues, our projects now contain an [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) file containing the supported version of node used to build the FE.

gtvj:
Hi @danbentley - sorry, I've only just seen your comment. We haven't used NVM before but I'd be happy to adopt it if it will help us avoid similar problems. I'll spend some time today getting familiar with it.

PR #7, opened 2021-06-08
Accessibility improvements

gtvj:
This is to make all of the improvements identified in the Accessibility Testing Report for Explorer (see: https://docs.google.com/document/d/1RxdxDNf0UMLP-AsxudWY7PzeHwZV7ktwj8fNKkh9rF8/edit). These are: 

- [x] Ensure all images are visible when CSS is not available (and update embedded Font Awesome licence information)
- [x] Ensure all images are displayed at a reasonable size when CSS is not available (some of the SVGs are _huge_)
- [x] Add `sr-only` span with "Current page" to pagination so that the current page can be more easily determined when CSS is not available
- [x] Address "Graphic" issue affecting NVDA users when interacting with image cards
- [x] Make time period date presentation to screenreaders meaningful (they're currently read out as _"One Thousand Four Hundred And Eighty Five One Thousand Seven Hundred and Fourteen"_
- [x] Fix issue with `aria-live` not firing for the sorting selector in NVDA
- [x] Ensure image alt text is below 100 characters (and update Functional and Technical specification to reflect this)
- [x] ~~Fix alt text for logo (note what GDS and accessibility specialists, such as WebAIM do)~~ (See comment below)
- [x] ~~Update page titles to reflect editorial standards (following current approach `[page name] - The National Archives`)~~ (See comment below)


gtvj:
Hi @AshTNA - I've been looking at the logo issue and I think the current `alt` text is OK. I'm basing this on the approach used by: 

* [WCAG](https://www.w3.org/TR/WCAG/) who have the W3C logo alt text as `alt="W3C"`
* [WebAIM](https://webaim.org) who have the logo alt text as `alt="WebAIM - Web Accessibility In Mind"`
* [Paciello Group](https://www.tpgi.com) have the logo alt text as `alt="TPGi"`

I'm therefore going to remove the related task from this PR, if that's OK.

AshTNA:
Thanks @gtvj - sounds good.

gtvj:
I've realised the issue with titles is due to a Wagtail setting on development machines. In order for the site name to be displayed in the title it needs to be set (which it will be in production). Otherwise it's just the page name which is displayed. 

You can see this in `templates/base.html`.

I'm therefore going to remove that task too.

gtvj:
I've just checked the technical and functional specification document and all references to alt text specify a maximum of 100 characters.

gtvj:
Hi @AshTNA - I think the one remaining issue might have been addressed by the restructuring of the cards I've done. Could you just check that's the case and, if so, mark the issue as done?

AshTNA:
Thanks @gtvj - it sounds good on NVDA now. It looks like the SVGs still may not have the correct licenses, so I can go in and fix those.

AshTNA (approved):
All good :+1:

PR #12, opened 2021-06-28
Introduce alert-message component

AshTNA:
Hi @gtvj 

Here is the PR for the alert message component, relating to https://national-archives.atlassian.net/browse/DS-233 

image

I have checked it on:

- Chrome, MacOS X
- Firefox, Windows 10
- IE11, Windows 10

A couple questions:

- Do these need a `role="alert"`? VoiceOver seemed to read them out first anyway, since they're the first elements on the page.
- Should I ask a designer to look over this before merging?

Thanks :+1:

### DIFF
--- a/sass/etna.scss
+++ b/sass/etna.scss
@@ -12,3 +12,4 @@
 @import 'includes/buttons';
 @import 'includes/results-options';
 @import 'includes/pagination';
+@import 'includes/alert-message';
--- a/sass/includes/_alert-message.scss
+++ b/sass/includes/_alert-message.scss
@@ -1,6 +1,6 @@
 .alert-message {
 
-    padding: 0.5rem;
+    padding: 0.5em;
 
     &--high {
         @extend .alert-message;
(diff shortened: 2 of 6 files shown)


gtvj:
Hi @AshTNA - I don't think it does need the `alert` role because that's for dynamically updated content (and I believe the intention is that these will be included in the initial HTML. 

What is the IE bug with `rem`? We're using the `rem` measurement throughout Etna?

AshTNA:
Thanks @gtvj - I can't seem to replicate the issue on IE11 anymore. When I was loading  the  page in IE11, the `rem` padding was not being  applied, even though it existed in DevTools. It only kicked in when I unchecked it, and checked it again in DevTools. I switched  it to `em` and then the issue stopped happening. I switched it back to `rem` just now to try and surface the bug again, but it's not having any problems with `rem` anymore.

gtvj:
Thanks Ash - if we're happy with `rem` could you revert it? I'd really like to have a single reference point for all relative measurements.

AshTNA:
Thanks @gtvj - I've reverted it.

gtvj (approved):

gtvj:
Thanks @AshTNA - merging now.

PR #13, opened 2021-06-29
DS 232 Frontend

AshTNA:
Hi @gtvj,

Would you be able to look at this PR?

It adds some HTML includes/static content to the homepage to match the UXPin Mockup in 
https://national-archives.atlassian.net/browse/DS-232

I have also changed the  card loops to use `page.get_children.specific`  instead of `page.get_children`, as Dan B let me know that without the `.specific`, you can only access fields on the default Wagtail Page class, instead of the extra fields that have been added to the Page via  the `models.py` file. So in this case,  I have made it so that `{{child.introduction}}` is accessible in the cards, so that their description texts are dynamic.

Here's a screenshot of the homepage with the same  cards as the UXPin:

image

Note that in the UXPin, the cards have a "NEW" label next  to them, however we decided this was a nice to  have, so this doesn't need to be accounted for.

Thanks. :+1:

### DIFF
--- a/etna/home/templates/home/home_page.html
+++ b/etna/home/templates/home/home_page.html
@@ -9,6 +9,38 @@
 
 {% block content %}
 
-<h1>{{ self.title }}</h1>
+{% include 'includes/generic-intro--dark.html' %}
+
+<div class="container pt-4">
+    <h2 class="sr-only">A subheading for the cards (for accessibility purposes)</h2>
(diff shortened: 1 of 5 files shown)


gtvj (approved):
Thanks Ash - looks good. Thanks for letting me know about `page.get_children.specific` too. I didn't know about that.

PR #14, opened 2021-07-02
Output explorer page content

danbentley:
I've renamed the page models for the explorer page to better convey their purpose. Also, I've created a second explorer page for time periods.

Additionally, there are new `introduction` and `teaser_image` fields for explorer pages.


AshTNA:
Thanks @danbentley, looks good. I didn't realise fields could be added to the promote tab!

I can see that the new card loop implementation breaks the cards that currently appear on the homepage. I'm guessing a function would need to be written in the future which grabbed the Explorer Index Page, Insights Index Page, and Details Index Page. 

Just so that visually the homepage makes sense for whoever implements it, do you think it's worth hardcoding an object that the homepage cards can loop through, just so they don't look like this?

image

Thanks :+1:

danbentley:
Thanks Ash! 

Yeah, adding fields to the promote tab is really useful. It's also possible to create custom tabs for pages too!

https://docs.wagtail.io/en/stable/advanced_topics/customisation/page_editing_interface.html#id1

Apologies for breaking the links on the homepage! Yeah, I think it's worth hardcoding the teasers for now. It might be a while before we have the relevant models in place to dynamically populate the home page.

AshTNA:
Thanks @danbentley - I've pushed a commit which hardcodes these homepage cards. Feel free to revert the commit or implement a better version yourself if needed. :+1:

PR #15, opened 2021-07-02
Integrate: Alerts

Rijidij:
Here is an example of a page displaying multiple alerts:
Screenshot from 2021-07-02 11-13-49


etna/alerts/templatetags/alert_tags.py
```diff
@@ -0,0 +1,51 @@
...
+    except AttributeError:
+        pass
+
+    if alert and alert.active:
+        return alert
+    else:
+        return None
+
+
+@register.inclusion_tag('alerts/tags/alerts.html', takes_context=False)
```
danbentley, line 28: `takes_context` defaults to `None` so you should be able to remove this.

etna/alerts/templatetags/alert_tags.py
```diff
@@ -0,0 +1,51 @@
...
+    # Get alert for current page, if any.
+    alert = get_page_alert(page)
+    if alert:
+        alerts.insert(0, alert)
+
+    # Get current page ancestors.
+    branch = Page.objects.ancestor_of(page).specific()
+    # Iterate backwards and, if page has alert and "cascade" is true,
+    # add to alerts list.
+    for i in reversed(range(branch.count())):
```
danbentley, line 44: A more idiomatic way to loop through the pages in reverse order is:

```
for page in Page.objects.ancestor_of(page).specific().reverse():
    …
```

Is there a reason why we need to gather the alerts in reverse order but insert them at the front of the list?

etna/home/models.py
```diff
@@ -1,7 +1,18 @@
 from django.db import models
 
 from wagtail.core.models import Page
+from wagtail.snippets.edit_handlers import SnippetChooserPanel
 
 
 class HomePage(Page):
-    pass
+    alert = models.ForeignKey(
```
danbentley, line 8: I can see the need to add the alerts to a lot of pages. Have you considered using a mixin?

etna/text_formats/blocks.py
```diff
@@ -0,0 +1,13 @@
+from wagtail.core import blocks
```
danbentley, line 1: I really like the idea of using a separate app for rich text fields, it helps reduce the need for the dreaded `base` app too!

I think the name `text_formats` might be a bit restrictive though. In the past, we've written rich text converters, editor features as well as supporting [legacy rich text](https://docs.wagtail.io/en/v2.12.4/reference/contrib/legacy_richtext.html) and it'd be nice to keep this all in one place.

Wagtail is a bit inconsistent with its naming here, using both [`rich_text`](https://github.com/wagtail/wagtail/tree/main/wagtail/admin/rich_text ) and [`richtext`](https://github.com/wagtail/wagtail/tree/main/wagtail/contrib/legacy/richtext) but something along those lines would be good.

danbentley (changes requested):

sass/includes/_alert-message.scss
```diff
@@ -25,13 +25,19 @@
...
     &__description {
         margin-bottom: 0;
         display: inline-block;
     }
 
+    &__description a {
+        color: $color__link-hover;
+    }
+
     &__link {
```
AshTNA, line 41: Would you be able to remove this `&__link` class, as it is not used if the only way to access the links  is via the `<a>` tag itself

sass/includes/_alert-message.scss
```diff
@@ -25,13 +25,19 @@
...
+        font-weight: bold;
+        text-transform: uppercase;
     }
 
     &__description {
         margin-bottom: 0;
         display: inline-block;
     }
 
+    &__description a {
```
AshTNA, line 37: Would you be able to remove this, as the styling  for the `<a>` should be nested inside main `&__description` class.

sass/includes/_alert-message.scss
```diff
@@ -25,13 +25,19 @@
         padding-left: 0.4rem;
         padding-right: 0.4rem;
         margin-right: 0.4rem;
+        font-weight: bold;
+        text-transform: uppercase;
     }
 
     &__description {
```
AshTNA, line 32: Could you please have the &__description class as so:

```SASS
    &__description {
        margin-bottom: 0;
        display: inline-block;
        a {
            color: $color__link-hover;
        }
        p {
            margin-bottom: 0;
        }
    }
```

This nests the existing `<a>` class you added within the parent `&__description` block, and also adds a `margin-bottom: 0`  to the `<p>` which we weren't expecting Wagtail to render in our static version.

AshTNA (changes requested):
Hi @Rijidij,

Would you be able make some changes to the SASS? This makes it a bit cleaner and accounts for the `<p>` tag we weren't expecting to be rendered.

Thanks :+1:

sass/includes/_alert-message.scss
Rijidij: No worries.

etna/text_formats/blocks.py
Rijidij: I've renamed the app `richtexts`, to contain any variations that may be required.

AshTNA (approved):
All good on the frontend :+1:

etna/alerts/templatetags/alert_tags.py
Rijidij: This code has been simplified. The list reversal is no longer required to return pages in hierarchical order.

etna/home/models.py
Rijidij: A mixin has been created for the alert field and added to the page models.

etna/text_formats/blocks.py
Rijidij: I've removed the richtext(s) app as it's not needed using the features defined in settings as we discussed.
I might need to add it back in later for blocks requiring different feature sets. But I will define those sets in settings as well.

danbentley:
Approved. Excellent work!

danbentley (approved):

danbentley:
I've rebased this work onto `main` and removed some commits that were causing conflicts but were later reverted in the timeline. 

Additionally, I've consolidated the migrations, manually added the `AlertMixin` to explorer pages (there were too many changes in `collections/models.py` to do it any other way), moved alerts to the settings tab and fixed the unit test.

@Rijidij would you mind checking out the branch and giving it a quick test?

etna/collections/models.py
```diff
@@ -155,6 +159,7 @@ class ResultsPage(TeaserImageMixin, Page):
         FieldPanel("introduction"),
     ]
     promote_panels = Page.promote_panels + TeaserImageMixin.promote_panels
+    settings_panels = Page.settings_panels + AlertMixin.settings_panels
 
     content_panels = [
```
Rijidij, line 164: This is the issue I found from the first merge.
content_panels is previously assigned. Line 154/158

Rijidij: Otherwise, all OK.

gtvj (approved):

etna/collections/models.py
danbentley: Thanks for checking.

I noticed that too. It's not from a merge but something that was introduced by mistake when I introduced the CIIM client work.

I've fixed in my explorer integration branch.

danbentley (approved):

PR #16, opened 2021-07-02
Introduce OpenSeadragon and JavaScript asset bundling

gtvj:
This PR integrates the image viewer, image browse introduces a JavaScript build system (using Webpack and Babel).

A few things that might be helpful to know are: 

* I've introduced a new `image-base.html` layout because these pages are a little different to others (in that they don't share a lot of the same furniture). I'll keep an eye on this but for the time being I'd rather have two layouts than one with a lot of conditions etc. in it. 
* In the new layout I've put placeholder text in the `<title>` tag but this should be the same as what's in the `<h1>`
* Within `image-browse.html` I've included static text (lines 17-166) that would be replaced with a for loop - for some reason I couldn't get Django to run a loop over a list (I think this might be because I was using a literal list in the template)
* `image-viewer.html` has a `<noscript>` tag in it with an image which will be presented if for any reason JavaScript isn't available. This should be the same image as is presented via the image viewer. 
* Within `image-viewer.html` I've included a JavaScript variable that has the URL for the image to be presented. This is referenced in `image-viewer.js` for initialising OpenSeadragon. This doesn't necessarily need to be provided to the UI as a variable within a script block. I could pick it up from a `<meta>` tag or from a `data-` attribute on an element - whichever is easiest for you
* I've updated the README to reflect the JavaScript build system - both for local development (which uses a watch task) and for the server (which is an `npx` command).

I think that's mostly it but let me know if anything might need to be changed or amended - or if you've got any suggestions.


danbentley (approved):
Apologies Gywn, I didn't realise this was assigned to me!

This looks good, especially the new base template to keep the main template clean!

PR #17, opened 2021-07-05
Bring TNA toolkit into the Wagtail repository

gtvj:
This decision follows a discussion between JC, MB, AB and GJ earliertoday during which we agreed that doing so will make it easier to both maintain Etna and identify candidate changes to the TNA Tookit that emerge from our work on Etna.


gtvj:
Hi @AshTNA @JamesWChan @matt-blair - this is the PR we discussed earlier. I've brought the Toolkit styles (and Bootstrap dependencies) into this repository and put a big comment into the `etna.scss` file.

AshTNA (changes requested):
image

Hi @gtvj , I've noticed the `tna_card` from the toolkit you have imported is different from the one we had in `toolkit_duplicates` previously. Therefore the results cards have broken. See this screenshot. Would you be able to fix/investigate this? If you are busy would you like me to look into this? 

See here: https://github.com/nationalarchives/ds-wagtail/pull/17/files#diff-bc250a9001fcdecae95e1c444916183be1bc47a1444351ff457a56c575c06d22


Thanks :+1:

AshTNA:
Hi Gwyn,

I did some investigation and looks like the `_card.scss` we originally had doesn't actually exist in the toolkit, but was created from the design that Chris created, and therefore was in the `toolkit-duplicates` folder.

This is the card design that Chris made: https://github.com/nationalarchives/tna-frontend-design-toolkit/blob/main/components/cards/example.png

Therefore I think we need to integrate  our original `_card.scss` file with the toolkit one

gtvj:
Hi @AshTNA - thanks for identifying this issue. Would you be able to look into identifying a fix?

AshTNA (approved):

PR #21, opened 2021-07-08
Create the InsightsPage framework

Rijidij:
A skeleton insights page (and index page) that we can add components to as they are developed.


etna/insights/models.py
```diff
@@ -0,0 +1,45 @@
...
+class InsightsIndexPage(Page):
+    """InsightsIndexPage
+
+    This page lists the InsightsPage objects that are children of this page.
+    """
+    introduction = models.CharField(max_length=200, blank=False)
+
+    def get_context(self, request):
+        context = super().get_context(request)
+        insights_pages = self.get_children().live()
```
danbentley, line 20: This looks good. 

It's probably worth excluding private pages:

https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html#wagtail.core.query.PageQuerySet.public

Along with the specific page type instances:

https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html#wagtail.core.query.PageQuerySet.specific

etna/insights/models.py
```diff
@@ -0,0 +1,45 @@
...
+    subpage_types = ['insights.InsightsPage']
+
+
+class InsightsPage(Page):
+    """InsightsPage
+
+    The InsightsPage model.
+    """
+    introduction = models.CharField(max_length=200, blank=False)
+    content = StreamField(InsightsPageStreamBlock, blank=True, null=True)
```
danbentley, line 37: Could we rename this attribute to `body`? This keeps it consistent with our other sites and is becoming a convention:

https://docs.wagtail.io/en/stable/reference/streamfield/blocks.html#streamfield-block-reference

danbentley (changes requested):
Thanks for this. It looks good. I like that you've included a `StreamBlock` to be populated with our other component types. 

I've requested a few minor changes, just to keep things consistent.

templates/insights/insights_page.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
+
+{% block content %}
+    {% include 'includes/generic-intro--dark.html' %}
+
+    <div class="container">
+        {{ page.content }}
```
danbentley, line 7: The template still references `content`

etna/insights/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.1.8 on 2021-07-08 13:26
```
danbentley, line 1: Could you squash these migrations down to a single file:

Roll back the insights migration:

`python manage.py migrate insights zero`

Delete the existing migration files

Recreate:
`python manage.py makemigrations insights`

templates/insights/insights_page.html
Rijidij: Done

etna/insights/migrations/0001_initial.py
Rijidij: and done.

PR #22, opened 2021-07-09
Amend KongClient to use an API key

danbentley:
Requests to Kong now require an API key to allow us access without needing a VPN


Rijidij (approved):
Looks good to me!

PR #23, opened 2021-07-09
Replace settings access with init params in client

danbentley:
Small tidy to the Kong client.

I've replaced code where `KongClient` directly accesses `django.conf.settings` in favour of `__init__` params


Rijidij (approved):
All good.

PR #24, opened 2021-07-12
Create quote streamfield component

Rijidij:
- Add to insights page

[DS-236]

Due to merge complexities, we decided it would be simpler to create a new branch.
This replaces branch DS-236.

[DS-236]: https://national-archives.atlassian.net/browse/DS-236


templates/quotes/blocks/quote.html
```diff
@@ -0,0 +1,9 @@
+{% load wagtailcore_tags %}
```
danbentley, line 1: I could be wrong but I think when outputting `StreamField` content  you only need to load `wagtailcore_tags` when iterating over the data

danbentley (changes requested):
Looks great. Thanks @Rijidij!

templates/quotes/blocks/quote.html
Rijidij: It's needed for `include_block`

danbentley (approved):

PR #25, opened 2021-07-12
Explorer pages backend integration.

danbentley:
I've included a `StreamField` on each explorer page. 

Primarily this was to support featured/promoted pages on topic/time explorer pages but I felt it made sense to add to each content type for editor consistency and flexibility later on.

@gtvj, @AshTNA I'd like to get your thoughts on `templates/collections/blocks/promoted_pages.html`. 

Ideally, the template should include `templates/includes/card-group-secondary-nav.html` but because the source data is coming from a [`wagtail.core.blocks.StructBlock`](https://docs.wagtail.io/en/stable/reference/streamfield/blocks.html#wagtail.core.blocks.StructBlock) and not a [`Page`](https://docs.wagtail.io/en/stable/topics/pages.html#page-models) instance, I'm unable to include the template.


Rijidij (approved):
Looks good!

templates/collections/blocks/promoted_pages.html
```diff
@@ -0,0 +1,44 @@
...
+
+        {% comment %}
+        Source copied from templates/includes/card-group-secondary-nav.html
+
+        Modified slightly to support data from a streamfield.
+
+        This is not intended to be production-ready code.
+
+        To discuss an approach to allow for reusability going forward.
+        {% endcomment %}
```
gtvj, line 22: Hi @danbentley - I think this is fine for the time-being. I'd rather have some pragmatic redundancy than additional complexity and the key thing is that we can determine where any given template is coming from, which we can from the block.

AshTNA:
Hi Dan,

Thanks for working on this, it's nice to learn about Wagtail just from looking at how the pros use it!

Gwyn and I looked at `templates/collections/blocks/promoted_pages.html` together and the approach looks good.

Something we have noticed is that the model structure for the pages in this branch is reflecting the older UI design. On July 5th SW shared a newer UI https://tna-digital.slack.com/archives/C01F9E7F6C9/p1625492968044000 which is also reflected now in `ds-wagtail-frontend` at http://ds-wagtail-frontend.herokuapp.com/ (with slightly different colours due to SW's latest message in `ds-etna-explorer` which had newer colours)

If I were to implement the new HTML into this repo I'm aware that these page models being different may make things confusing, and you'd have to change the models again. Therefore some changes to the backend is probably needed if we incorporate the new UI. What I could do is simply paste the static content into each template file again, regardless of the current model structure, if that is preferable. Or, if you're happy to maybe you could modify the model structure first, and then I could add in the new HTML taking into account the dynamic population of content.

Apologies about the design changing midway through development - I'm aware that it's probably frustrating having these  changes of direction! If you think there's not enough time to change this, feel free to raise this with the relevant people at TNA as I'm mindful that you're probably quite busy trying to work on other parts of the site.

Thanks :+1:

gtvj (approved):
Hi @danbentley - this looks good to me. When I was looking at it with Ash I saw that it reflects the design as currently represented in `ds-wagtail-frontend`. At some point we'll need an update to reflect Simon's changes he shared last week and I'm not sure if that's best to happen in this PR or if a separate one.

danbentley:
Thanks Ash!

> If I were to implement the new HTML into this repo I'm aware that these page models being different may make things confusing, and you'd have to change the models again. Therefore some changes to the backend is probably needed if we incorporate the new UI. What I could do is simply paste the static content into each template file again, regardless of the current model structure, if that is preferable. Or, if you're happy to maybe you could modify the model structure first, and then I could add in the new HTML taking into account the dynamic population of content.

Good question.  Looking at the new UI, I think we need:
 - Amend explorer landing page to show topic, time period teasers
 - New landing page for topic and time period pages
 - Teaser for topic landing page on time period landing page and vice-versa

Because we need a new page type, I think it's worth me making the structural changes with html from the prototypes first.

What do you think?

AshTNA:
Thanks @danbentley - sounds good. Let me know if have any questions with anything :+1:

danbentley:
I've got a couple of amends for the new page structure/UI. The templates almost certainly need work but the basic structure is in place to support:

**Explorer landing page to show topic, time period teasers**

Screenshot 2021-07-13 at 16 03 41

**New landing page for topic and time period pages**

Screenshot 2021-07-13 at 16 08 04

**Teaser for topic landing page on time period landing page and vice-versa**

Screenshot 2021-07-13 at 16 07 35

Would you like me to add the changes to this PR or merge this PR in and create a new one?

AshTNA:
Hi @danbentley, sorry I somehow missed the notification for your comment! Thanks for doing this, looks  good.

Feel free to add the changes wherever you feel is best :+1:

PR #30, opened 2021-07-12
Integrating static HTML and SASS from DS-wagtail-frontend repo

matt-blair:
@AshTNA, @danbentley 

We've migrated the HTML and SASS for details page into Wagtail. I couldn't get it to preview locally using the templates/records/records_page.html template as I couldn't get a URL working, but I've added the HTML into a new template called details-page-front-end.html for the purpose of getting it all in there and viewable. Hopefully we can move stuff from this file into the right place, but for now it's displaying as it should statically.

@danbentley - Is this ok, or would you like this moved somewhere else, or done differently? I've added you as a reviewer so you can comment if you'd like it modified.

@AshTNA - once Dan has confirmed he's happy, are you ok to review it? 

Matt.


danbentley (commented):
Thanks Matt. Looks good! It's great to see the details page making its way into Wagtail!

I've been thinking a lot about `includes` and how [generic/reusable they should be](https://github.com/nationalarchives/ds-wagtail/pull/25/files#diff-c651d865e070f4a8ed6824b8aa20db470e2e6798a40e4cc09b3b7a77c5aef380R15-R21)

The details page appears to be unique in that most (if not all) of its components aren't shared with other page types and in fact, because the templates render data from a `RecordPage`, which itself is a unique model, probably can't be reused easily anyway.

Because of this, I'm wondering whether we should either:

Use a flat `records/details-page-front-end.html`/`records/details_page.html`

Or, if you prefer organising code with `includes`, we could 'namespace' them by putting them in `templates/records/includes/` or `templates/includes/records` to convey that these templates aren't necessarily available for use by any page type.

AshTNA (approved):

AshTNA:
I've approved  the current implementation but happy to look again if anything changes. :+1:

matt-blair:
Thanks @AshTNA and @danbentley. 

@danbentley - I think this sounds very sensible. I'd probably opt for the namespacing approach as it'll keep the high level 'skeleton' of the page clear and the detail of the component parts of the page within their own files. Hopefully makes it easier to read and manage. 

I'll make those changes now in the directory `templates/includes/records` and then re-submit for @AshTNA to look over.

matt-blair:
@AshTNA can you look at the two latest commits, give the thumbs up and merge the pull request if all looks ok?

AshTNA (approved):
All good :+1:

PR #32, opened 2021-07-13
Create hero_image field mixin

Rijidij:
[DS-259]

Add the hero image in the same manner as teaser image.

[DS-259]: https://national-archives.atlassian.net/browse/DS-259


danbentley (approved):
Nice work!

PR #38, opened 2021-07-14
Fix attribution logic in template

Rijidij:
[DS-236]

[DS-236]: https://national-archives.atlassian.net/browse/DS-236


danbentley (approved):
Looks good!

PR #39, opened 2021-07-14
Use hero_image field in template

Rijidij:
[DS-259]

[DS-259]: https://national-archives.atlassian.net/browse/DS-259


danbentley (approved):
Looks good!

PR #41, opened 2021-07-15
Feature/explorer frontend updates

AshTNA:
Hi @gtvj & @danbentley 

Here is the PR for the Explorer frontend updates in Wagtail.

I've had to tweak the SASS/HTML a bit more than expected to make everything work, as the HTML has become out of sync with `ds-wagtail-frontend` (which I'll try and resynchronise at some point).

The changes include:

- An updated header with some descriptive text. The logo has been fixed too (the "BETA" text was serif outside of Chrome)
- Updates to the `card-group-promo` component styles, to allow for the new index page card styles
- Padding fixes on the `card-group-secondary-nav` and `card-group-record-summary` cards
- Adding the amount of items returned on the results page, which also involved some styling changes to the `results_options` component. 

I've hard coded some H2 elements into a couple of the blocks/templates. They're needed for accessibility - they'll probably need to be added to the pages model.

Below I've pasted a screenshot of the `results_options` component on `results_page.html`. The model validation prevents you from creating a `results_page` currently, so I used the `urls.py` to check that page. It's also slightly different from `ds-wagtail-frontend` as in that repo I forgot to account for the non-JS submit button!

#### Results options on Desktop
image
#### Results options on Mobile (320px)
image

Would you be able to approve the changes if all is OK? Thanks :+1:

### DIFF
--- a/sass/includes/_header.scss
+++ b/sass/includes/_header.scss
@@ -1,5 +1,5 @@
 .tna-header {
-  padding: 1rem 3rem 0rem 3rem;
+  padding: 1rem 3rem 0 3rem;
 
   @media only screen and (max-width: $screen__md) {
   padding: 1rem 1rem 1rem 1rem;
@@ -19,7 +19,7 @@
   }
 
   &__description {
-    font-size: 1.2em;
+    font-size: 1.2rem;
     margin-left: auto;
     margin-top: auto;
     margin-bottom: 0;
@@ -37,8 +37,8 @@
 
     &__description {
       padding-top: 0;
-      padding-left: 5em;
-      max-width: 600px;
+      padding-left: 5rem;
+      max-width: 37.5rem;
       text-align: right;
       &-heading {
         text-align: right;
--- a/sass/includes/_header.scss
+++ b/sass/includes/_header.scss
@@ -1,9 +1,48 @@
 .tna-header {
+  padding: 1rem 3rem 0rem 3rem;
+
+  @media only screen and (max-width: $screen__md) {
+  padding: 1rem 1rem 1rem 1rem;
+
+  }
+
+  &__home-link {
+    width: 100%;
+    max-width: 8.938rem;
+  }
 
   &__logo {
 
     &--beta {
       max-width: 8.938rem;
+    }
+  }
+
+  &__description {
+    font-size: 1.2em;
+    margin-left: auto;
+    margin-top: auto;
+    margin-bottom: 0;
+    padding-top: 1rem;
+    
+    &-heading {
+      font-weight: bold;
+    }
+  }
+
+  @media only screen and (min-width: $screen__md) {
+    display: flex;
+    justify-content: space-between;
+    padding-bottom: 1rem;
+
+    &__description {
+      padding-top: 0;
+      padding-left: 5em;
+      max-width: 600px;
+      text-align: right;
+      &-heading {
+        text-align: right;
+      }
     }
   }
 }
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -1,6 +1,6 @@
 .results-options {
   background-color: $color__green;
-  padding: 1em;
+  padding: 1rem;
 
   &__count {
     margin-bottom: 0;
@@ -17,7 +17,7 @@
     border-radius: 0;
 
     @media only screen and (max-width: $screen__sm) {
-      margin-top: 1em;
+      margin-top: 1rem;
     }
 
     &:focus {
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -33,7 +33,7 @@
       outline: 0.313rem solid $color__focus-blue-outline-light-bg;
     }
     padding: 0.5rem;
-    margin-left: 0rem;
+    margin-left: 0;
     @media only screen and (max-width: $screen__md) {
       margin-left: 0.3rem;
     }
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -6,6 +6,10 @@
     margin-bottom: 0;
   }
 
+  &__sort {
+    text-align: right;
+  }
+
   &__select {
     padding: 0.2rem;
     border-radius: 0;
@@ -20,11 +24,15 @@
   }
 
   &__submit {
-    @extend .tna-button;
+    @extend .tna-button--dark;
+    border: none;
     cursor: pointer;
     &:focus {
       outline: 0.313rem solid $color__focus-blue-outline-light-bg;
     }
+    padding: 0.5rem;
+    margin-left: 0rem;
+    margin-top: 0.5rem;
   }
 
   &__loading-message {
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -7,7 +7,7 @@
   }
 
   &__sort {
-    @media only screen and (min-width: 575px) {
+    @media only screen and (min-width: $screen__sm) {
      text-align: right;
     }
   }
@@ -16,7 +16,7 @@
     padding: 0.2rem;
     border-radius: 0;
 
-    @media only screen and (max-width: 575px) {
+    @media only screen and (max-width: $screen__sm) {
       margin-top: 1em;
     }
 
@@ -34,6 +34,9 @@
     }
     padding: 0.5rem;
     margin-left: 0rem;
+    @media only screen and (max-width: $screen__md) {
+      margin-left: 0.3rem;
+    }
     margin-top: 0.5rem;
   }
 
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -7,7 +7,9 @@
   }
 
   &__sort {
-    text-align: right;
+    @media only screen and (min-width: 575px) {
+     text-align: right;
+    }
   }
 
   &__select {
--- a/sass/includes/_results-options.scss
+++ b/sass/includes/_results-options.scss
@@ -1,10 +1,18 @@
 .results-options {
-  text-align: right;
+  background-color: $color__green;
+  padding: 1em;
+
+  &__count {
+    margin-bottom: 0;
+  }
 
   &__select {
     padding: 0.2rem;
     border-radius: 0;
-    margin-left: .3rem;
+
+    @media only screen and (max-width: 575px) {
+      margin-top: 1em;
+    }
 
     &:focus {
       outline: 0.313rem solid $color__focus-blue-outline-light-bg;
--- a/templates/collections/blocks/featured_page.html
+++ b/templates/collections/blocks/featured_page.html
@@ -1,3 +1,5 @@
-<ul class="row card-group">
-    {% include 'includes/card-group-promo--dark.html' with page=value.page.specific heading=value.heading only %}
-</ul>
+<div class="row">
+    <ul class="card-group">
+        {% include 'includes/card-group-promo--dark.html' with page=value.page.specific heading=value.heading only %}
+    </ul>
+</div>
--- a/templates/collections/blocks/featured_page.html
+++ b/templates/collections/blocks/featured_page.html
@@ -1,3 +1,3 @@
-<ul class="card-group">
+<ul class="row card-group">
     {% include 'includes/card-group-promo--dark.html' with page=value.page.specific heading=value.heading only %}
 </ul>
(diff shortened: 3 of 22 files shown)


sass/includes/_header.scss
```diff
@@ -1,9 +1,48 @@
 .tna-header {
+  padding: 1rem 3rem 0rem 3rem;
```
gtvj, line 2: I don't think you need the `rem` unit on the `0` here. `padding: 1rem 3rem 0 3rem` should be fine.

sass/includes/_header.scss
```diff
@@ -1,9 +1,48 @@
...
 
   &__logo {
 
     &--beta {
       max-width: 8.938rem;
     }
   }
+
+  &__description {
+    font-size: 1.2em;
```
gtvj, line 22: Could we change this to use `rem`

sass/includes/_header.scss
```diff
@@ -1,9 +1,48 @@
...
+  }
+
+  @media only screen and (min-width: $screen__md) {
+    display: flex;
+    justify-content: space-between;
+    padding-bottom: 1rem;
+
+    &__description {
+      padding-top: 0;
+      padding-left: 5em;
```
gtvj, line 40: Could we change this to use `rem`?

sass/includes/_header.scss
```diff
@@ -1,9 +1,48 @@
...
+      max-width: 600px;
```
gtvj, line 41: Could we use a `rem` value here? I think it's `37.5rem`

sass/includes/_results-options.scss
```diff
@@ -1,22 +1,43 @@
 .results-options {
-  text-align: right;
+  background-color: $color__green;
+  padding: 1em;
```
gtvj, line 3: Could we use `rem` here

templates/collections/blocks/featured_page.html
```diff
@@ -1,3 +1,3 @@
-<ul class="card-group">
+<ul class="row card-group">
```
gtvj, line 1: We're mixing BEM and Bootstrap here and should separate them so that the `row` is on a parent - `<div>` perhaps - and the component is within. See: https://github.com/nationalarchives/tna-frontend-design-toolkit/blob/main/GUIDELINES.md#avoid-mixing-bem-and-grid-classes-on-the-same-element

templates/collections/blocks/promoted_pages.html
```diff
@@ -3,13 +3,13 @@
 <h2>{{ value.heading }}</h2>
 <p>{{ value.sub_heading }}</p>
 
-<ul class="row card-group">
+<ul class="row card-group--list-style-none">
```
gtvj, line 6: Similar to the comment above. We should avoid mixing BEM and Bootstrap on the same element.

danbentley (approved):
Thanks Ash. Looks Great!

templates/collections/blocks/time_period_explorer.html
```diff
@@ -5,11 +5,11 @@
 {% image value.page.specific.teaser_image fill-508x311 as teaser_image_large %}
 {% image value.page.specific.teaser_image fill-626x383 as teaser_image_extra_large %}
 
-<div class="card-group-promo--green">
+<li class="card-group-promo--green">
     <div class="card-group-promo__card">
         <div class="row">
             <div class="col-md-12 col-lg-7 col-xl-6">
-                <a href="{{ value.page.url }}" class="card-group-promo__card-link">
+                <a href="{{ value.page.url }}" class="card-group-promo__card-link" tabindex="-1">
```
gtvj, line 12: More of a comment than anything else, but we've removed the image link on the basis that the link is also presented in the text below and this avoids the user having to tab multiple times for each card.

templates/collections/explorer_index_page.html
```diff
@@ -5,13 +5,12 @@
...
-    <div class="container">
-        <div class="row">
-            <div class="col-md-12" id="content">
-                {% for block in page.body %}
-                    {% include_block block %}
-                {% endfor %}
-            </div>
-        </div>
+    <div class="container mt-4">
+       <ul class="row card-group">
```
gtvj, line 9: Similar to the comments above, we should avoid mixing BEM and Bootstrap on the same element.

templates/collections/results_page.html
```diff
@@ -14,7 +14,7 @@
                     <h2 class="sr-only">Results</h2>
                     <h3 class="sr-only">Options</h3>
                     {% include 'includes/results-options.html' %}
-                    <ul class="row card-group">
+                    <ul class="row card-group--list-style-none">
```
gtvj, line 17: Similar to comments above, we should avoid mixing BEM and Bootstrap on the same element.

templates/collections/time_period_explorer_index_page.html
```diff
@@ -5,16 +5,16 @@
 {% block content %}
     {% include 'includes/generic-intro--dark.html' %}
 
-    <div class="container">
-        <div class="row">
-            <ul class="row card-group">
+    <div class="container mt-4">
+            <h2 class="sr-only">Select a time period</h2>
+            <ul class="row card-group--list-style-none">
```
gtvj, line 10: Similar to comments above, we should avoid mixing BEM and Bootstrap on the same element.

templates/collections/topic_explorer_index_page.html
```diff
@@ -5,16 +5,16 @@
 {% block content %}
     {% include 'includes/generic-intro--dark.html' %}
 
-    <div class="container">
-        <div class="row">
-            <ul class="row card-group">
+    <div class="container mt-4">
+            <ul class="row card-group--list-style-none">
```
gtvj, line 9: Similar to comments above, we should avoid mixing BEM and Bootstrap on the same element.

templates/includes/card-group-promo--dark.html
```diff
@@ -1,33 +1,32 @@
...
         <div class="row">
-            <div class="col-md-6 col-lg-5 col-xl-4">
-                <a href="{{ page.url }}" tabindex="-1">
-                    <picture>
-                        <source media="(max-width: 576px)" srcset="{{ teaser_image_extra_large.url }}">
-                        <source media="(max-width: 768px)" srcset="{{ teaser_image_large.url }}">
-                        <source media="(max-width: 991px)" srcset="{{ teaser_image_extra_small.url }}">
+            <div class="col-md-12 col-lg-7 col-xl-6">
+                <a href="{{ page.url }}" tabindex="-1" class="card-group-promo__card-link">
+                    <picture class="mt-auto">
```
gtvj, line 15: Rather than using Bootstrap's `mt-auto` here, because this padding is a responsibility of the component I think the margin styles should be within the Sass for the component. I think it's fine to use something like `class="container mt-4"` as you have elsewhere because the relevant element is being controlled solely by Bootstrap. Hope that doesn't sound contradictory - happy to explain on a call if that would be helpful.

templates/includes/card-group-promo--green.html
```diff
@@ -0,0 +1,33 @@
...
+{% image page.teaser_image fill-626x383 as teaser_image_extra_large %}
+
+<li class="card-group-promo--green">
+    <h2 class="card-group-promo__heading">{{ heading }}</h2>
+
+    <div class="card-group-promo__card">
+        <div class="row">
+            <div class="col-md-12 col-lg-7 col-xl-6">
+                <a href="{{ page.url }}" tabindex="-1" class="card-group-promo__card-link">
+                    <picture class="mt-auto">
```
gtvj, line 15: Similar to the comment above, I think the margin for this `picture` element should be within the Sass for the component.

templates/includes/related-content.html
```diff
@@ -12,8 +12,8 @@ <h3 class="sr-only">Event - ‘Before Shakespeare’ at The National Archives</h
                 </picture>
             </div>
             <div class="card-group-secondary-nav__body">
-                <h3 class="tna-card__heading" aria-hidden="true">Event - ‘Before Shakespeare’ at The National Archives</h3>
-                <p><time datetime="2018-08-01">Wednesday 1 August 2018</time></p>
+                <h3 class="card-group-secondary-nav__heading" aria-hidden="true">Event - ‘Before Shakespeare’ at The National Archives</h3>
```
gtvj, line 15: Just a quick question - why are we hiding this from assistive technology with `aria-hidden`?

gtvj (changes requested):
Hi Ash, good work on this PR. Thanks. I've requested a few changes and made a couple of observations. There is also a question or two I've asked.

gtvj:
@AshTNA - one other thing to bear in mind is that, for the sorting we'll need to think about how we're going to support [WCAG 2.1 Success Criterion 3.2.2 (Level A)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html) in the JS version. While this is a pattern we've used before, the testing of KBS revealed it to be a failure. Our options are either to: 

- Warn the user that there's going to be a change of context, or;
- Not cause a change of context (i.e. asynchonously load the updated results)

Happy to discuss. 

All the best, 
Gwyn

templates/includes/related-content.html
AshTNA: Hi @JamesWChan, would you be able to answer this?

templates/includes/card-group-promo--green.html
AshTNA: Thanks @gtvj, looks like the class is not needed with the new styles anyway so I have removed it.

templates/includes/card-group-promo--dark.html
AshTNA: Thanks @gtvj, looks like the class is not needed with the new styles anyway so I have removed it.

AshTNA:
Hi @gtvj,

Thanks for the comments - I have changed the `em` values to `rem`, 0 values, and removed the `mt-auto` from the promo cards.

I'll look into the `<ul class="row card-group>` issue. This was introduced when we conducted the accessibility fixes for the Explorer so it was approved previously in a PR at some point.When I initially implemented that, I was aware that CB's guidance said _avoid_ instead of strictly outlawing it, which is why I believe it was merged when it was first introduced. If I recall correctly, the styles go a bit awry when the `row`  is in a parent `<div>`. And of course a `<ul>` can only have `<li>` elements as its children. I will try and fix it but it may be easier to look at this on a call together - I don't think I figured out the exact reason it was giving  weird behaviour.

Thanks :+1:

gtvj:
Thanks @AshTNA - on the BEM/Bootstrap stuff, sorry to ask you to fix something that had previously been merged in another PR but as I reflect on this (while we feel our way through the new approaches) I would like to keep the BEM and Bootstrap separate if we possibly can. In this instance I'd suggest introducing a containing `<div>` to hold the Bootstrap class - any issues are likely to be a margin or padding. Happy to jump on a call to look at it together if you'd like.

templates/includes/related-content.html
JamesWChan: I copied this code directly from card-group-secondary-nav.php, I haven't changed it. There is an`<h3>` on line 4 with a class of `sr-only` so I guess this is to prevent screen readers from reading the heading out twice. If you want it changed just let me know.

AshTNA: Thanks @JamesWChan 

I'll also need to update this file so that the card is an `<li>` instead of a `<div>`, as we made all the cards sit within a `<ul>` in `ds-wagtail` but I didn't reflect this in `ds-wagtail-frontend` (apologies) - it shouldn't affect anything visually on insights as the SASS in this repo is accounting  for the `<ul>` version of the cards.

JamesWChan: No worries 👍

AshTNA:
Thanks @gtvj - I've got it working by:

- Separating the `row` out into a parent div
- Adding ` display: flex;` `flex-wrap: wrap;` and `width: 100%;` to the `card-group` which mimics what the `row` was doing when both classes were on the same element.

I've also checked the results cards using the Kong environment variables, and they seem to be OK as well.

templates/includes/related-content.html
gtvj: I've been thinking about the doubling of `<h3>` elements within this component and have spoken to @matt-blair about it. Matt has said that he sees the images as being decorative, in which case we can treat them as such - see https://www.w3.org/WAI/tutorials/images/decorative/ 

If we do that we would not need the initial `<h3>` and could treat the second one as a normal heading. This seems OK to me.

@AshTNA @JamesWChan - what do you think?

AshTNA: Thanks @gtvj - that makes sense to me, though I don't know much about this component in general (aside from the card stuff)

JamesWChan: Sounds good to me. In which case, should we also remove the alt text as per the W3 advice?

gtvj: Yes - we'd need to set `alt=""` and remove the heading. If we're agreed, I think we should proceed with this change for all cards that have this duplicate heading pattern.

gtvj (approved):

gtvj (approved):

gtvj:
All merged @AshTNA - thanks for your work on this 👍🏽

PR #42, opened 2021-07-15
Add tests for record description override

danbentley:
Record description override is already merged in. This PR just adds some tests.


Rijidij (approved):
Sorry for the delay.
All looks good to me.

PR #52, opened 2021-07-21
Insights styling fixes

AshTNA:
Hi @JamesWChan 

Would you be able to merge this branch?

This contains the following fixes:

- The `100vw` fix
- Renaming `callout` to `box-out`
- Fixing some issues with the icons. The URLs were incorrect. Also, as the `border-radius` was applied directly to the `` tag, the images were being cut off by this `border-radius`. Therefore I've changed them to `<div>` elements with `background-image` which doesn't get affected by `border-radius`. Hopefully this is OK.
- Preventing `sr-only` elements appearing as contrast issues (by giving them a `background-color` and `color` even though they're never visible) This is just so they don't pollute the WAVE/Lighthouse issues across the whole application.

Would you be able to merge if this is OK?

Thanks :+1:

### DIFF
--- a/sass/etna.scss
+++ b/sass/etna.scss
@@ -55,7 +55,7 @@ These are Etna specific components, created using BEM and following our guidelin
 @import 'includes/specifics';
 @import 'includes/content-block';
 @import 'includes/quote';
-@import 'includes/callout';
+@import 'includes/box-out';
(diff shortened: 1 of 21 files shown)


JamesWChan (approved):
All good 👍

PR #54, opened 2021-07-22
Feature/record disambiguation page

AshTNA:
Hi @gtvj and @danbentley 

Would you be able to approve this PR if all is OK? The editorial has been approved.

@gtvj would you be able to check that the HTML/CSS I have written is OK? I've checked it in Chrome/Firefox/IE11, with WAVe and Lighthouse, and at 400% zoom. To see the disambiguation page, you may need to add the Kong credentials to your `.env` file and then restart your docker containers (you can do this using the docker UI). The URL is here: http://localhost:8000/catalogue/MAF/42/8/

A question as well @gtvj - should I include the `gtm` script as well in this branch, since this is the first branch to include some analytics data?

@danbentley My "Introduce code to include analytics JSONs" commit proposes a technique for including the metrics JSON files. Does this look ok to you? We would then simply use this block to include the metrics JSON on the other applications. The empty JSON values would need to be populated via the backend though. 

Thanks :+1:

Screenshot:

image


danbentley:
Thanks Ash! This looks good.

For a bit of background, when we populate the datalayer with content from the backend, we may create a custom [`filter`](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#built-in-filter-reference) to make testing easier, especially if the requirements become more complex (like branching logic, for example)

Additionally, we'll use [`json_script`](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#json-script) to output the JSON:

> Safely outputs a Python object as JSON, wrapped in a <script> tag, ready for use with JavaScript.
> 
> This is compatible with a strict Content Security Policy that prohibits in-page script execution. It also maintains a clean separation between passive data and executable code.

This is a snippet from another project where we've done this in the past:

```
{% block extra_js %}
	{{ page | datalayer_data | json_script:“datalayer-data” }}

	<script type="application/javascript">
		datalayer = [JSON.parse(document.getElementById(‘datalayer-data’).textContent)];
	</script>
{% endblock extra_js %}
```

danbentley (approved):

AshTNA:
@gtvj I noticed I haven't added the metrics tags to the cards yet - I was going to do this during the Explorer branch as this is where I would properly add the card tags and loops. Is this OK?

I can add a container name to the template now.

gtvj:
Hi @AshTNA - in answer to your question above, I don't think we need to include the GTM script at this point. Helen and Heather need to set it up from their end and I'd like most of the changes relating to GTM to come in within a single PR when they've confirmed things are working on `ds-wagtail-frontend` (though I can see there is backend `dataLayer` work being added here as well as a `data-container-name` attribute).

gtvj (changes requested):

AshTNA:
Thanks @gtvj - I've made these changes. :+1:

gtvj (approved):

PR #56, opened 2021-07-23
Remove line that says toolkit CSS is loaded

gtvj:
This is no longer relevant since we are using the SASS directly within the repo.


AshTNA (approved):
All good :+1:

PR #58, opened 2021-07-23
Feature/refactor generic intro

gtvj:
Hi @AshTNA - as discussed, here's the PR for better managing generic intro.


AshTNA:
Thanks @gtvj - as the `generic-intro-with-prefix` is only used in the design for the [results page ](https://ds-wagtail-frontend.herokuapp.com/collection-explorer/results.php) which is `templates/collections/results_page.html` , what are your thoughts on using the standard `generic-intro` for all the other templates?

gtvj:
Sounds good to me 👍🏽 - I've just pushed a commit which does this.

AshTNA (approved):
All good :+1:

PR #59, opened 2021-07-23
Feature/insights index page

JamesWChan:
Hi @danbentley, @gtvj,

Would it be possible to merge this PR if you're both happy? Sorry there are a few commits that are unrelated to insights (e.g. analytics for the homepage and alerts), they were small and just thought I'd bundle them all together. Thank you 👍

### DIFF
--- a/templates/alerts/tags/alerts.html
+++ b/templates/alerts/tags/alerts.html
@@ -6,7 +6,7 @@
         <div class="row">
             <div class="col-md-12">
                 <span class="alert-message__label" data-alert-title="{{ alert.title }}">{{ alert.title }}</span>
-                <span class="alert-message__description" data-alert-message="{{ alert.message|richtext }}">{{ alert.message|richtext }}</span>
+                <span class="alert-message__description" data-alert-message="{{ alert.message|striptags }}">{{ alert.message|richtext }}</span>
             </div>
         </div>
     </div>
--- a/templates/alerts/tags/alerts.html
+++ b/templates/alerts/tags/alerts.html
@@ -5,8 +5,8 @@
     <div class="container">
         <div class="row">
             <div class="col-md-12">
-                <span class="alert-message__label">{{ alert.title }}</span>
-                <span class="alert-message__description">{{ alert.message|richtext }}</span>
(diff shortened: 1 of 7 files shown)


templates/alerts/tags/alerts.html
```diff
@@ -5,10 +5,10 @@
     <div class="container">
         <div class="row">
             <div class="col-md-12">
-                <span class="alert-message__label">{{ alert.title }}</span>
-                <span class="alert-message__description">{{ alert.message|richtext }}</span>
+                <span class="alert-message__label" data-alert-title="{{ alert.title }}">{{ alert.title }}</span>
+                <span class="alert-message__description" data-alert-message="{{ alert.message|richtext }}">{{ alert.message|richtext }}</span>
```
gtvj, line 9: Hi @JamesWChan - I've just been looking at the `richtext` filter within the Wagtail docs. https://docs.wagtail.io/en/stable/topics/writing_templates.html#rich-text-filter it says this filter will output HTML, which isn't going to work in the context of a `data-` attribute. I might be wrong about this but was wondering if we might be able to use the Django `striptags` filter so that we just get the content: https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#striptags

gtvj (commented):
Hi @JamesWChan - this looks broadly OK to me but I've asked a question about template filters. Other than that, if Dan is happy I'm happy for it to be merged 👍🏽

templates/alerts/tags/alerts.html
JamesWChan: Thanks @gtvj, I've changed that now.

danbentley (changes requested):
Thanks James!

#60 adds dynamic content to the home page which unfortunately means the `HomePage` and `home_page.html` have significant changes.

On Ash's PR (#54), I hinted that the BE implementation of the `datalayer`[ is likely to not use `includes`](https://github.com/nationalarchives/ds-wagtail/pull/54#issuecomment-884818727) so I'm wondering whether it might be worth reverting the changes to `etna/home/models.py` and `home_page.html` but keeping `etna-homepage-analytics-json.html` for reference.

PR #61, opened 2021-07-26
Create Author streamfield component

Rijidij:
- Added to insights page stream block

[DS-283]

[DS-283]: https://national-archives.atlassian.net/browse/DS-283


etna/authors/blocks.py
```diff
@@ -0,0 +1,17 @@
+from django.conf import settings
+
+from wagtail.core import blocks
+from wagtail.images.blocks import ImageChooserBlock
+
+
+class AuthorBlock(blocks.StructBlock):
```
danbentley, line 7: Have you looked at using a separate model for authors so editors don't need to type an author's details each time?

danbentley (changes requested):

etna/authors/blocks.py
Rijidij: Good idea

danbentley: Thanks for adding the `Author` snippet!

PR #63, opened 2021-07-27
Fix migration conflict in insights

Rijidij:
CommandError: Conflicting migrations detected;
multiple leaf nodes in the migration graph:
(0007_add_section_block, 0007_add_teaser_image in insights).


danbentley:
Thanks for this!

PR #64, opened 2021-07-27
Insights promoted item (Full width teaser)

Rijidij:


danbentley:
I've updated the tests to use `staticfiles.finders.find` to check for the existence of the referenced static file. 

I've also included a migration to create the initial categories.

PR #66, opened 2021-07-27
Integrate with disambiguation page with latest frontend changes DS-278

danbentley:
While working on the disambiguation page. I took the opportunity to refine the `SearchManager` and improve the tests.


Rijidij (approved):
Thanks for the walk-through.
Looks good to me!

PR #67, opened 2021-07-27
Create Promoted Item List block

Rijidij:
Branched from DS-277 - Insights promoted item (Full width teaser), which already contains the categories code also required by this block.


danbentley:
Thanks for this. Approved and merged into `main`.

PR #69, opened 2021-07-28
Parse RecordPage.held_by from Kong response

danbentley:
This is a small change to extract held_by from a Kong response.


Rijidij (approved):
Looks good!

PR #78, opened 2021-08-03
Add GTM attributes

gtvj:
This PR adds GTM attributes to all relevant partials (see commit history) and a JavaScript module that manages the `data-card-position` attribute (since this could not be achieved server side).


gtvj:
Problem with the commits here so will continue this in another branch.

PR #79, opened 2021-08-03
Introduce analytics front end

gtvj:
This PR:

1. Introduces the GTM data-* attributes
2. Introduces JavaScript to manage specific GTM attributes which could not be achieved on the back end
3. Removes some redundant includes (for alerts)
4. Adds GTM script and no-script blocks


JamesWChan (approved):
All good 👍

PR #81, opened 2021-08-04
DS-309 + removing next and previous links for MVP

matt-blair:
Hi @gtvj,

Could you look at this PR to make sure you're happy with it? It fixes the issue with the broken hierarchy links.

We've also tagged on removing the next/previous links from the HTML so they don't appear in the MVP for launch.

@danbentley - Copying you into this so you're aware.

Thanks,

Matt.


gtvj (approved):
Thanks @matt-blair. Looks good 👍🏽

PR #84, opened 2021-08-05
Add load static to all templates using it

gtvj:


danbentley:
Thanks Gywn, 

I might have given some misleading advice, we should only need to `load static` in the page's template. All templates included by that template will have access to `static`. Sorry for the confusion.

I'll approve and merge now and we can remove the redundant calls at a later date.

PR #85, opened 2021-08-05
Update all 'src' paths to use 'static'

gtvj:
Hi @danbentley - I hadn't realised that _not_ using `static` within the path would cause an error (my previous PR had updated the code to ensure that where `static` was used it had been loaded). I've done quite a thorough search through the code and I think this is the only instance of a local `src` attribute that doesn't use `static`.

### DIFF
--- a/templates/quotes/blocks/quote.html
+++ b/templates/quotes/blocks/quote.html
@@ -2,7 +2,7 @@
 
 <div class="quote">
     <div class="container">
-        <img class="quote__icon" src="/static/images/fontawesome-svgs/quote-left.svg" alt="" />
+        <img class="quote__icon" src="{% static 'images/fontawesome-svgs/quote-left.svg' %}" alt="" />
         <h3 class="sr-only">{{ value.heading }}</h3>
 
         {% if value.attribution %}


danbentley:
Thanks Gwyn, yeah it's a bit unusual and unfortunately, something that only surfaces on production. 

For reference I'll paste my Slack message on `static`:

> 
> In development mode, if Django receives a request for a static file (`/static/css/dist/styles.css`), it’ll look within the project for `static` directories. These directories can live within an application (`etna/home/static`), or within the templates directory (`templates/static`).
> 
> Once it gathers a list of `static` directories, it’ll search within them for a matching file (`styles.css`).
> 
> You can find the configuration for where Django looks for static content here:
> 
> https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-STATICFILES_FINDERS
> 
> Searching around adds quite a bit of overhead and so in production Django isn’t responsible for serving static content, it’s instead the job of the web server (usually nginx or Apache) which can serve static content much faster and more efficiently than Django.
> 
> However the web server isn’t interested in searching around within an application for static content, it wants all those files to be in a single place, which is where the root `static` directory comes in.
> 
> During a deploy, we run the `collectstatic` command. To gather the static content located within the application and consolidate to a single directory:
> 
> https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django-admin-collectstatic
> 
> (We’re also able to do interesting things during `collectstatic` like add a content hash to filenames so we can cache the  file indefinitely)

gtvj:
Thanks Dan 👍🏽

gtvj:
@JamesWChan @matt-blair @AshTNA - it's worth having a read of Dan's message above about how local development and deployments handle static files (and why the `static` template tag should be used to avoid paths causing problems when moving to deployments).

PR #93, opened 2021-08-09
Add default (empty value) to datalayer

Rijidij:


danbentley (approved):
Thanks for this. Looks Good!

PR #94, opened 2021-08-09
Amend promoted item teaser image implementation

JamesWChan:
Hi @gtvj,

Would it be possible to have a look at this PR? It implements the teaser image for promoted items (e.g. a blog embed) in the same fashion as the hero image so that the image isn't distorted as the screen is resized. Thank you 👍


gtvj (approved):
Thanks @JamesWChan - looks good 👍🏽

PR #95, opened 2021-08-09
Amend which date value is output in templates DS-313

danbentley:
- Remove redundant date fields
 - Rename date_range
 - Output origination_date in templates


Rijidij (approved):
All good! :+1:

PR #101, opened 2021-08-10
Add validation messages for additional circumstances

gtvj:
This will cover where there are non-field errors, including:

- User not found
- Too many failed attempts


JamesWChan (approved):
All good 👍

PR #107, opened 2021-08-11
Feature/fix details page

JamesWChan:
Hi @gtvj, @danbentley, 

Would it be possible to merge this if you're happy? It adds an `id` to the record details `h2` so that the "See full description" jumplink works and it also fixes the details page title (previously it displayed the url and now it displays the record title e.g. "9 Battalion Royal Irish Rifles - The National Archives"). Thank you 👍

### DIFF
--- a/templates/records/record_page.html
+++ b/templates/records/record_page.html
@@ -36,7 +36,7 @@
                 {% endif %}
                 {% include "includes/records/record-access-options.html" %}
 
-                <h2>Description and record details</h2>
+                <h2 id="record-description">Description and record details</h2>
 
                 {% include "includes/records/record-details.html" %}
 
--- a/templates/records/record_page.html
+++ b/templates/records/record_page.html
@@ -1,7 +1,12 @@
 {% extends "base.html" %}
 
 {% load static %}
+{% load i18n %}
 {% load datalayer_tags %}
+
+{% block title %}
+    {{ page.title }} - {% trans 'The National Archives' %}
+{% endblock %}
 
 {% block extra_gtm_js %}
 {{ page | datalayer:request | json_script:"datalayer" }}
--- a/templates/records/record_page.html
+++ b/templates/records/record_page.html
@@ -1,12 +1,7 @@
 {% extends "base.html" %}
 
 {% load static %}
-{% load i18n %}
 {% load datalayer_tags %}
-
-{% block title %}
-    {{ page.title }} - {% trans 'The National Archives' %}
-{% endblock %}
 
 {% block extra_gtm_js %}
 {{ page | datalayer:request | json_script:"datalayer" }}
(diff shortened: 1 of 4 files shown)


gtvj (approved):
Thanks @JamesWChan - this all looks good to me. If @danbentley is happy with how the title tag is being populated then I'm happy to merge it).

templates/records/record_page.html
```diff
@@ -1,8 +1,13 @@
 {% extends "base.html" %}
 
 {% load static %}
+{% load i18n %}
 {% load datalayer_tags %}
 
+{% block title %}
+    {{ page.title }} - {% trans 'The National Archives' %}
```
danbentley, line 8: Thanks James, good fix!

Wagtail passes uses two keys for the current page, `page` and `self`:

https://github.com/wagtail/wagtail/blob/7bf3f813efe8f25e11ddf68a20fd1165f5cd67da/wagtail/core/models/__init__.py#L1092-L1102

The reason why the page title isn't output in `base.html` is because the view used to fetch the record data from Kong is only using `page`.

https://github.com/nationalarchives/ds-wagtail/blob/3f52a64245efd4ed99ce0039d77e60cc760710e8/etna/records/views/records.py#L66-L67

`base.html` uses `self`, which is inconsistent with the rest of our templates. A quick fix for this would be to rename `self` to `page` in `base.html` and then all pages will populate `<title>` in the same way.

danbentley (changes requested):

templates/records/record_page.html
JamesWChan: I must give credit to @gtvj for that one! Thanks @danbentley, I've made those changes but "The National Archives" doesn't seem to be coming through for the details page title and I'm not too sure why

JamesWChan: I must give credit to @gtvj for that one. Thanks @danbentley, I've made those changes but I can't figure out how to get the "The National Archives" suffix to display on details pages

danbentley: Ah, apologies James, I've only given you half a solution there!

When rendering a page. Wagtail queries which site the page belongs to using `Page.get_site` (a single Wagtail instance can host multiple sites)

https://github.com/wagtail/wagtail/blob/7bf3f813efe8f25e11ddf68a20fd1165f5cd67da/wagtail/core/models/__init__.py#L1278

Since the `RecordPage` isn't CMSable it technically doesn't belong to any site. 

But we can override the `get_site` method to return the appropriate site instance:

```# in records/models.py
from wagtail.core.models import Site

class RecordPage(Page):
   …
   def get_site(self):
         """Override to return the Site instance despite this page not belonging to the CMS."""
         return Site.objects.first()
   …
```

I'm happy to make this change if you'd like.

JamesWChan: Thanks Dan, that would be great if you could. Apologies, I know it's a simple change but I'm getting errors when I try running Wagtail locally atm

danbentley: No problem, I've pushed a commit.

JamesWChan: Thanks Dan :)

PR #108, opened 2021-08-11
Add GTM JavaScript to Wagtail repo and refactor

gtvj:
Hi @JamesWChan - could you have a look at this PR? 

It introduces all the GTM tracking code that had previously been stored in a separate repo. The reason for this change is that we've been informed by the analytics team that their previous approach of adding scripts via the GTM console is no longer preferred and they would like us instead to include the scripts within the main repo.


JamesWChan (approved):
Looks good 👍

PR #109, opened 2021-08-11
Fix disambiguation page

danbentley:
Some disambiguation pages returned data that couldn't be parsed into a `RecordPage`.

Additionally, the `datalayer` template tag raised an error because there's no page object to query.


Rijidij (approved):
All good! :+1:

PR #113, opened 2021-08-12
Improve media embed editor experience

Rijidij:
Removed unused, and confusing, fields (width, height, thumbnail)
Got date picker working
Added validation so uploaded files will match the media type.
Removed upload tabs from chooser until Torchbox (completely) fix an [issue](https://github.com/torchbox/wagtailmedia/releases/tag/v0.7.1) with RichText fields.

### DIFF
--- a/etna/media/forms.py
+++ b/etna/media/forms.py
@@ -11,6 +11,11 @@ from wagtail.admin.forms.collections import BaseCollectionMemberForm
 from django.forms.models import modelform_factory
 
 
+# Valid audio/video media types - https://www.iana.org/assignments/media-types/media-types.xhtml
+ALLOWED_AUDIO_MIME_TYPES = ["audio/mpeg",]
+ALLOWED_VIDEO_MIME_TYPES = ["video/mp4",]
+
+
 class BaseMediaForm(BaseCollectionMemberForm):
     class Meta:
         widgets = {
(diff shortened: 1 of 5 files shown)


etna/media/forms.py
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+    permission_policy = media_permission_policy
+
+    def __init__(self, *args, **kwargs):
+        super(BaseMediaForm, self).__init__(*args, **kwargs)
+
+        # Remove unused fields.
+        for name in ("width", "height", "thumbnail"):
+            # these fields might be editable=False so verify before accessing
```
danbentley, line 29: What do you mean by this? It doesn't look like we're checking for `editable=False`

etna/media/forms.py
```diff
@@ -0,0 +1,64 @@
+import magic
+
+from django import forms
+from django.core.exceptions import ValidationError
+
+from wagtail.admin import widgets
+
+from wagtailmedia.permissions import permission_policy as media_permission_policy
+from wagtail.admin.forms.collections import BaseCollectionMemberForm, collection_member_permission_formset_factory
```
danbentley, line 9: Looks like there's an unused import for `collection_member_permission_formset_factory`

etna/media/forms.py
```diff
@@ -0,0 +1,64 @@
...
+            # these fields might be editable=False so verify before accessing
+            if name in self.fields:
+                del self.fields[name]
+
+    def clean_file(self):
+        error_messages = {
+            "mime_type": "%(filename)s (%(mime_type)s) is not an allowed %(type)s file.",
+        }
+
+        audio_types = [
```
danbentley, line 38: I can see us needing to update this list. Could you promote these to be class attributes or [defined in the module](https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/images/fields.py#L12) for visibility and rename to convey their purpose (valid/allowed audio mime types)

danbentley (changes requested):
Looks good. You're getting into some advanced stuff here!

Just a few bits of reorganising required.

etna/media/forms.py
Rijidij: Done

etna/media/forms.py
Rijidij: Thanks. I've removed it now.

PR #114, opened 2021-08-12
Create topic import script.

danbentley:
Because the some of the topic records haven't been pushed to Kong yet. I've updated the templates to skip a record if the fetch fails.

Verified with integration tests.


Rijidij (approved):
Looks good. :+1:

PR #115, opened 2021-08-12
Feature/details page image link

JamesWChan:
Hi @gtvj,

Would it be possible to have a look at this PR? It wraps the image in the image viewer within a link and adds the appropriate focus and hover styles (the image browse page uses light blue for hover and focus for images on a black background and so that's why it was chosen here). I also updated the text link focus style to light blue for consistency.

I added a BEM class to the image link because it seemed like the best way to isolate it. I added a BEM class to the text link for consistency in the HTML.

Thank you 👍

### DIFF
--- a/sass/includes/_image-viewer-panel.scss
+++ b/sass/includes/_image-viewer-panel.scss
@@ -13,17 +13,31 @@
       display: block;
       margin: 0 auto;
     }
+  }
 
-    a {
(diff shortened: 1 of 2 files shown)


gtvj:
Thanks @JamesWChan - this looks good. I just made a tiny tweak to so that the SASS is referencing the outline specific variable.

PR #117, opened 2021-08-12
Allow pseudo references in details-page-human-readable

danbentley:
Pseudo references were parsed out separately from reference numbers in anticipation of Kong needing both the reference_number and pseudo reference passing via separate parameters

Now that Kong treats both reference number and reference numbers with a pseudo reference equally, we should do the same.

Example of a reference with a pseudo reference: /catalogue/CP/~1/


Rijidij (approved):
All good! :+1:

PR #118, opened 2021-08-12
Style disambiguation pagination and trucate item descriptions

gtvj:
This introduces:

1. the same component pattern HTML for pagination as is used in other places (directly in the view)
2. uses the `truncatewords` Django [built-in filter](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#truncatewords) to truncate descriptions within `card-group-record-summary-no-image.html`

### DIFF
--- a/sass/includes/_pagination.scss
+++ b/sass/includes/_pagination.scss
@@ -16,6 +16,7 @@
 
             ol {
                 padding: 0;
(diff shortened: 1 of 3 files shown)


danbentley (approved):
Looks great. Thanks for this!

PR #119, opened 2021-08-12
Add condition to prevent page.subheading being rendered empty

gtvj:
This is just a quick change to prevent an issue which is currently present on the record disambiguation page (that I've just spotted). I've just added a condition to prevent a `<p>` being rendered empty.


danbentley (approved):
Good spot! Thanks for this!

PR #121, opened 2021-08-12
Replace publication date with optional duration field

Rijidij:
[DS-314]

Prob requires some frontend attention.

[DS-314]: https://national-archives.atlassian.net/browse/DS-314


danbentley (approved):
Looks good!

PR #122, opened 2021-08-12
Ensure the edit page is still accessible even if a record isn't available in Kong

danbentley:
It's possible that a referenced record on an insights or results page isn't available in Kong, this fix ensure the edit page is still accessible.


Rijidij (approved):
Looks good! :1st_place_medal:

PR #124, opened 2021-08-12
Explorer - results page: Hide the 'sort' block

Rijidij:
[DS-320]

[DS-320]: https://national-archives.atlassian.net/browse/DS-320


danbentley:
Looks good!

PR #127, opened 2021-08-12
Add summary text to featured page streamfield

Rijidij:
Field named "description" to be consistent with other blocks.

The template for this block uses an include which prints the "introduction" field of the featured page. 
Description takes precedence if populated, otherwise page.introduction is rendered if supplied.

[DS-321]

[DS-321]: https://national-archives.atlassian.net/browse/DS-321


danbentley (approved):
Thanks. Looks good!

PR #128, opened 2021-08-12
Make generic-intro-with-prefix specific to Results

gtvj:
From UAT we have been asked to add the text 'Selected highlights of' which makes this a non-general include. It retains the component HTML but, to avoid confusion, I have renamed the include.


JamesWChan (approved):
All good 👍

PR #133, opened 2021-08-13
Fix issue where record doesn't have a teaser image

danbentley:
`teaser_image` for a record on the results page is optional.

Images should be associated with all highlight records for launch but currently many of them don't have images yet.


Rijidij (approved):
All good! :+1:

PR #137, opened 2021-08-13
Add field for optional curated image

Rijidij:
[DS-324]

The frontend code for this was lifted from the promoted item block template. Which was in turn from a static include file.
So the class names don't correlate with the content type.

Just sayin'...

[DS-324]: https://national-archives.atlassian.net/browse/DS-324


etna/insights/blocks.py
```diff
@@ -13,6 +13,9 @@
 
 class FeaturedRecordBlock(blocks.StructBlock):
     record = RecordChooserBlock()
+    image = ImageChooserBlock(
```
danbentley, line 16: Would you mind renaming to `teaser_image` to be consistent with the record embed on the results page?

https://github.com/nationalarchives/ds-wagtail/blob/5077695c52b20ed67846dbca60466df2145f9f02/etna/collections/models.py#L261-L267

danbentley (changes requested):

etna/insights/blocks.py
Rijidij: No problem.

danbentley (approved):
Thanks!

PR #138, opened 2021-08-13
Put transcript in details element

JamesWChan:
Hi @gtvj,

This PR puts the transcript within a `<details>` element so it can be toggled. This is to solve the issue where a very long transcript would result in a very long webpage.

Thank you 👍

### DIFF
--- a/sass/includes/_media-embed.scss
+++ b/sass/includes/_media-embed.scss
@@ -65,12 +65,33 @@
     background-color: $color__dark;
     font-size: 1.2rem;
     margin-top: 2rem;
-    padding: 2rem;
+    padding: 1rem;
+
+    details {
+      summary{
+        text-align: left;
+        padding: 1rem;
+
+        &:focus {
+          outline: 0.313rem solid $color__focus-blue-outline-dark-bg;
+        }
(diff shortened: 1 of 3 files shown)


sass/includes/_media-embed.scss
```diff
@@ -65,12 +65,33 @@
     background-color: $color__dark;
     font-size: 1.2rem;
     margin-top: 2rem;
-    padding: 2rem;
+    padding: 1rem;
+
+    details {
+      summary{
+        text-align: left;
```
gtvj, line 72: Hi James - minor point but I wonder if the `text-align: left` would be better placed on the `<details>` element and removed from `&__transcript-text`. Apologies if I'm missing something because I can't actually see the rendered output.

gtvj (commented):
Hi James - this all looks good. I've just asked a quick question about the placement of `text-align`.

sass/includes/_media-embed.scss
JamesWChan: Cool, I've made that change

gtvj (approved):
Thanks James 👍🏽

PR #146, opened 2021-08-17
Feature/image viewer fe fixes

JamesWChan:
Hi @gtvj,

This PR fixes the details page image resize issue mentioned in the stand up. It uses a media query at 768px to modify the height and width. It also fixes an issue with the image viewer text link hover style; on localhost it stays white but on live it switches to dark blue on hover so I've added an explicit rule. Thank you 👍


gtvj (approved):
Brilliant - thanks James. Merging now.

PR #147, opened 2021-08-17
Fix the padding bottom on browse pagination

gtvj:
I noticed on the Beta site that there's a margin bottom on the pagination component which seems to have been inherited by the list. This PR overrides that.


JamesWChan (approved):
All good 👍

PR #149, opened 2021-08-17
Add CSS to provide some spacing around XSLT generated markup

gtvj:
We were asked by the data team to provide add CSS in order to improve the presentation of XLST generated markup which appears within the Record Description block of details pages. This included a request to make spans appear to be list items.

I have explained that using CSS to make HTML that lacks appropriate semantics appear as it would if the correct semantics were in place would fail a [Level A WCAG Success Criterion 1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html). and that the best I can do is to introduce some padding.

It has been acknowledged by data colleagues that this is an issue which needs to be addressed at source.

### DIFF
--- a/sass/etna.scss
+++ b/sass/etna.scss
@@ -66,3 +66,4 @@ These are Etna specific components, created using BEM and following our guidelin
 @import 'includes/disambiguation-explanation';
 @import 'includes/login-status';
 @import 'includes/forms';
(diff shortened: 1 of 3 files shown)


JamesWChan (approved):
All good 👍

PR #150, opened 2021-08-17
Add title to 404 page

gtvj:
Having been contacted by analytics colleagues this morning to add a title to the 404 page I noticed that base.html relies upon values that are not available in the 404 page.

Having done a bit of research online it seems the default 404 view provides two variables to the view: `request_path` and `exception`. It might be that `exception` would allow us to add a condition to `base.html` but since I can't set `DEBUG` to `False` (which I'd need to do in order to examine `exception`) it seemed most pragmatic to replicate the static content of base.html within 404.html

### DIFF
--- a/templates/404.html
+++ b/templates/404.html
@@ -1,9 +1,31 @@
-{% extends "base.html" %}
+{% load static %}
 
-{% block body_class %}template-404{% endblock %}
+<!DOCTYPE html>
+<html class="no-js" lang="en">
+<head>
+    <meta charset="utf-8"/>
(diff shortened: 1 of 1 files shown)


gtvj:
Hi @danbentley - does what I've done here seem OK to you? Ideally I'd have used `exception` but this seemed most pragmatic in the circumstances.

danbentley (approved):
Hi Gwyn,

This is perfect! 

Speaking from experience, we don't want to be getting too clever in our error pages and potentially raising a new error.

PR #151, opened 2021-08-18
Front end fixes - Part 1

gtvj:
Hi @AshTNA  - could you review this PR for me and, once merged, ask Dan or Craig to put it onto a staging environment for testing (which I'm hoping @JamesWChan and @matt-blair might be able to help with as I'm on leave tomorrow). If we're able to release it to Live late tomorrow I'll be able to pick up the continued testing and fixes when I'm back on Friday.  

These are fixes I've identified while reviewing the front end of beta.nationalarchives.gov.uk, the sign in (and out) process, the insights index page and the Shakespeare insights page. What I'd like to do is release these before continuing tests.

- [x] Remove redundant meta tags that were causing validation errors
- [x] Set image height to integer
- [x] Convert alerts to use `<div>` rather than `<span>`
- [x] Add ARIA region to alert component
- [x] Add ARIA region to sign in component
- [x] Set `<p>` bottom margin to meet [Success Criterion 1.4.12](https://www.w3.org/TR/WCAG21/#text-spacing)
- [x] Move "Skip to content" link above Alert component (which is repeated on multiple pages) to improve [Success Criterion 2.4.1](https://www.w3.org/TR/WCAG21/#bypass-blocks)
- [x] Edit heading outline for the Insights Index Page. It currently bypasses a `<h2>`.
- [x] Close the `<ol>` which is used in the Jumplinks component
- [x] Remove duplicate `id="analytics-blog-embed"` attributes in Insights pages
- [x] Update `related-resources__description` to use a `<div>` rather than a `<p>`. This is to address an issue where the `richtext` editor field can allow for paragraphs to be added (which obviously can't be children of paragraphs)  
- [x] Remove `<p>` from quote component. This is necessary because the Wagtail richtext field is outputting a
`<p>` tag (and a paragraph can't be a child of a paragraph)
- [x] Remove duplicated `id="analytics-related-resources"` attributes in Insights pages
- [x] Change related-resources component to use a `<ul>` rather than a `<dl>` because the second field of this component is optional in Wagtail.

**Thoughts and observations**

- I've considered whether the footer links should be a in a `<nav>` element but the W3C have said the footer itself is sufficient and `<nav>` would be redundant - see https://www.w3.org/TR/2011/WD-html5-author-20110809/the-nav-element.html
- I've considered whether the `<alert>` component could be supplemented with ARIA or a heading. I decided not to introduce a heading on the basis that its DOM placement would require introducing additional headings that are clearly redundant (which might make the UX worse for users of AT). I have introduced an region role.
- I've considered removing `role="presentation"` from decorative images since they already have an empty alt attribute. I've decided not to do this because I don't think it will do any harm to users but provides a clear signal to developers that the given image is presentational.
- I considered using the `en-GB` variant of the `lang` attribute but found that GOV.UK are using the same as we are (which I'm assuming is both sufficient and for good reason)


AshTNA (approved):
All good :+1:

PR #156, opened 2021-08-20
Front end fixes - Part 2

gtvj:
This PR includes further changes from the front end review and some additional analytics requirements that have been requested. The full changes are: 

- [x] Moving `data-*` attributes onto the link within related items cards
- [x] On details page, adding ID to taxonomy tags list `id="analytics-taxonomy-tags"`
- [x] Adding unique IDs to blog embeds - for example `analytics-blog-embed-1` (incrementing from 1)
- [x] Adding unique IDs to related resources `<ul>` -  for example `analytics-related-resources-1` (incrementing from 1) 
- [x] On details page, push reference and series to data layer` {'event': 'reference', 'series': 'WO 95 - War Office: First World War and Army of Occupation War Diaries', 'reference': 'WO 95/2/3/4'}` - if there’s no third item (i.e. series) then series should be `Above series level`
- [x] Move all DOM dependent JavaScript into a `DOMContentLoaded` event listener 
- [x] Remove `role=presentation` from images with empty `alt` attribute
- [x] Remove `<time>` element from Insights page record embeds 
- [x] Add `aria-label` to `<audio>` elements referencing the transcript which follows 

**Related considerations**

I've found something odd happening with headings in the Letter from Cyril to Morris insights page. We’ve got two headings at different levels that are identical. They are:
- `<h2>` Transcript of the letter
    - `<h3>` Transcript of the letter

It seems the `<h3>` - which is presented only to screen readers - is redundant. I think this is going to come up in the accessibility audit as something which is annoying to users. Looking at the editorial interface it seems these are coming from the **Section Heading** component and the **Quote** component - both of which require a heading. It seems in this circumstance we need the heading within **Quote** to be optional. I have raised this on Slack and added it to the UAT spreadsheet.

### DIFF
--- a/scripts/src/image-browse.js
+++ b/scripts/src/image-browse.js
@@ -4,12 +4,14 @@ import calc_scroll_percent_increment from "./modules/analytics/scroll_tracking/c
 import update_scroll_obj from "./modules/analytics/scroll_tracking/update_scroll_obj";
 import get_scroll_percentage from "./modules/analytics/scroll_tracking/get_scroll_percentage";
 
-add_data_link('.image-browse__listing a');
-
 let scrollObj = {
     'event': 'image-viewer-browse',
     highestScrollPercentage: get_scroll_percentage()
 };
+
+document.addEventListener('DOMContentLoaded', () => {
+    add_data_link('.image-browse__listing a');
+});
 
 document.addEventListener('scroll', e => {
     const percentageIncrement = calc_scroll_percent_increment();
--- a/scripts/src/image-browse.js
+++ b/scripts/src/image-browse.js
@@ -4,20 +4,20 @@ import calc_scroll_percent_increment from "./modules/analytics/scroll_tracking/c
 import update_scroll_obj from "./modules/analytics/scroll_tracking/update_scroll_obj";
 import get_scroll_percentage from "./modules/analytics/scroll_tracking/get_scroll_percentage";
 
-let scrollObj = {
-    'event': 'image-viewer-browse',
-    highestScrollPercentage: get_scroll_percentage()
-};
-
 document.addEventListener('DOMContentLoaded', () => {
     add_data_link('.image-browse__listing a');
-});
 
-document.addEventListener('scroll', e => {
-    const percentageIncrement = calc_scroll_percent_increment();
-    scrollObj = update_scroll_obj(percentageIncrement, scrollObj);
-});
+    let scrollObj = {
+        'event': 'image-viewer-browse',
+        highestScrollPercentage: get_scroll_percentage()
+    };
 
-window.addEventListener('beforeunload', e => {
-    push_to_data_layer(scrollObj);
+    document.addEventListener('scroll', e => {
+        const percentageIncrement = calc_scroll_percent_increment();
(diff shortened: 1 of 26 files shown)


templates/media/blocks/media-block--audio.html
```diff
@@ -26,7 +26,7 @@ <h3 class="media-embed__heading">
 {% endif %}
 
 <div class="media-embed__audio-container">
-    <audio controls="">
+    <audio controls="" {% if value.transcript %}aria-label="Transcript provided below"{% endif %}>
```
JamesWChan, line 29: Would we also want to put this on video blocks?

scripts/src/image-browse.js
```diff
@@ -4,13 +4,15 @@ import calc_scroll_percent_increment from "./modules/analytics/scroll_tracking/c
...
 let scrollObj = {
     'event': 'image-viewer-browse',
     highestScrollPercentage: get_scroll_percentage()
 };
 
+document.addEventListener('DOMContentLoaded', () => {
+    add_data_link('.image-browse__listing a');
+});
+
 document.addEventListener('scroll', e => {
```
JamesWChan, line 16: Just out of curiosity, would we want to move everything inside `DOMContentLoaded`? I think `calc_scroll_percent_increment` uses `document` to calculate the height and I think it's also called in `get_scroll_percentage` under certain conditions. I'm probably wrong but was just wondering.

scripts/src/modules/analytics/add_unique_ids.js
```diff
@@ -0,0 +1,23 @@
+const add_unique_ids = () => {
+
+    const items = [
+        {selector: '.blog-embed', id_prefix: 'analytics-blog-embed'},
+        {selector: '.related-resources ul', id_prefix: 'analytics-related-resources'}
+    ];
```
JamesWChan, line 6: Would we also want to include the `record-embed-no-image` class? It's currently a component in `featured_record.html`. I've checked the Insights pages and I've only seen it used once but it can be used more than once and it currently has `analytics-record-embed-no-image` set as the `id` in the HTML.

JamesWChan:
Hi @AshTNA,

Would you be able to give my changes a once-over? I've made some comments above which hopefully justify the changes I've made. If you're happy, would it be possible to merge and ask for a release? Thank you 👍

AshTNA:
Hi @gtvj @JamesWChan 

I have pushed fixes for the details page focus issues mentioned on Slack.

I will approve this branch but ideally i'd like this branch to go onto a staging server before I merge it, as recreating all the content locally is a bit daunting. I don't think the backend have the capacity to put this on any servers today though so hopefully this work will be done by tomorrow.

Thanks :+1:

AshTNA (approved):

scripts/src/image-browse.js
gtvj: I think it's OK as it is. Think of `DOMContentLoaded` as an event, whereas `calc_scroll_percent_increment` and `get_scroll_percentage` will access `document`, they'll do in different contexts. We also need access to `scrollObj` from within the `scroll` event handler. Happy to discuss this though if you'd like to.

PR #159, opened 2021-09-06
Details page fixes

gtvj:
- [x] Add name to image preview link
- [x] Fix `<title>` tags on Image Browse and Image Viewer pages
- [x] Remove unnecessary number from Image Viewer `<h1>`
- [x] Fix hierarchy global so that `<div>` does not appear as child of `<summary>`
- [x] Add ARIA role and label to image viewer component


AshTNA (approved):
All good :+1:

PR #162, opened 2021-09-07
Feature/fix image viewer focus

JamesWChan:
Hi @gtvj,

This PR applies the light blue focus styles to the image viewer canvas when it receives focus. I switched to using `border` rather than `outline` because the `outline` wasn't visible when zoomed in, whereas `border` is and I had to apply `!important` because OpenSeadragon's CSS was overriding it. 

I've applied `height: 95%` when full screen is activated because the bottom border wasn't showing otherwise. I did try experimenting with `padding`, `margin`, and positioning i.e. `bottom` but I couldn't quite get it to work properly with those properties. And I had to apply `!important` because OpenSeadragon's CSS was overriding it again.

I've also put an `outline: none` rule on `focus-visible` because this was causing a thin dark blue line to appear on the sides of the window when the canvas received focus.

I've tested this in Chrome, Firefox, Edge, Safari, and IE11 and it all seems to be working okay (apologies for the delay, I kept encountering issues when testing locally in LambaTest).

This PR also addresses a transpilation issue which was causing the image viewer to not display in IE11.

Would it be possible to merge if you are happy? Thank you 👍

### DIFF
--- a/scripts/src/image-viewer.js
+++ b/scripts/src/image-viewer.js
@@ -40,13 +40,16 @@ if (image_source) {
         }
     });
 
-    canvas.addEventListener("focus", () => {
-        header.style.borderBottom = "solid 0.3125rem #8EC0EC";
-        footer.style.borderTop = "solid 0.3125rem #8EC0EC";
-    });
+    if(canvas) {
+        canvas.addEventListener("focus", () => {
+            header.style.borderBottom = "solid 0.3125rem #8EC0EC";
+            footer.style.borderTop = "solid 0.3125rem #8EC0EC";
+        });
 
-    canvas.addEventListener("blur", () => {
-        header.style.borderBottom = "none";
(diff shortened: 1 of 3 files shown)


scripts/src/image-viewer.js
```diff
@@ -36,4 +39,14 @@ if (image_source) {
             full_screen_button.focus();
         }
     });
-}
\ No newline at end of file
+
+    canvas.addEventListener("focus", () => {
```
gtvj, line 43: Hi James - could you make a tiny tweak to this. Because it's possible that `document.getElementsByClassName` might  return an empty `HTMLCollection` could we wrap the calls to `addEventListener` in either a `try / catch` or an `if`?

gtvj (changes requested):

gtvj (approved):
Many thanks James. I'm going to merge this now.

PR #164, opened 2021-09-17
Refactor scroll_to_bottom so that it works in IE11

JamesWChan:
Hi @gtvj,

This PR modifies the scroll_to_bottom function to work in IE11 (the previous approach, which used `scrollTo`, is not compatible with IE). I've tested on Chrome, FF, IE, Edge, and Safari and didn't see any issues. Would it be possible to merge if you are happy?

Thank you 👍


gtvj (approved):
Hi James, thanks for your work on this. Following the feedback observation of a Dragon user I expect we'll be moving away from the scroll within a scroll. But it's still worth fixing this.

PR #191, opened 2021-10-08
Fix/focus order issue

### LINKED ISSUE #165: Manage focus on Sign In form following validation failure - DAC_Focus_Order_Issue1
DAC have identified a usability issue on the Sign In form following a validation error. This is described on Page 12 of the report with ID DAC_Focus_Order_Issue1

We can fix this by:

- determining if the page is being rendered following a validation failure and, if so;
- using JavaScript to direct focus to the validation error summary message (being careful to ensure we are managing the focus of this non-focusable element in a sensible way).

### Why wasn't this identified in the in-house testing

While DAC have described this as a failure of [Success Criteria 2.4.3 (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html) it doesn't seem to be relevant to that Success Criterion because:

- The focus order of the page is correct and does not change upon a validation failure
- We are performing server-side validation, so the focus direction they describe is actually a change of context (i.e. a new page being loaded). 

On this basis I would argue that this is not a Level A failure but is instead a usability issue. That said, it has been identified as a possible improvement so we should fix it as outlined above. It might be useful to discuss this with DAC so that we can determine how best to handle multi-page forms and server side validation in future (is it best to direct focus to a specific part of a new page etc.)

gtvj:
This PR fixes #165

### DIFF
--- a/scripts/src/modules/focus_element_if_found.js
+++ b/scripts/src/modules/focus_element_if_found.js
@@ -1,0 +1,22 @@
+/**
+ * This module manages focus in the event that
+ * a given element is found to be on the page
+ *
+ * @param {string} needle_class - a class on the element we are looking for
+ */
+
+const focus_element_if_found = (needle_class = '') => {
+
+    const needle_elements = document.getElementsByClassName(needle_class);
+
+    if (!!needle_elements.length) {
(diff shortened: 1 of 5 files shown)


AshTNA (approved):
All good :+1:

PR #192, opened 2021-10-12
Accessibility improvements to hierarchy global

### LINKED ISSUE #167: Breadcrumb links
DAC have identified an issue on page 15 with ID `DAC_Breadcrumb_Links_Issue1` with the hierarchy on the details page which relates to: 

* [Success Criterion 1.3.3 Sensory Characteristics](https://www.w3.org/WAI/WCAG21/Understanding/sensory-characteristics.html)
* [Success Criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)

They have described the hierarchy as acting like a breadcrumb trail (which I suppose it does).

We can fix this by: 

* Using an ordered list for the items
* Using `aria-label` to describe the purpose of the navigation. The W3C [provide an example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html) but it may be that simply labelling the hierarchy as 'breadcrumb' will not suffice. Perhaps 'Where am I in the catalogue?' (which is also used in the associated heading) would be better (?).

### Why was this not identified by in-house testing
To be honest, I've never thought of the hierarchy as a breadcrumb. The list of links are contained within a `<details>` element that is described within the `<summary>` by a heading which says "Where am I in the catalogue?". What the DAC testing has shown is that, while this approach might work well for contextual interaction it does not provide sufficient information when a user encounters the links within the navigational element out of context (i.e. navigating the page by links). I've learned something new here.

gtvj:
This PR fixes #167 

Rather than using `aria-label` as has been suggested by DAC I've opted to use `aria-labelledby` in order to avoid introducing redundancy (i.e. having the text "Where am I in the catalogue" appear in multiple places) and follow the DRY principle.

### DIFF
--- a/templates/includes/records/hierarchy-global.html
+++ b/templates/includes/records/hierarchy-global.html
@@ -4,10 +4,10 @@
 
     <details id="js-hierarchy-global" open>
         <summary id="analytics-hierarchy-link">
-            <h2 class="hierarchy-global__heading">Where am I in the catalogue?</h2>
+            <h2 class="hierarchy-global__heading" id="hierarchy-label">Where am I in the catalogue?</h2>
         </summary>
 
-        <ul class="hierarchy-global__list">
+        <ol class="hierarchy-global__list" aria-labelledby="hierarchy-label">
             {% for item in page.hierarchy %}
                 {% is_page_current_item_in_hierarchy page item as is_current_item %}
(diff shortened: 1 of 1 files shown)


JamesWChan (approved):
All good 👍

PR #194, opened 2021-10-14
Feature/dac fixes

JamesWChan:
Hi @AshTNA,

This PR resolves the following issues:

- DAC_Usability_Errors_Issue1
- DAC_Illogical_Heading_Structure_Issue1
- DAC_Consistent_Navigation_Issue1
- DAC_Non-descriptive_Headings_Issue1

It might be useful to have a call so I can talk you through them if you think that would be a good idea. When you are happy, would it be possible to merge? Thank you 👍

### DIFF
--- a/sass/includes/_image-viewer.scss
+++ b/sass/includes/_image-viewer.scss
@@ -64,7 +64,7 @@
 
   &__toolbar {
     margin: 1.3rem;
-    text-align: right;
+    text-align: left;
     line-height: 1rem;
 
     &.fullpage {
@@ -72,7 +72,6 @@
       position: absolute !important;
       margin: 0;
       bottom: 0;
-      padding: 0.5rem;
       text-align: left;
       z-index: 999;
     }
@@ -95,6 +94,7 @@
       border-radius: 0;
       cursor: pointer;
       color: $color__white;
+      margin: 0.5rem;
 
       &:focus {
         outline: 5px solid $color__focus-blue-outline-dark-bg;
@@ -141,6 +141,7 @@
 
     li {
       display: inline-block;
+      padding: 0.5rem;
     }
 
     @media (max-width: $screen__sm) {
@@ -165,14 +166,18 @@
   &__pagination {
     padding: 1rem;
     text-align: right;
+    line-height: 1rem;
 
     @media (max-width: $screen__sm) {
       padding: 0.2rem;
     }
 
+    li {
+      margin: 0.5rem;
+    }
+
     a {
       margin-right: 1rem;
-      padding: 0 0.2rem;
 
       &:focus {
         text-decoration: none;


templates/records/image-viewer.html
```diff
@@ -11,26 +11,6 @@
                 <div class="image-viewer__reference">
                     <h1>{{ images.has_other_pages }}<strong>{{ page.reference_number }}</strong> <span>- Image {{ index|add:'1' }} of {{ images_count }}</span></h1>
                 </div>
```
AshTNA, line 13: Hi James,  

I noticed a couple issues  with the image viewer changes which may need to be fixed:

- When hitting full screen, the toolbar with Zoom In, Zoom Out, etc. goes back to the top. On my mobile, because "Exit Fullscreen" is now in the same place as the green "All Records" link, sometimes when I hit "Exit fullscreen" it also clicks the "All Records" link and  takes me back to  the page listing all the images. I'm guessing the toolbar moving is also an issue as the links are moving which was one of the accessibility issues.
-  If possible, I'm also wondering if the links could be given some more margin or padding as they are also quite close together when using mobile with touch screen.

templates/includes/card-group-record-summary-no-image.html
```diff
@@ -8,7 +8,14 @@
 <li class="col-sm-12 col-md-6 col-lg-4">
     <div class="card-group-record-summary">
         <a href="{% url 'details-page-machine-readable' iaid=record_page.iaid %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{{ record_page.title }}">
-            <h3 class="card-group-record-summary__heading">{{ record_page.title }}</h3>
+            <p class="card-group-record-summary__heading">
+                {% if record_page.reference_number %}
+                <span class="sr-only">
+                    {{ record_page.reference_number }}
```
AshTNA, line 14: Question, do we want the hyphen between the catalogue reference and title still for screen readers? e.g. `{{record_page.reference_number}} - ` otherwise the links  on  Dragon VoiceOver and the way it's read out by a screen reader may be different without the hyphen.

templates/includes/card-group-record-summary.html
```diff
@@ -3,7 +3,14 @@
 <li class="col-sm-12 col-md-6 col-lg-4">
     <div class="card-group-record-summary">
         <a href="{% url 'details-page-machine-readable' iaid=record_page.iaid %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary" data-card-title="{{ record_page.title }}">
-            <h3 class="card-group-record-summary__heading">{{ record_page.title }}</h3>
+            <p class="card-group-record-summary__heading">
+                {% if record_page.reference_number %}
+                <span class="sr-only">
+                    {{ record_page.reference_number }}
```
AshTNA, line 9: Same as previous comment with the hyphen

AshTNA (approved):
All good :+1:

PR #200, opened 2021-11-01
Fix/links that leave site

### LINKED ISSUE #174: Links that leave the site - DAC_Non-descriptive_Links_Issue6
On page 30 of the report DAC describe an issue with ID `DAC_Non-descriptive_Links_Issue6` that affects YouTube video links which: 

* Do not describe the video
* Do not explain that the link will open in a new window

They have suggested we add this information to the visible link text. 

### Why wasn't this picked up by in-house testing
It doesn't seem to relate to a WCAG failure.

AshTNA:
Hi @gtvj 

Would you be able  to merge this PR? It fixes #174 by:

> * Baking the "Opens in a new window" message (and, if needed, `target="_blank"`) into the template.
> * Getting an OK for the design approach to incorporating ("opens in a new site") - speaking to Simon and/or Matt and/or Maria
> * Change the name "CTA label" in the editorial interface to something that's going to be more meaningful to editors (for example, "Call to action text" - or even "Link text" might be an option) - speaking to Piriya and/or Matthew
> * Update the `help_text` to explain that the call to action text will be presented to the user along with a message that this link will open in a new window - speaking to Piriya and/or Matthew
> * Update or the template and SASS include for `promoted-item`(removing any references to `blog-`)

One thing I've noticed is that we use "Opens in a new window" around the site, however  here I've gone for "Opens in a new tab" and that was okayed from MDV. Am I correct in assuming that "Opens in a new tab" is more correct than "Opens in a new window" and we should use that from now on? A window in my opinion implies it opens a whole new browser window.

To test this component, create an Insights Page and create a Promoted Item.

Thanks :+1:


sass/includes/_buttons.scss
```diff
@@ -8,6 +8,11 @@
   margin: 0.5rem 0;
   border: none;
 
+  &__target-blank-label {
+    display: block;
+    font-size: .7em;
```
gtvj, line 13: Hi @AshTNA - would this be better as `rem` rather than `em`?

gtvj (changes requested):
Hi Ash - just a small comment about CSS units but this otherwise looks good. Thanks for your work on this.

sass/includes/_buttons.scss
AshTNA: Thanks @gtvj  - I've made this change

PR #206, opened 2021-11-11
DS-330

mozaki1:


mozaki1:
Wagtail upgrade to 2.15

jacobtoppm (changes requested):
Thanks @mozaki1! It looks like you've got some Poetry conflicts, and the CI currently isn't running against your PR - would you be able to rebase it on main, fix any conflicts, and push again so we can see if it passes CI?

etna/records/tests/test_blocks.py
```diff
@@ -266,18 +268,22 @@ def test_add_featured_records(self):
...
-        self.assertEqual(len(responses.calls), 2)
+        self.assertEqual(
+            self.insights_page.body[0].value["introduction"],
+            "This is some text"
+        )
+        self.assertEqual(
+            self.insights_page.body[0].value["records"][0].iaid,
+            "C123456"
+        )
+        self.assertEqual(len(responses.calls), 3)
```
jacobtoppm, line 286: Why is there an extra API call now? Might be worth investigating

pyproject.toml
```diff
@@ -8,10 +8,10 @@ authors = ["Dan Bentley <dan@numiko.com>"]
 python = "^3.8"
 gunicorn = "^20.1.0"
 elasticsearch = "^7.12.0"
-wagtail = "^2.12.3"
+wagtail = "2.15"
```
jacobtoppm, line 11: 2.15.1 is now released, which contains some bug fixes. Probably worth switching to that - it should be very straightforward

jacobtoppm (changes requested):
Generally looks great! Just one thing to check and a quick update

etna/records/tests/test_blocks.py
mozaki1: The call for self.insights_page.body[0].value["record"].iaid , calls FetchQuery(self.model).get_multiple(iaid=iaid) , which is another KongClient fetch.

jacobtoppm (approved):
Nice work! I'd suggest squashing the commits when we merge this in, but otherwise all looks good to me

PR #207, opened 2021-11-15
Feature/card redesign

### LINKED ISSUE #171: Card implementation and link text - DAC_Non-descriptive_Links_Issue3
On page 22 of the report DAC describe an issue with ID `DAC_Non-descriptive_Links_Issue3`. The full description needs to be read to be understood (I won't try to summarise it here) but this does not seem to relate to a WCAG success criterion.

To address the issues which are described (including those that are included in the analyst comments), I suggest we need to:

- to update all card components so that the link text includes only what is necessary (some new styling may be necessary and we'll need to test the impact of this change quite thoroughly). **This might be easiest to achieve through a co-design session, but it might be sufficient to just include the heading text**. Once we have come up with a solution and implemented it, we will need to do a review of the pages to ensure all links are sufficiently well described (and not duplicated).
- remove the use of `<figure>` and `<figcaption>` elements from cards (on the basis that the images within the `<figure>` element is hidden from assistive technology by an empty `alt` attribute. 

A separate issue is mentioned for JAWS users where "group start" is announced. According to the [JAWS documentation](https://freedomscientific.github.io/VFO-standards-support/html.html), the voice UI only announces this when it encounters a `<fieldset>` and I wonder if this might be a bug which has arisen from the use of `<figure>` that contains an image with an empty `alt` (that is very much a guess though). We should explore this further too.

### Why wasn't this picked up in in-house testing
It doesn't seem to relate to any Level A or Level AA success criteria

AshTNA:
Hi @JamesWChan  

Would you be able to merge this?

It is the card redesign work which fixes #171 - it essentially separates the card from being one big link. Now the card image is a separate link with an `aria-hidden=true` and `tabindex=-1`. The card title then has its own regular link, and the description is no longer a link. 

## Notes
- As our card HTML is duplicated I made the changes across 7 templates. Hopefully there hasn't been any drift between the HTML as it did get a bit confusing keeping them in sync! 
- This removes the `<figure>`  and `<figcaption>`  from the cards. I recall us discussing why we were doing this, but I can't remember where. I think it was to do with screen readers reading out the cards in a certain way when there was a `<figure>`.
- I added a conditional loop to render the `no-image` card HTML on results pages, instead of the regular card, when a record has no image.

The design was approved by SW, and it is based on the GDS cards. Feel free to suggest any changes to the design as well.

## Testing

You can test the cards by:
1. Cloning the branch/pulling the live database.
2. Running the SASS so the styles update.

### card-group-secondary-nav
cards

### card-group-record-summary
results

Thanks :+1:


gtvj:
Hi @AshTNA,

Thanks for your work on this: I think this PR is going to need a populated database to be reviewed (there are so many variants of cards and content that I'd like to see them all 'in action', so to speak). 

In terms of your questions - we spent quite a bit of time going through each of the cards with Helen to ensure the data attributes were being placed in the right places (and, if I recall correctly, there were some variations). It might be that those things you've identified would reflect that but I think we'll need to check with @hoake. @AshTNA - could you pick this up with Helen?

@jacobtoppm - how is the work on setting something up so that developers can get local instances of the database up and running going? I'd like to do this PR when that's in place.

jacobtoppm:
@gtvj - Unfortunately most of my time has been going towards putting the sessions together for the last few days, I've only had a little time to look at the CMS data import so far, so not a huge amount of progress. I'm hoping to get some more time for this soon!

gtvj:
Thanks @jacobtoppm - it would be great (if at all possible) if you were able to progress setting up the 'data onto developer machines' process this week. Let me know if you think that would be doable.

jacobtoppm:
No problem @gtvj - I definitely think that's doable. It's just been a lack of time so far between the sessions (and pairing work) over the last couple of days, but from what I've been able to look into so far, I think getting a basic pull + anonymise working shouldn't be too tricky. In future I think it would be good to restrict access to the production environment and just occasionally pull and anonymise data from that onto staging (and have devs exclusively pull from staging) as we discussed, but for now I'll focus on getting the basic version working.

JamesWChan (approved):
Looks good 👍

PR #208, opened 2021-11-18
Feature/data pull

jacobtoppm:
- Add utility commands in local development via a fabfile
- Add the ability to pull production data and media, and anonymise

In future, we should probably repurpose this to:
- Have production periodically pulled and anonymised to staging
- Have developers only pull from staging
Thereby limiting access to unanonymised data further


ababic:
@gtvj - I too struggled with the connection issues you mentioned on Slack. I fixed this here by ensuring the relevant certificates are installed before doing anything SSH-related (see `enable_platform_ssh()`)

Everything is now rebased, tidied and working like a treat for me here. Would it be okay to merge?

PR #210, opened 2021-11-22
Remove negative margin from audio container

JamesWChan:
Hi @AshTNA,

This PR just removes the styling on the `&__audio-container` which was applying `margin-right: -3rem` and `margin-left: -3rem` as it was causing a spacing issue when viewing an insights page on mobile. I've checked the change in Chrome, Edge, Firefox, IE11, and Safari in LambdaTest and didn't spot any issues.

Would it be possible to merge if you are happy? Thank you 👍

### DIFF
--- a/sass/includes/_media-embed.scss
+++ b/sass/includes/_media-embed.scss
@@ -51,11 +51,6 @@
     text-align: left;
   }
 
-  &__audio-container {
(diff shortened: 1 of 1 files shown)


AshTNA (approved):
All good :+1:

PR #211, opened 2021-11-22
Chore: Add fabric commands for common tasks

ababic:
I've added these fab commands locally to make it easier to do most common things, so thought I might as well contribute them back (along with some documentation changes).

Fabric can be installed everywhere that Python can, so is a great option for simplifying this sort of thing on Django projects.


JamesWChan (approved):
Looks good 👍

PR #212, opened 2021-11-22
Chore: Silence warnings on startup

ababic:
TODO:
- [x] Add an AppConfig class for all apps with models, with a `default_auto_field` value specified
- [x] Update dependencies to silence warnings for other apps where possible


ababic:
There are still two startup warnings remaining, but these relate to models from `django-allauth`, which hasn't yet been updated (this can happen later, once they release a new version)

PR #214, opened 2021-11-25
Chore: Tweak data anonymisation settings

ababic:
Having `BIRDBATH_REQUIRED = True` in dev settings is problematic for those who have not pulled data from a live environment and still want to run the app.

Since the `pull-production-data` command runs birdbath automatically, I think it's safe enough to remove these settings overrides.


AshTNA (approved):
All good :+1:

PR #215, opened 2021-11-29
Add superuser re-creation instruction to README

gtvj:
Explains that superuser will need to be recreated after having run `fab pull-production-data`. While this information is output on the terminal when the command is run it is easily missed and the inclusion of it in the README is to make it more evident alongside other steps outlined there.


ababic (approved):
LGTM!

PR #216, opened 2021-11-30
Update wagtailmedia config to match the new format (0.8+)

ababic:
`wagtailmedia` was recently upgraded in another PR, and we failed to spot that our custom config was no longer being respected (the format for configuration changed - without warning - to use a single dictionary). This was most likely missed because nobody had content locally, and so couldn't see the impact of the change. Hopefully, now that realistic content is available to pull down from other environments, this sort of thing should be much easier to spot in future.


AshTNA (approved):
All good :+1:

PR #217, opened 2021-12-02
Added sr-only span to fix issue#169

### LINKED ISSUE #169: Sign up (to the newsletter) link text - DAC_Non-descriptive_Links_Issue1
On page 18 of the report with ID `DAC_Non-descriptive_Links_Issue1` DAC describe an issue relating to the text used on the newsletter sign up link. They describe this as being relevant to [2.4.4 Link Purpose (In context) Level A](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) and [2.4.9 Link Purpose (Link only) Level AAA](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only.html)

We can fix this by adding a `sr-only` `<span>` to the element which appends "to The National Archives' mailing list"

### Why wasn't this identified by in-house testing
In house testing was against WCAG Levels A and AA only. The stated Level A failure refers to "programmatically determined link context" which I had mistakenly understood to include the associated heading. It turns out that this is not the case: 

> In HTML information that is programmatically determinable from a link in English includes text that is in the same paragraph, list, or table cell as the link or in a table header cell that is associated with the table cell that contains the link. [W3C definition of programmatically determined]

We'll need to bear this in mind and ensure this is reflected in editorial guidance.

jamesbiggs:
Added sr-only span with required text to make links more accessible for screen reader users, which fixes #169.

### DIFF
--- a/templates/includes/footer.html
+++ b/templates/includes/footer.html
@@ -12,7 +12,7 @@
                     <p class="cta-banner-newsletter__description">Subscribe now for regular news, updates, and priority booking events.</p>
                 </div>
                 <div class="col-sm-12 col-lg-2">
-                    <a class="tna-button--dark" href="https://www.nationalarchives.gov.uk/contact-us/email-updates/">Sign up</a>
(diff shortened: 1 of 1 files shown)


AshTNA (approved):
All good :+1:

PR #218, opened 2021-12-02
Added sr-only span with required text to fix issue

### LINKED ISSUE #172: Links within hierarchy local - DAC_Non-descriptive_Links_Issue4
On page 26 of the report DAC describe an issue with ID `DAC_Non-descriptive_Links_Issue4` which says the purpose of the links within should be contained in the link itself. The report doesn't specify which WCAG Success Criterion this relates to but, since the required text is provided in the programatically determined context (which would, I think, meet [Success Criterion 2.4.4 Link Purpose (In context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html), it seems this might be [Success Criterion 2.4.9 Link Purpose (Link only) at Level AAA](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-link-only)

DAC have suggested we fix this by introducing a `sr-only` `<span>` element that contains the information "Previous", "Next" and "This record is in" statements for each link. 

I'm not entirely sure about this approach because I'd have thought it might lead to users encountering the same information twice as they move through the page in context, for example: "This record is in This record is in ...". I wonder whether it would be better to simply bring the visually shown text into the link. 

### Why wasn't this picked up by in-house testing
It seems to relate to a Level AAA success criterion.

jamesbiggs:
Added sr-only span with required link text to make links more accessible to screen reader users, which fixes #172.

### DIFF
--- a/templates/includes/records/hierarchy-local.html
+++ b/templates/includes/records/hierarchy-local.html
@@ -6,7 +6,7 @@
                     <p class="hierarchy-local__nav-item">
                         <span class="hierarchy-local__nav-indicator hierarchy-local__nav-indicator--left"></span>
                         <span class="hierarchy-local__nav-label">Previous</span>
-                        <a href="{% url 'details-page-machine-readable' page.previous_record.iaid %}" data-link="Previous" class="link--bold">{{ page.previous_record.iaid }}</a>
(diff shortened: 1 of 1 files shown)


AshTNA (approved):
All good :+1:

PR #220, opened 2021-12-14
DF-29: Add project-wide cache-control and vary headers

ababic:
Relates ticked: https://national-archives.atlassian.net/browse/DF-29

This PR implements the non-infra-specific changes to Wagtail, which includes:
- Making sure caching is disabled for private URLs
- Applying a default set of cache control and vary headers to hardcoded public urls
- Applying a default set of cache control and vary headers to Wagtail page responses

The aim is to have all of this stuff managed centrally and applied in a consistant way, so that you generally only have to worry about whether views are public or private in `urls.py` - and don't really have to do much more than that.

This PR also introduces a `BasePage` model to this project, which is generally a pretty good idea. It gives you a central place to override page functionality to meet your needs, or to add fields that you know will be needed for every page.


danbentley (approved):
Thanks for talking me through these changes @ababic . Looks great!

PR #221, opened 2022-01-10
Fix: Windows compatibility changes 'pull-data' operations

ababic:
For `enable_platform_ssh()`:
* Use os.path.expanduser to get a platform-specific known_hosts path
* Use ssh-keygen to remove existing keys before adding a new one

General:
* Use subprocess.run() instead of local() for docker commands, as `shlex.escape()` doesn't work for windows


ababic:
Changes tested by @TNA-Allan on local windows environment.

PR #224, opened 2022-01-17
Convert RecordPage to dataclass [DF-45]

danbentley:
Coverted `RecordPage` to `dataclass`.

The amend went well; kong data correctly mapped to dataclass' attributes. Attributes render correctly but the amend raised a few issues (all of which should be fixed by this PR):

 - Broken analytics. `datalayer` tag expects a page instance. Following the lead from the image viewer, I created a separate tag for records.
 - Record picker would no longer work. Amended Chooser/Chosen mixins to inherit from non-model mixins.
 - Alerts. Alerts tag expects a page to query for attached alert snippet. Amended `base.html` to wrap alert output in block to allow templates to override.


templates/base.html
```diff
@@ -47,7 +47,7 @@
 <body class="{% block body_class %}{% endblock %}">
 {% wagtailuserbar %}
 <a href="#maincontent" class="skip-to-content-link sr-only sr-only-focusable" data-link="Skip to main content">Skip to main content</a>
-{% alerts page=page %}
+{% block alerts %}{% alerts page=page %}{% endblock %}
```
ababic, line 50: Instead of this, I wonder if it would be better to have the following template structure?

1. `base.html` (can be subclassed for anything, regardless of whether it's a page)
2. `pages/base.html` (extends `base.html` to add page specific things (like the alerts)
3. `pages/specific_page_type.html` (extends `pages/base.html` to renders any content unique for this page type)

You will soon have all of the custom search views that will no doubt also want to extend `base.html`, so that template would do well to assume less about how it's being used.

etna/records/views/records.py
```diff
@@ -65,7 +65,7 @@ def record_page_view(request, iaid):
 
     return render(
         request,
-        page.get_template(request),
+        "records/record_page.html",
```
ababic, line 68: Now that the record isn't a page, maybe following regular Django template naming conventions would be a good shout here? (e.g. `records/record_detail.html`)

etna/records/models.py
```diff
@@ -1,16 +1,16 @@
...
-from wagtail.core.models import Page, Site
+from wagtail.core.models import Site
 
 from ..ciim.models import MediaManager, SearchManager
 from .transforms import transform_image_result, transform_record_page_result
 
 
-class RecordPage(Page):
+@dataclass
+class RecordPage:
```
ababic, line 13: Maybe just `Record` would be a more appropriate name now?

templates/base.html
ababic: Wagtail pro tip: You can use the [`{% wagtail_site %}`](https://docs.wagtail.io/en/stable/topics/writing_templates.html#multi-site-support) tag to access the current `site` in the `base.html`, which you don't need a page for.

This is generally a lot more efficient, because the tag will cache the appropriate `Site` on the current `request` object, where it will be reused by things like the `pageurl` tag and retrieving site-specific settings. So, although it isn't 'free' to use - it's likely that query is already being triggered by other code, so you might as well profit. `Page.get_site()` always triggers at least two database/cache backend queries.

etna/records/models.py
danbentley: Agreed! That makes much more sense to me. I've pushed the change.

etna/records/views/records.py
danbentley: Good point! I've renamed the templates and views for records.

templates/base.html
danbentley: Thanks Andy! That's a good tip. I've made the change to `base.html` and removed `Record.get_site`.

The template restructuing makes sense. It might be worth opening the discussion up to a wider audience, particularly frontend. Worth discussing in today's dev meeting?

ababic: > Worth discussing in today's dev meeting?

I have no time scheduled for this week unfortunately (unsure what is happening contractually), but would be good if you could mention it :)

etna/records/views/choosers.py
```diff
@@ -59,17 +58,29 @@ def get(self, request, pk):
...
-    model = RecordPage
+    model = Record
     page_title = "Choose a record"
     per_page = 10
-    order_by = "iaid"
-    fields = ["iaid", "title"]
+
+    def get_choose_view_attrs(self):
+        attrs = super().get_choose_view_attrs()
+        if hasattr(self, "model"):
```
mozaki1, line 73: Hi Dan, Do we need to do check if hasattr(self, "model"): ? I can see it has already a model and assigned to the Record

mozaki1: Maybe can do  return super().get_choose_view_attrs() ?

mozaki1: Or maybe  attrs["model"] = self.model and then return attrs ?

etna/records/views/choosers.py
```diff
@@ -59,17 +58,29 @@ def get(self, request, pk):
...
+            attrs["model"] = self.model
+
```
ababic, line 75:
```suggestion
        attrs.update(model=self.model)
        return attrs

```

etna/records/views/choosers.py
```diff
@@ -59,17 +58,29 @@ def get(self, request, pk):
...
+        return attrs
+
+    def get_chosen_view_attrs(self):
+        attrs = super().get_chosen_view_attrs()
+        if hasattr(self, "model"):
+            attrs["model"] = self.model
+
+        return attrs
```
ababic, line 83:
```suggestion
        attrs = super().get_chosen_view_attrs()
        attrs.update(model=self.model)
        return attrs
```

danbentley: Thanks both. Committed the change.

The unnecessary check was left behind from when we copied `ModelChooserViewSet.get_chosen_view_attrs` over to our chooser:

https://github.com/wagtail/wagtail-generic-chooser/blob/6d21c64d2f0e3ab6342def4548785c93df9be2db/generic_chooser/views.py#L670-L675

mozaki1 (approved):

ababic (approved):

PR #225, opened 2022-01-20
Reduce width of main body text

JamesWChan:
Hi @AshTNA,

Would it be possible to merge this if you are happy? It reduces the text width on insights pages and uses `margin: auto` to keep everything centred. Thank you 👍


AshTNA (approved):
All good :+1:

PR #227, opened 2022-01-26
Basic Integration with Kong Open Beta API DF-38

danbentley:
This branch fixes a few immediate issues and will become the starting point for sprint 4's integration work.


Puththiran:
I looked at the changes, it looks good to me

etna/ciim/models.py
```diff
@@ -181,9 +163,6 @@ def count(self):
         self._fetch(start=0, size=0)
         return self._response["hits"]["total"]["value"]
 
-    def __iter__(self):
-        return SearchQueryIterator(self)
-
     def first(self):
         """Return the first item in resultset without or None if nothing is found."""
         try:
```
mozaki1, line 168: Not sure why do we rely on exceptions to return, can we do something like this if self[0] is None: return None

etna/ciim/client.py
```diff
@@ -8,15 +8,15 @@
...
 class KongClient:
-    def __init__(self, base_url, api_key):
+    def __init__(self, base_url, api_key, verify_certificates=True):
         self.base_url = base_url
         self.session = requests.Session()
         self.session.headers.update({"apikey": api_key})
+        self.session.verify = verify_certificates
 
     def fetch(self, **kwargs):
         kwargs["ref"] = kwargs.pop("reference_number", None)
```
mozaki1, line 18: What if pop fails here ? and the statement not within the try.

mozaki1:
I would like to ask in general, how exception logging is handled?

etna/ciim/models.py
mozaki1, line 168: Not sure why do we rely on exceptions to return, can we do something like this if self[0] is None: return None

danbentley: Unlike a lot of languages where an exception needs to be handled in an _exceptional_ case, in Python it's common to first attempt to perform an action (like fetching an item by it's key) and handle the fact possibility of the value not being present than first checking for it:

Easier to ask for forgiveness than permission ([EAFP](https://docs.python.org/3/glossary.html#term-EAFP))

vs 

Look before you leap ([LBYL](https://docs.python.org/3/glossary.html#term-LBYL))

etna/ciim/client.py
danbentley: `kwargs` is adictionary containing keyword arguments for the function.

We're using `pop` here to fetch a value with the key `"start"` from the dictionary or return the default value `0`.

https://docs.python.org/3/library/stdtypes.html?highlight=pop#dict.pop

mozaki1: I see since you are providing the default value None as a second argument.

PR #228, opened 2022-02-01
Disable poetry virtualenv and add devcontainer setup for VS Code

ababic:
- Rewrites Dockerfile to install poetry without using a virtualenv
- Adds some files to for VS Code to pick up

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.163.1/containers/docker-existing-docker-compose
+{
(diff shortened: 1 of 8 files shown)


ababic:
@danbentley I wanted to check whether the Dockerfile changes here are okay? It appeared to me that the Dockerfile used by platform.sh is completely different, but I could be wrong about that!

danbentley:
Thanks for talking me through these changes @ababic. Looks great!

Just to confirm, the changes to the `Dockerfile` will only affect local development, configuration for the platform.sh build process is held within [`.platform.app.yaml`](https://github.com/nationalarchives/ds-wagtail/blob/9d7a94ccb91d0eb05ad0c9fde093ea79efb44a87/.platform.app.yaml#L40-L42)

gtvj (approved):
Thanks @ababic - looks good 👍🏽

PR #229, opened 2022-02-02
Formalize Black formatting for the project

ababic:
Commit messages should be pretty self-explanatory :) Shout if you are unsure about anything.


danbentley:
Thanks Andy! This looks great! Feels good to finally have the whole project formatted.

I think I'm definitely going to be caught out and add a push a few unformatted migrations!

Is it worth documenting in the README?

PR #230, opened 2022-02-03
Df 52

mozaki1:
Integrating with kong public beta.

### DIFF
--- a/etna/ciim/models.py
+++ b/etna/ciim/models.py
@@ -6,7 +6,7 @@ from .client import KongClient
 from .exceptions import (
     DoesNotExist,
     InvalidQuery,
-    KongError,
+    KongAPIError,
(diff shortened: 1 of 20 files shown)


etna/ciim/models.py
```diff
@@ -152,7 +152,7 @@ def __getitem__(self, key):
             return self.model(**transformed_result)
 
         self._fetch(start=key.start, size=count)
-        translated_results = [self.model.transform(r) for r in self._results[key]]
+        translated_results = [self.model.transform(r) for r in self._results[key] if r is not None]
```
danbentley, line 155: Thanks for this @mozaki1! Looks good. 

Thanks for updating the record fixture and the transform function.

Is this amend to filter out `None` and/or the amend to the chooser something you'd be able to write a test for?

PR #232, opened 2022-02-08
DF-81

danbentley:
- Improved documentation
 - Explicit keyword parameters (matching Kong’s parameters where possible)
 - Type hints for parameters  (including enums matching Kong's parameter definitions)
 - Return types
 - Transform lists into comma-separated strings to pass to Kong
 - Simplified error handling. Kong should correctly respond with a non-200 status code on error.
 - Add support for `fetchAll/`

Additionally, In order to run the whole test suite I needed to disable image-based tests as the Kong Open beta API has no `/media` endpoint.

### DIFF
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -1,11 +1,17 @@
 import enum
+import json
 import logging
 
-from typing import Optional
+from typing import Any, Optional
 
 import requests
 
-from .exceptions import ConnectionError, KongError
+from .exceptions import (
+    KongBadRequestError,
+    KongCommunicationError,
+    KongInternalServerError,
+    KongServiceUnavailableError,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -73,6 +79,13 @@ def format_list_param(items: Optional[list]) -> Optional[str]:
 
 class KongClient:
     """Client used to Fetch and validate data from Kong."""
+
+    http_error_classes = {
+        400: KongBadRequestError,
+        500: KongInternalServerError,
+        503: KongServiceUnavailableError,
+    }
+    default_http_error_class = KongCommunicationError
 
     def __init__(
         self,
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -39,7 +39,7 @@ class SortOrder(str, enum.Enum):
 
 
 class Template(str, enum.Enum):
-    """Include @template block to assist parsing/rendering of record data.
+    """@template block to include with response.
 
     Supported by all endpoints.
     """
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -24,7 +24,7 @@ class Stream(str, enum.Enum):
     INTERPRETIVE = "interpretive"
 
 
-class Sort(str, enum.Enum):
+class SortBy(str, enum.Enum):
     """Options for sorting /search results by a given field."""
 
     TITLE = "title"
@@ -135,7 +135,7 @@ class KongClient:
         keyword: Optional[str] = None,
         web_reference: Optional[str] = None,
         stream: Optional[Stream] = None,
-        sort: Optional[Sort] = None,
+        sort_by: Optional[SortBy] = None,
         sort_order: Optional[SortOrder] = None,
         template: Optional[Template] = None,
         show_buckets: Optional[bool] = None,
@@ -160,7 +160,7 @@ class KongClient:
             Return matches on references_number
         stream:
             Restrict results to given stream
-        sort:
+        sort_by:
             Field to sort results.
         sortOrder:
             Order of sorted results
@@ -183,7 +183,7 @@ class KongClient:
             "keyword": keyword,
             "webReference": web_reference,
             "stream": stream,
-            "sort": sort,
+            "sort": sort_by,
             "sortOrder": sort_order,
             "template": template,
             "showBuckets": show_buckets,
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -130,17 +130,7 @@ class KongClient:
             "expand": expand,
         }
 
-        # remove empty values to make logged requests cleaner.
-        params = {k: v for k, v in params.items() if v is not None}
-
-        try:
-            response = self.session.get(
-                f"{self.base_url}/data/fetch", params=params, timeout=self.timeout
-            )
-        except requests.exceptions.ConnectionError as e:
-            raise ConnectionError from e
-
-        return self.validate_and_parse_response(response)
+        return self.make_request(f"{self.base_url}/data/fetch", params=params).json()
 
     def search(
         self,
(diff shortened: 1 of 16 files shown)


etna/records/views/choosers.py
```diff
@@ -23,7 +23,7 @@ class KongModelChooserMixinIn(ChooserMixin):
 
     def get_object_list(self, search_term=""):
         """Filter object list by user's search term"""
-        return self.model.search.filter(term=search_term, stream="evidential")
```
mozaki1, line 26: Still passing empty, the chooser pop will fail on the initial load. In DF 52 I return an empty list if Not keyword.

mozaki1: This is still hardcoded  'stream="evidential" ', I believe this is an enum now can we not use something like stream=Stream.EVIDENTIAL

etna/ciim/client.py
```diff
@@ -1,85 +1,266 @@
...
-            or kwargs.pop("reference_number", None)
-            or kwargs.pop("term", None)
-            or ""
-        ):
-            kwargs["term"] = term
-
-        # In Python 'from' cannot be a kwargs. Map 'start' to 'from' before making request
-        kwargs["from"] = kwargs.pop("start", 0)
+        # remove empty values to make logged requests cleaner.
+        params = {k: v for k, v in params.items() if v is not None}
```
mozaki1, line 198: This line of code "params = {k: v for k, v in params.items() if v is not None}"  is repeated in other places , can we not make it a function ?

etna/ciim/client.py
```diff
@@ -1,85 +1,266 @@
...
+    interpretive:
+        Any content written _about_ evidential i.e: Research Guides, Blog posts
+    """
+
+    EVIDENTIAL = "evidential"
+    INTERPRETIVE = "interpretive"
+
+
+class Sort(str, enum.Enum):
+    """Options for sorting /search results by a given field."""
```
mozaki1, line 28: Maybe SortBy?, for example, SortBy.TITLE

etna/ciim/client.py
```diff
@@ -1,85 +1,266 @@
...
+class SortOrder(str, enum.Enum):
+    """Options for sort order for /search results."""
+
+    ASC = "asc"
+    DESC = "desc"
+
+
+class Template(str, enum.Enum):
+    """Include @template block to assist parsing/rendering of record data.
+
```
mozaki1, line 43: If this is the case " to assist parsing/rendering"  why we are passing it to Kong ?

etna/ciim/client.py
mozaki1: Maybe this code  if any({k: v for k, v in kwargs.items() if v is None}): in the SearchManager is no longer needed since the validation is in the KongClient.

etna/records/views/choosers.py
danbentley: Thanks Mo.

I can understand the confusion as there are multiple streams of work being carried out in similar areas. This PR doesn't supersede the work in DF-52.

Good point on using the enum! No point in writing it if I'm not going to use it! 😄

etna/ciim/client.py
danbentley: Cheers @mozaki1 

Good point, I think as soon as cleaning the params becomes more complex, we can move to a function. I think since the code is straightforward single line, it's ok as is for now.

ababic:
Hi @danbentley. I have a pattern for this sort of thing, where I convert exceptions to more specific error types to help with debugging, and basically try to hide as much "special" behaviour in some higher-order functions, so that I can keep the individual endpoint methods as clean as possible: https://github.com/nationalarchives/ds-judgments-public-ui/pull/16/files#diff-cdccc6f4bc34c68c2f90781a4cded97dcf7058208ff15098c6db2f9501fd9

I like the use of a Session in your solution though - definitely keep that! And the enums... beautiful beautiful enums :)

etna/ciim/client.py
danbentley: Yeah, I agree `SortBy` is a better name. I'm in two minds though because I'd like the client's method signature to match Kongs spec where possible (the exception being reserved works like `from`):

`        
def search(
    …
    sort: Optional[Sort] = None,
    …
`

Using a `SortBy` enum for a `sort` parameter seems like a mismatch that could cause confusion.

On the other hand, we could name our enums, `<kong-param>Option` i.e: `SortOption` to avoid the issue.

@ababic do you have any thoughts?

etna/ciim/client.py
ababic: I would suggest having this sort of 'prep' done in another method... a little bit like what I've done here: https://github.com/nationalarchives/ds-judgments-public-ui/pull/16/files#diff-cdccc6f4bc34c68c2f90781a4cded97dcf7058208ff15098c6db2f9501fd9

etna/ciim/client.py
ababic: I prefer `SortBy` too. On the consistency front, I would say that it probably doesn't matter too much. After all, the parameters passed to the API are already in a different case much of the time. I think we should go with `sort_by` / `SortBy` if it makes the client easier to interact with. As long as this it passed to the API as `sort`, we should be good :)

danbentley: Thanks both! Updated to `sort`/`SortBy`

etna/ciim/client.py
danbentley: The `@template` block is easier to parse than searching through nested collections for data (see the transform functions).

I've updated the comment to make it clear that Kong is including the `@template` block in its response.

mozaki1: Thank you , but the question is why do we pass **@template** in the request ?

danbentley: We can request a specific `@template` for Kong to include in the response.

From the API spec:

```json
{
    "name": "template",
    "in": "query",
    "description": "Template to return",
    "required": false,
    "schema": {
        "type": "string",
        "enum": [
            "details",
            "results"
        ]
    }
},
```

It's worth noting this is Kong's concept of a template, not Django/Wagtail's.

etna/ciim/models.py
```diff
@@ -182,7 +182,7 @@ def __init__(self, model):
         self.model = model
 
     def get(self, iaid=None, reference_number=None, expand=False):
-        response = self.client.fetch(iaid=iaid, reference_number=reference_number, expand=expand)
+        response = self.client.fetch(iaid=iaid, id=reference_number, expand=expand)
```
TNA-Allan, line 186: @danbentley - Currently we are 'fecth'ing with id=None. `{record_detail_view->Record.search.get(iaid, expand=True), SearchManager.get()} ` since 'id' is Generic identifier to fetch, Thought we can we have 'reference_number' replaced with 'id' to match kong api param used for fetch.

danbentley: Thanks Allan, I agree.

It's not something I can address in this PR but it will be addressed in subsequent work.

ababic (approved):

PR #233, opened 2022-02-14
Df-52b

mozaki1:
- Fixing chooser popup
- Removing @summary from record.json to fix unit tests
- Added InValidResult exception to handle error in the transform function
- Added create_record_no_source to test hits value equal to zero
- Added test_transforms.py to test the transform function.
- Remove @ from summary in test_related_record_with_no_identifier
- Add three unit tests to test 'invalid results coming from kong


danbentley (approved):
Thanks for this Mo!

Particularly for updating the transform function to handle the updated response from Kong.

As discussed on our call, the checks in the transform function are likely to be superseded by `DF-100`, where the simpler `APIManager` shouldn't call the transform function with invalid data. Apologies again for the overlapping streams of work!

PR #234, opened 2022-02-15
DF-100

danbentley:


etna/records/views/records.py
```diff
@@ -19,25 +20,30 @@ def record_disambiguation_view(request, reference_number):
...
+    page_number = int(request.GET.get("page", 1))
+    offset = (page_number - 1) * per_page
 
-    if len(pages) == 0:
+    count, records = Record.api.search(
+        web_reference=reference_number, offset=offset, size=per_page
+    )
+
+    if len(records) == 0:
         raise Http404
```
TNA-Allan, line 32: Can the check for Http404 be handled in APIManager like fetch

Puththiran:
Can we remove the "DeprecatedSearchManagerTest" as SearchManager is no longer exist?

Other than that generally it looks good to merge with main

etna/records/views/records.py
danbentley: Hi @TNA-Allan. Thanks for taking a look at this PR!

In this instance, we're actually raising a `Http404` exception if Kong response contains no results for the fetch by `webReference`.

In other words, if there are no records belonging to the `webReference` respond with a 404.

danbentley:
> Can we remove the "DeprecatedSearchManagerTest" as SearchManager is no longer exist?
> 
> Other than that generally it looks good to merge with main

Thanks @Puththiran!

You're right that `SearchManager` no longer exists. Because of this, the test you mention ensures that if someone tries to access `Record.search` then a `DeprecationWarning` is raised.

etna/records/widgets.py
```diff
@@ -30,7 +30,7 @@ def get_instance(self, pk):
         """Fetch related instance on edit form."""
         try:
             return Record.search.get(iaid=pk)
-        except (KongAPIError, SearchManagerException):
```
TNA-Allan, line 33: - APIModel.search raises DeprecationWarning, so Record.search could raise 
- API for Search does not have iaid param, instead Fetch has iaid param

etna/records/widgets.py
```diff
@@ -30,7 +30,7 @@ def get_instance(self, pk):
...
+        except (KongAPIError, APIManagerException):
             # If there's a connection issue with Kong, return a stub Record
             # so we have something to render on the ResultsPage edit form.
             return Record(iaid=pk, title="", reference_number="")
```
TNA-Allan, line 36: - Why do we return a stub record to render something when its an error. Should we not have an error page so that it must be rectified - may be some alert?
- Should these error be handled separately?

etna/records/blocks.py
```diff
@@ -84,8 +85,8 @@ def value_from_form(self, value):
             return value
 
         try:
-            return self.target_model.search.get(iaid=value)
-        except (KongAPIError, SearchManagerException):
+            return self.target_model.api.fetch(iaid=value)
+        except (KongAPIError, APIManagerException):
             # If there's a connection issue with Kong, return a stub Record
             # so we have something to render on the ResultsPage edit form.
             return self.target_model(iaid=value)
```
TNA-Allan, line 92: - Why do we return a stub record to render something when its an error. Should we not have an error page so that it must be rectified - may be some alert?
- Should these error be handled separately?

mozaki1: I believe returning a stub record, may hide the actual error, in this case, we may have two errors KongAPIError or APIManagerException

etna/records/views/choosers.py
```diff
@@ -17,21 +19,32 @@ class KongModelChooserMixinIn(ChooserMixin):
     """
 
     # Model belonging to this chooser, set via <model>ChooserViewSet.
-    model = None
+    model: Record = None
 
     # Allow models to be searched via chooser. Hides the search box if False
     is_searchable = True
 
     def get_object_list(self, search_term=""):
```
mozaki1, line 27: Maybe this is redundant?

etna/ciim/models.py
```diff
@@ -1,237 +1,77 @@
...
-        response = self.client.fetch(iaid=iaid, id=reference_number, expand=expand)
-
-        results = response["hits"]["hits"]
-        results_count = response["hits"]["total"]["value"]
-        if results_count == 0:
+    def search(self, **kwargs) -> tuple[int, list[APIModel]]:
+        """Make request to /search and transform the response."""
+        response = self.client.search(**kwargs)
+        count, hits = self._extract_hits_and_count(response)
+        return count, [self.model.from_api_response(h) for h in hits]
```
mozaki1, line 54: Maybe we can do  : 
 if count == self.__RecordNotFound: """this could be a const private member"""
            return count, []
since we know the count result now, so to  avoid executing this line :  [self.model.from_api_response(h) for h in hits]

etna/records/views/records.py
```diff
@@ -51,17 +57,14 @@ def record_detail_view(request, iaid):
     view is accessible from a fixed URL.
     """
     try:
-        page = Record.search.get(iaid, expand=True)
+        page = Record.api.fetch(iaid=iaid, expand=True)
```
mozaki1, line 60: Maybe we could use , something like that: 
api_manager = APIManager(Record)
page = api_manager.fetch(iaid=iaid, expand=True)
I find it a bit clearer, for me, it seems like we are using the record to get the APIManager.

etna/records/views/records.py
mozaki1: How is the Http404 rendered to the user?

etna/records/views/records.py
```diff
@@ -19,25 +20,30 @@ def record_disambiguation_view(request, reference_number):
...
-    if len(pages) == 0:
+    count, records = Record.api.search(
+        web_reference=reference_number, offset=offset, size=per_page
+    )
+
+    if len(records) == 0:
         raise Http404
 
     # if the results contain a single record page, redirect to the details page.
-    if len(pages) == 1:
```
mozaki1, line 28: Maybe a const, "magic number", So maybe __OnlyOneRecodFound

etna/ciim/models.py
```diff
@@ -1,237 +1,77 @@
...
+
+    @classproperty
+    def search(cls):
+        raise DeprecationWarning(
+            f"{cls.__name__}.search is deprecated. Use {cls.__name__}.api instead."
+        )
 
-        return result
+    @classproperty
+    def api(cls) -> APIManager:
```
mozaki1, line 31: I think that the model needs No access to the APIManager, for me, this is the separation of concern , and all api calls are within the concern of the APIManager, for example I don't think we should do something like that : 
page = Record.api.fetch(iaid=iaid, expand=True).

etna/ciim/models.py
```diff
@@ -1,237 +1,77 @@
...
 
 
-class ResultPending(Exception):
-    """Raised if result being requested that hasn't yet been fetched."""
+class APIModel(ABC):
+    """Model representation of API data.."""
 
+    @classmethod
+    @abstractmethod
+    def from_api_response(cls, response: dict) -> APIModel:
```
mozaki1, line 17: Not sure why the model in our case the Record is handling the api reponse , I believe this is the APIManage duty

etna/ciim/models.py
```diff
@@ -1,237 +1,77 @@
...
+    def fetch(self, **kwargs) -> APIModel:
+        """Make request to /fetch and transform the response."""
+        response = self.client.fetch(**kwargs)
+        count, hits = self._extract_hits_and_count(response)
+        if count == 0:
             raise DoesNotExist
-        if results_count > 1:
+        if count > 1:
             raise MultipleObjectsReturned
+        return self.model.from_api_response(hits[0])
```
mozaki1, line 65: Here there are no guarantees hits[0] will actually succeed, maybe  “Ask forgiveness” pattern?

etna/collections/models.py
```diff
@@ -260,8 +260,8 @@ def record(self):
         skip this record on the results page.
         """
         try:
-            return Record.search.get(iaid=self.record_iaid)
-        except (KongAPIError, SearchManagerException):
+            return Record.api.fetch(iaid=self.record_iaid)
+        except (KongAPIError, APIManagerException):
             return None
```
mozaki1, line 265: We are eating the exception here:   except (KongAPIError, APIManagerException):             
                                                         return None
Maybe logit somewhere before return

etna/records/views/images.py
```diff
@@ -79,10 +66,8 @@ def image_browse(request, iaid):
     if not page.is_digitised:
         raise Http404
```
mozaki1, line 67: Why Http404 always ? I think we should check the response status code from Kong first, and if the status is 404 then maybe then.

etna/records/widgets.py
danbentley: Thanks Allan. Good spot. I've replace the call to `Record.search`.

etna/records/blocks.py
danbentley: Thanks both.

You're right, We are hiding API errors from the user. 

Let's say record `C123456` is added to an `InsightsPage`. 

But the day after, `C123456` is removed from the catalogue.

If we don't consume the error, editors would be unable to remove the record from the `InsightsPage`.

etna/records/widgets.py
danbentley: Moving conversation to https://github.com/nationalarchives/ds-wagtail/pull/234#pullrequestreview-887408124

etna/records/views/records.py
mozaki1: So I believe 404 file was not found , but here we are saying the record is not found so the actual status code is 200 , but raising 404 file not found. I believe we should have an exception class with, maybe RecordNotFoundException

etna/ciim/models.py
danbentley: Defining an abstract method on the model allows us to accommodate different model's needs.

Later down the line, we may re-implement `Image` as an `APIModel`. It's then the responsibility of the `Image` class to dictate how we go from response data to an `Image` instance.

Think of this method a bit like constructor overloading in C# 😄

etna/ciim/models.py
danbentley: This is a similar pattern to the way we'd use Django's ORM (`Model.objects`)

https://docs.djangoproject.com/en/4.0/topics/db/queries/#retrieving-objects

`Records.api` is just a convenient way of accessing a preconfigured `APIManager` instance, ready to fetch and transform `Record` data.

etna/ciim/models.py
mozaki1: Not really the answer :)

etna/ciim/models.py
mozaki1: Thank you Dan :) , 
But the actual question and form design point view, why the model is performing the Kong api search, It is slightly breaking MVC pattern, in MVC a model is only responsible for the data definition, required or not, length, relationship with the other models, primary key and foreign, ...,

etna/records/blocks.py
TNA-Allan: - If the record is removed from the catalogue, then it will also impact its access from the front-end user.  How long does a record live in ES in order for it to be retrieved? 
- Is there some functional understanding between how long a record can be retained in ES, accessed by front-end user and changed by Editor user?

etna/ciim/models.py
danbentley: I understand your concern, this might worth a chat if you want to discuss further.

The Model isn't performing the fetch for data. 

```python
records = Record.api.fetch(iaid=iaid, expand=True)
```

Is shorthand for:

```python
manager = APIManager(Record)
records = manager.fetch(iaid=iaid, expand=True)
```

mozaki1: Thank you Dan.
Maybe this is my last comment on this, I believe the Model must not reference the APIManger.

etna/ciim/models.py
danbentley: I know what you mean. I feel uncomfortable directly accessing data from a list at the best of times! 

However, before we call `hits[0]`, we have ensured there are some results:

```python
        if results_count == 0:
             raise DoesNotExist
```

And there's only a single result:

```python
        if results_count > 1:
             raise MultipleObjectsReturned
```

etna/records/views/images.py
danbentley: A 404 is an appropriate response if the resource (not necessarily a file) we're requesting can't be found:

https://docs.djangoproject.com/en/4.0/topics/http/views/#the-http404-exception

In this instance, we're saying if the page doesn't have images (if it's not digitised) then 404.

etna/ciim/models.py
ababic: @mozaki1 was this comment meant for elsewhere?

etna/ciim/models.py
```diff
@@ -1,237 +1,77 @@
...
-        result = self.results[index]
-        if not result:
-            raise ResultPending
+        raise NotImplementedError
+
+    @classproperty
+    def search(cls):
+        raise DeprecationWarning(
+            f"{cls.__name__}.search is deprecated. Use {cls.__name__}.api instead."
+        )
```
ababic, line 28:
```suggestion
    def search(cls):
        raise DeprecationWarning(
            f"{cls.__name__}.search is NOT deprecated. Use {cls.__name__}.api instead."
        )
```

etna/ciim/models.py
ababic: Thanks for explaining on slack @mozaki1. So, the proposal as for something like this?

```suggestion
        if not count:
            return 0, []
        return count, [self.model.from_api_response(h) for h in hits]
```

mozaki1: Yes , looks good

etna/records/blocks.py
danbentley: Discussed logging exceptions in separate ticket.

etna/records/views/records.py
danbentley: Moving discussion to https://github.com/nationalarchives/ds-wagtail/pull/234#pullrequestreview-888019163

etna/ciim/models.py
danbentley: Related discussion https://github.com/nationalarchives/ds-wagtail/pull/234#pullrequestreview-888019163

etna/records/views/records.py
danbentley: Thanks @mozaki1 , I believe this was discussed with @ababic on the call?

etna/records/views/choosers.py
danbentley: Thanks @mozaki1, you're right!

Pushed an update to remove.

PR #237, opened 2022-03-01
CHORE: Prep for datalayer tickets DF-74 + UN-141

ababic:
The aim of this PR is to lay the groundwork for individual datalayer enhancement tickets by providing a more convenient way for page classes (and record/result classes) to make type-specific additions/changes.

In the process, I also:
- Simplified the way datalayer code is rendered (a simple tag that renders to a template)
- Introduced a `base_page.html` template where page-specific template code can live - Allowing us to remove a large amount of repeating code from individual page templates
- Used the `{% wagtail_site %}` tag in `base.html` to derive the 'current site' from the request (which is then cached and used for other things)

### DIFF
--- a/templates/includes/gtm-datalayer.html
+++ b/templates/includes/gtm-datalayer.html
@@ -1,6 +1,5 @@
+{{ data | json_script:"gtmDatalayer" }}
 <script>
     window.dataLayer = window.dataLayer || [];
-    {% if data_json %}
-    window.dataLayer.push({{ data_json|safe }});
-    {% endif %}
+    window.dataLayer.push(JSON.parse(document.getElementById('gtmDatalayer').textContent));
 </script>
--- a/templates/includes/gtm-datalayer.html
+++ b/templates/includes/gtm-datalayer.html
@@ -1,0 +1,6 @@
+<script>
+    window.dataLayer = window.dataLayer || [];
+    {% if data_json %}
+    window.dataLayer.push({{ data_json|safe }});
+    {% endif %}
+</script>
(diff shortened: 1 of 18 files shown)


templates/includes/gtm-datalayer.html
```diff
@@ -0,0 +1,6 @@
+<script>
+    window.dataLayer = window.dataLayer || [];
+    {% if data_json %}
+    window.dataLayer.push({{ data_json|safe }});
```
danbentley, line 4: I think I have a preference for [`json_script`](https://github.com/django/django/blob/7119f40c9881666b6f9b5cf7df09ee1d21cc8344/django/utils/html.py#L62) here.

Some records properties contain markup (see `C1383853`) and I can see a scenario where the `dataLayer` and  is broken because we're not escaping `<`, `>`, or `&`.

Worse-case scenario testing here, I've manually set:

```python
self.availability_delivery_condition = '</script><script>window.alert(window);</script>'
```
Screenshot 2022-03-04 at 11 25 31

danbentley:
Thanks for this! 

Moving dataLayer configuration to the models and out of content-specific template tags feels much cleaner and extensible going forwards. 

Also, huge thanks also for beginning the process of re-organising the templates.

danbentley (changes requested):

templates/includes/gtm-datalayer.html
ababic: Cheers @danbentley. From what I can tell, we can do all of the same escaping that `json_script` does, without the complication of using multiple script tags or Django tags/filters. Would you be okay with a fix that addressed things in that way?

danbentley (approved):

PR #240, opened 2022-03-05
Un 136a

mozaki1:
Create topic,time and tags for the insight page:

- Created Migrations.
- Add tap migration to populate the taggit table with the new values.
- Add foreign field to the insight model, related to the time and topic collections 
- Add snippet so to limit the tags choices to those stored in the database


mozaki1:
Close pull request due to conflicts issues

PR #241, opened 2022-03-05
Un 136a

mozaki1:
- Add migrations for topic,time and tags
- Create custom migration to populate taggit.
- Add snippet so to populate tags from the database


etna/insights/models.py
```diff
@@ -46,12 +59,29 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
     body = StreamField(InsightsPageStreamBlock, blank=True, null=True)
-
+    topic = models.ForeignKey(
+        "collections.TopicExplorerPage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+    time = models.ForeignKey(
```
ababic, line 69: Maybe `time_period` would be a more obvious name for this field?

etna/insights/models.py
```diff
@@ -30,14 +32,25 @@ def get_context(self, request):
...
+    content_object = ParentalKey(
+        "insights.InsightsPage", on_delete=models.CASCADE, related_name="tagged_items"
+    )
+    free_tagging = False
+
+    class Meta:
+        verbose_name = "name"
+        verbose_name_plural = "slug"
+
+
```
ababic, line 53: You've defined a new 'through model' here, but I was expecting there to also be a custom `Tag` subclass too, [as [demonstrated  in the example here](https://docs.wagtail.org/en/stable/reference/pages/model_recipes.html#custom-tag-models). It is the custom `Tag` model that should have `free_tagging = False` applied to it to disable the free tagging, rather than the through model.

etna/insights/migrations/0033_auto_20220304_2226.py
```diff
@@ -0,0 +1,411 @@
+# Generated by Django 3.2.12 on 2022-03-04 22:26
+
+
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    # We get the model from the versioned app registry;
+    # if we directly import it, it'll be the wrong version
+    tag = apps.get_model("taggit", "Tag")
```
ababic, line 10: This tag model is actually used by everything... I would expect this set of tags to be added to their own model, keeping them separate from the others.

etna/insights/migrations/0033_auto_20220304_2226.py
```diff
@@ -0,0 +1,411 @@
...
+    db_alias = schema_editor.connection.alias
+    tag.objects.using(db_alias).bulk_create(
+        [
+            tag(name="Africa", slug="Africa"),
+            tag(name="Aid and development", slug="Aid and development"),
```
ababic, line 15: Slugs in Django usually consist of lower-case letters, with words separated by hyphens. There's a util function you can use that will take care of conversion for you: [django.utils.text.slugify](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.text.slugify)

Using this method you could write something like this:
```
from django.utils.text import slugify

...

to_create = [
    tag(name=v, slug=slugify(v)) for v in (
        "Africa",
        "Aid and development",
        "Air Force",
        ....
    )
]
tag.objects.bulk_create(to_create)
```

etna/insights/migrations/0033_auto_20220304_2226.py
```diff
@@ -0,0 +1,411 @@
...
+        ]
+    )
+
+
+def reverse_func(apps, schema_editor):
+    # forwards_func() creates tags instances,
+    # so reverse_func() should delete them.
+    tag = apps.get_model("taggit", "Tag")
+    db_alias = schema_editor.connection.alias
+    tag.objects.using(db_alias).filter(name="Africa", slug="Africa").delete()
```
ababic, line 159: If these tags were added to a separate model, all you would need to do in the reverse function is:
```
InsightsTag.objects.all().delete()
```

etna/insights/models.py
```diff
@@ -18,9 +23,6 @@ class InsightsIndexPage(TeaserImageMixin, BasePage):
 
     sub_heading = models.CharField(max_length=200, blank=False)
     body = StreamField(InsightsIndexPageStreamBlock, blank=True, null=True)
-    featured_insight = models.ForeignKey(
-        "insights.InsightsPage", blank=True, null=True, on_delete=models.SET_NULL
-    )
```
ababic, line 23: You appear to have removed a field here that was added recently. This will need to be reverted.

etna/insights/models.py
```diff
@@ -30,14 +32,25 @@ def get_context(self, request):
 
     content_panels = BasePage.content_panels + [
         FieldPanel("sub_heading"),
-        PageChooserPanel("featured_insight"),
```
ababic, line 33: This was added recently by myself for another ticket, and will need adding back in.

etna/insights/models.py
```diff
@@ -46,12 +59,29 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
+        related_name="+",
+    )
+    time = models.ForeignKey(
+        "collections.TimePeriodExplorerPage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+    tags = ClusterTaggableManager(through=InsightsPageTag, blank=True)
```
ababic, line 76: As mentioned in the dev notes: 
> The tag value will need to be indexed in order to make use of Wagtail’s search() feature for identifying ‘similar’ pages to display on each page. 

Basically, this means you'll need to add something like this:

```
    def get_insight_tag_names(self):
        return '\n'.join(self.tags.all().values_list('name', flat=True))

    search_fields = Page.search_fields + [
        index.SearchField('get_insight_tag_names'),
    ]
```

Here is the [relevant section of the docs](https://docs.wagtail.org/en/stable/topics/search/indexing.html#indexing-extra-fields).

templates/insights/insights_page.html
```diff
@@ -12,6 +12,14 @@
...
         {% include_block block %}
     {% endfor %}
+    {# Check if there are tags #}
+    {% if page.tags.count %}
+        <h1>Tags:</h1>
+        {% for tag in page.tags.all %}
+            {# Loop through all the existing tags #}
+            <a href="{{ self.get_parent.url }}?tag={{ tag.slug }}">{{ tag }}</a><br />
+        {% endfor %}
+    {% endif %}
```
ababic, line 22: See https://national-archives.atlassian.net/browse/UN-140 for details on what needs to be output. I think this HTML can be removed from here.

etna/insights/migrations/0036_remove_insightsindexpage_featured_insight.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ("insights", "0035_merge_20220305_1110"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="insightsindexpage",
+            name="featured_insight",
+        ),
+    ]
```
ababic, line 17: This migration will need to be removed

etna/insights/models.py
```diff
@@ -30,14 +32,25 @@ def get_context(self, request):
...
+@register_snippet
+class InsightsPageTag(TaggedItemBase):
+    content_object = ParentalKey(
+        "insights.InsightsPage", on_delete=models.CASCADE, related_name="tagged_items"
+    )
+    free_tagging = False
+
+    class Meta:
+        verbose_name = "name"
+        verbose_name_plural = "slug"
```
ababic, line 51: I can sense a bit of confusion here. These are labels for the model class. I would expect them to look something like:

```
    verbose_name = "insights tag"
    verbose_name_plural = "insights tags"
```

etna/insights/migrations/0036_remove_insightsindexpage_featured_insight.py
mozaki1: Thank you Andy , Removed now.

templates/insights/insights_page.html
mozaki1: Thank you Andy , I picked this one from the online tutorial by Kalob, It was only for testing but it found its way here.

etna/insights/models.py
mozaki1: Yes absolutely , done now

etna/insights/models.py
mozaki1: Yes, sorry for this, It was missed out during the merge, which had a few conflicts.

etna/insights/models.py
mozaki1: Yes , sorry about that, again merge issues.

etna/insights/models.py
mozaki1: This is now in a new custom model InsightsTag.

etna/insights/models.py
mozaki1: Yes, changed now.

etna/insights/migrations/0033_auto_20220304_2226.py
mozaki1: Yes , Tags for the insights now have their own model InsightsTag. it looks like this now:

tag = apps.get_model("insights", "InsightsTag")

etna/insights/models.py
```diff
@@ -46,12 +70,30 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+    tags = ClusterTaggableManager(through=TaggedInsights, blank=True)
     content_panels = (
         BasePage.content_panels
         + HeroImageMixin.content_panels
         + [
             FieldPanel("sub_heading"),
+            PageChooserPanel("featured_insight"),
```
ababic, line 93: I think this needs to be added to ``InsightsIndexPage.content_panels``, rather than ``InsightsPage.content_panels`` ?

mozaki1: Yes , the relationship is in the InsightsIndexPage via featured_insight foreign field, so the pagechooserpanel must be visible on the InsightsIndexPage admin page.

etna/insights/migrations/0033_auto_20220304_2226.py
mozaki1: Yes works nicely :)

etna/insights/migrations/0033_auto_20220304_2226.py
mozaki1: Yes , very useful, done now!

etna/insights/models.py
mozaki1: Yes, thank you, done now.

etna/insights/models.py
```diff
@@ -38,6 +45,26 @@ def get_context(self, request):
     subpage_types = ["insights.InsightsPage"]
 
 
+@register_snippet
```
ababic, line 48: Nice addition!

ababic (approved):
Thanks @mozaki1! This looks good now. I just want to give this latest iteration a spin locally to check a couple of things.

Suggestion for future: The `auto_` naming of migrations is convenient, but not great for helping other developers understand what changes are being made, or trying to address their own migration tree issues. Consider using the `-n` option next time to add a more explicit description

PR #243, opened 2022-03-09
UN-16X: Refactor Insights page blocks

AshTNA:
Resolves:
- https://national-archives.atlassian.net/browse/UN-163
- https://national-archives.atlassian.net/browse/UN-164
- https://national-archives.atlassian.net/browse/UN-162

Also changes the overall approach to content on Insights pages:

- 'Section' is the only valid block option at the top-level, which has a required 'heading' value that always renders as a `<h2>`.
- 'Sub-section' can be added to a 'Section' to nest content even further, which has a required 'heading' that always renders as a `<h3>`.
- Any block added to a 'Section' that has a 'heading' will automatically use a `<h3>`
- Any block added to a 'Sub-section' that has a 'heading' automatically use a `<h4>`
- Any sub-blocks of these auto-adjusting content blocks should automatically use the correct heading tag too (one level higher than the parent)

Included data migration to migrate existing content to new structure - However, due to the limitation of the current blocks, there are cases where invalid use of `<h3>` headings will change the order of content slightly. Realistically though, this is as good as it's going to get! All content will be revisited anyway - we just didn't want to lose everything entirely.


ababic:
@AshTNA would you be okay to review this?

TNA-Allan:
> On testing: What is it that we actually want to test? The base blocks and standard page editor interface are already tested by Wagtail. Maybe there's an argument for testing that the `get_context_data()` overrides behave as expected, but other than that, there's not really anything 'Custom' going on here. Testing that configurations of blocks work properly is a bit like testing built-in Django model fields:
> 
> * They behave according to how you configure them
> * You follow documented APIs at the time of configuring
> * If that API changes, Wagtail mention it in the release notes and bump the version to account for it
> * When you upgrade to the next version, you review the release notes and search for relevant code at that time
> * Things are very often 'deprecated' for a couple of versions before support is completely removed too, so you upgrade one version at a time and look out for warnings.

Thanks  @AshTNA @ababic for this info. I was not sure if those new blocks required test cases or not as its just new code added or as I understand now configured in this case.

JamesWChan:
Hi @AshTNA, @ababic

Thanks for this 👍 Just a few things:

- As @AshTNA mentions, we’ll need the `data-controlled-by` attribute on the `.section-content` rather than `.section-wrapper`.

- There’s a `data-section-level` attribute on `.section-wrapper` elements; what’s this used for?

- I believe we’d only need id’s on `h2` elements as that’s what the in-page nav links will use to jump to the relevant section. Id’s won’t be required on any other headings I don’t think.

ababic:
> There’s a data-section-level attribute on .section-wrapper elements; what’s this used for?

It's available in the template context, so thought it might be useful for targeting content of a specific level (for whatever reason). Also useful for debugging :)

> I believe we’d only need id’s on h2 elements as that’s what the in-page nav links will use to jump to the relevant section. Id’s won’t be required on any other headings I don’t think.

The sections and other elements all extend from a base block that makes them aware of the content depth and makes these values available in the context. The IDs are added to all h3s too simply because the same template is used for sections and sub-sections. If it is a problem that the ids are set for sub-section heading too, It's entirely possible to put a condition around that output in the template. But, if not, it simply doesn't make sense to complicate the template with unnecessary conditional.

JamesWChan:
Okay, let's leave the headings as they are then.

PR #245, opened 2022-03-09
Feature/bucketed search

danbentley:
It's a bit barebones at the moment but in the interests of getting the catalogue search form in front of people, this should be enough to perform a basic non-filtered, but bucketed search.

This is most of the progress made so far during the paired programming sessions.

Worth noting that at the moment, the template is outputting the raw ES result object.


Puththiran (approved):

TNA-Allan:
Looks good to me

PR #251, opened 2022-03-15
Integrate search frontend + backend

AshTNA:
## General tasks
- [ ] Add the text box 'Date' field to the search landing and figure out where it will sit on the results pages (similar to the one on Discovery https://discovery.nationalarchives.gov.uk/ Between YYYY and YYYY)

## Catalogue results frontend tasks

- [ ] Add JavaScript for show/hiding filters and buckets on mobile
- [ ] Figure out the HTML syntax and styling of error messages

## Potential backend tasks
- [ ]  The header menu elements will need `aria-current=true` if the page is a child of that menu item. This then displays it as selected via the CSS. It's currently hard coded so that the search menu item is always selected.
- [ ] If you’re not on  a search page, add a `data-id='search-nav-link` to the header menu search link, so that our JavaScript component can render global search by replacing that link.
- [ ] Add an `aria-current=true` to the corresponding search tab depending wether you’re on Featured results, catalogue results, or website results
- [ ] There is a description  paragraph for each results page. For example: `This page displays search results for records that are held at The National Archives, Kew.` with a shorter one for mobiles `Results for records held at The National Archives, Kew.` These will need to change based on the search tab and search bucket.
- [ ] Fix the "Records from The National Archives" bucket which is currently a duplicate of "Online records from The National Archives"
- [ ] The bucket links will also need an `aria-current=true` if it is the selected one. The bucket `<li>` elements will need a  `data-current=true` or `data-current=false` depending on if it is selected, for a JavaScript enhancement on mobile.
- [x] The sort by options will need to moved to a Django form, I've added a placeholder for now.
- [ ] The grid view/list view icons (next to the "Sort by" component) will need URLs, and a python variable to determine wether the grid results or list result card HTML is rendered.
- [ ] The “Show online records only” checkbox was missing in the filters so I’ve added a placeholder. I’m aware that this is a duplication of the “Online records from The National Archives” bucket, we will need to figure out from UX if this is staying
- [ ] The filters were missing the “Date” (century) checkboxes, so I’ve added a placeholder
- [ ] There needs to be a way for the labels inside a DynamicMultipleChoiceField to have a custom CSS class. This might be a bit complex but it would provide the benefit of the labels appearing clickable cursor wise, some small spacing adjustments, and helping us to be able to customise label css in the future.
- [ ] No separate collection/topic filter pages are created at the moment. I’m aware that this is because only 10 collections are appearing for an asterisk search currently. However discovery displays around 100 collections for the same search so I'm assuming this will need to be created. I'm aware this is a bit of a complex backend task as the whole search state will need to be persisted via a query string.
- [x] Missing the record opening date filter, so I’ve added a placeholder.
- [ ] The held by will need to be hidden if we’re on the TNA records only search
- [ ]  Result records will need a boolean to determine if they’re downloadable
- [ ] Result records will need a 150x150 thumbnail if it’s downloadable
- [ ] Result records will need the URL, catalogue reference, description, held by, and date available
- [ ] The export results URL will need to work eventually
- [ ]  Selected filters will need to show just above the results, with the links to remove the filters working
- [ ] The search_results_hero.html `<h2>` will need to be populated with the current search type (Catalogue/Featured/Website results)


AshTNA:
We're going to merge this PR to prevent merge conflicts down the line. The tasks above still need completion.

danbentley:
Thanks Ash!

As discussed, merging in to make merging a little smoother and for visibilty

PR #254, opened 2022-03-16
CHORE: Simplify API data transformation

ababic:
This PR begins to address some of the technical debt around the Kong API integration - by removing unnecessary components, defining clearer roles for remaining ones, and making the code more transparent. 

Key aims:
- Make the 'KongApiClient' return more useful values, so that the fake 'Manager' classes can be dropped
- Introduce a 'result parser' concept to make results more useful by default (making minimal changes to the original value), and provide a way to use a custom 'result parser' where needed.
- Show how the various views might be simplified to work with the api client directly


ababic:
Closing this for now, as this will need more work to bring in-line with other proposed changes, and maintaining that isn't worth it at this point in time. Will resurrect once we're in a more stable state.

PR #256, opened 2022-03-16
Add logic to highlight selected bucket

danbentley:
From DF-158:

> The bucket links will also need an `aria-current=true` if it is the selected one. The bucket `<li>` elements will need a  `data-current=true` or `data-current=false` depending on if it is selected, for a JavaScript enhancement on mobile.


AshTNA (approved):
All good - thanks Dan :+1:

PR #260, opened 2022-03-18
Incorporate search landing HTML/SASS

AshTNA:
Hi @danbentley 

Here's the rest of the search landing HTML. I've put the buckets into an include so there's another task of making the  bucket content dynamic to add to the list!

Would you be able to merge this?

Thanks :+1:


danbentley:
Looks great. Thanks Ash!

PR #263, opened 2022-03-21
Df 74 tna catalogue

TNA-Allan:
Changes for datalayer tags for TNA Catalogue pages
Updated for open beta response.


ababic:
Hi @TNA-Allan. The changes here look fine to me. Excellent work on tracking down/fixing the transformation errors.

etna/records/models.py
```diff
@@ -63,9 +72,19 @@ def availability_condition_category(self) -> str:
     def get_gtm_content_group(self) -> str:
         """
         Overrides DataLayerMixin.get_gtm_content_group() to
-        return the name of the class.
+        return content group otherwise the name of the class.
         """
-        return self.__class__.__name__
+        if (
+            self.catalogue_source == "CAT"
+            and self.repo_summary_title == "The National Archives"
```
ababic, line 79: @TNA-Allan is this second condition necessary? What are the other potential values for `catalogue_source == "CAT"`?

etna/ciim/tests/test_models.py
```diff
@@ -329,7 +364,7 @@ def test_record_with_origination_but_no_date(self):
         record = create_record(
             iaid="C123456",
         )
-        del record["_source"]["@origination"]["date"]
+        del record["_source"]["origination"]["date"]
```
ababic, line 367: @TNA-Allan has this changed for good in the API?

etna/ciim/tests/test_models.py
```diff
@@ -276,6 +277,7 @@ def test_related_records(self):
             ],
         )
 
+    @unittest.skip("Data not supported for the json record")
```
ababic, line 280: Should this test be deleted, rather that disabled?

etna/ciim/tests/test_models.py
```diff
@@ -252,18 +252,19 @@ def test_next_record(self):
...
     def test_previous_record(self):
         self.assertEqual(
             self.record.previous_record,
             {
-                "iaid": "C441748",
+                "iaid": "C10296",
             },
         )
 
+    @unittest.skip("Data not supported for the json record")
```
ababic, line 267: Should this test be deleted, rather that disabled?

ababic (changes requested):
Hi @TNA-Allan. A few question marks around the tests. Essentially, you've switched the example record out for one of a different type, but we still want coverage for all of the things we had before. Instead of disabling tests, we should have separate test cases for different types of example record. Skiptest is not designed to be used in this way.

etna/ciim/tests/test_models.py
TNA-Allan: The response now received is without @ symbol.

TNA-Allan:
> Hi @TNA-Allan. A few question marks around the tests. Essentially, you've switched the example record out for one of a different type, but we still want coverage for all of the things we had before. Instead of disabling tests, we should have separate test cases for different types of example record. Skiptest is not designed to be used in this way.

The example record prior to this one was customized to fit all the test cases in ModelTranslationTest . Currently the Kong API response is not yet returning those values for which the tests are skipped.

A similar usage of skiptest is in TestRecordView.test_record_renders_for_record_with_no_image, test_record_renders_for_record_with_image

etna/records/models.py
TNA-Allan: Does not look necessary, but I am confirming.

ababic:
@TNA-Allan Okay, I'm willing to accept the loss of coverage, provided it is documented somewhere that we come back to this. Maybe you could ask one of the managers to create a ticket for specifically looking at test coverage on the API stuff.

etna/records/models.py
TNA-Allan: There are other potential values from CAT. However removed condition as I have confirmed that CAT will apply to all TNA. Also updated the spreadsheet Custom Dimension for catalogue pages accordingly.

TNA-Allan:
> @TNA-Allan Okay, I'm willing to accept the loss of coverage, provided it is documented somewhere that we come back to this. Maybe you could ask one of the managers to create a ticket for specifically looking at test coverage on the API stuff?

Thank you Andy. Noted this and will action.

etna/records/models.py
TNA-Allan: This change is done

ababic (approved):
LGTM :)

PR #274, opened 2022-03-28
CHORE: Upgrade Wagtail to v2.16.1

ababic:
2.16 adds a tonne of improvements, and there appears to be very little in the way of backwards incompatible changes that impact Etna.

I also need to write something that extends `ListBlock` in a way that would have to be rewritten if we updated later - hence pushing for an update sooner rather than later.


Puththiran:
Wow that looks quite easy :), probably verifying it's not breaking anything is challenging.

ababic:
> probably verifying it's not breaking anything is challenging.

Not if there's a basic unit test for anything "custom" :) I know it might feel like we're doing a lot of custom things in this project, but really, it's all pretty clean and tidy, and overriding things in the documented way.

ababic:
Thanks very much @TNA-Allan!

PR #275, opened 2022-03-29
CHORE: Upgrade wagtail third-party apps (and Black)

ababic:
Whilst the current versions of wagtailmedia and wagtail-generic-chooser appear to be compatible with Wagtail 2.16, there are more recent releases, so we might as well take this opportunity to upgrade.

Also updating Black to fix a known compatibility issue https://github.com/psf/black/issues/2964 which is breaking CI today.


TNA-Allan (approved):
Thank you Andy

PR #278, opened 2022-03-30
Feature/featured search results

AshTNA:
Hi @ababic 

Thanks for the help with the featured search results.

Would you be able to check that the Python code is OK? 

We do have that issue with unicode characters e.g. `\n` in the interpretive content descriptions so if you're able to fix that that would be great but I can investigate if needed.

Thanks :+1:

### DIFF
--- a/templates/search/blocks/featured-search__interpretive-card.html
+++ b/templates/search/blocks/featured-search__interpretive-card.html
@@ -1,22 +1,29 @@
-<div>
-    <a href="#" aria-hidden="true" tabindex="-1" class="featured-search__interpretive-card-link">
-        <div>
-            <picture class="featured-search__interpretive-card-image">
-                <source media="(max-width: 768px)" srcset="https://via.placeholder.com/508x304">
-                <source media="(max-width: 991px)" srcset="https://via.placeholder.com/268x160">
-                <source media="(max-width: 1199px)" srcset="https://via.placeholder.com/348x208">
-                <source media="(min-width: 1200px)" srcset="https://via.placeholder.com/348x208">
-                <img src="https://via.placeholder.com/348x208" alt="" class="featured-search__interpretive-card-image-fallback"> 
-            </picture>
-        </div>
-    </a>
-        <div>
-            <a href="#" class="featured-search__results-link">
-                <h3 class="featured-search__heading">
-                    Card title
-                </h3>
-            </a>
-            <p class="featured-search__interpretive-card-description">Card description</p>
-        </div>
+{% load search_tags %}
+{% load humanize %}
 
-</div>
+{% if result.hits.hits %}
+    {% with record=result.hits.hits.0 %}
+        <div>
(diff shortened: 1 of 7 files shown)


templates/search/featured_search.html
```diff
@@ -11,49 +11,28 @@
     <p class="search-results__explainer--desktop">This page displays a summary of catalogue and non-catalogue content that matches your search.</p>
     <p class="search-results__explainer--mobile">Catalogue and non-catalogue content that matches your search.</p>
     <div class="featured-search__results">
-        {% with heading='Records from The National Archives' group="group:tna" %}
+        {% with results=result_groups.tna%}
             {% include './blocks/featured-search__results-block.html'%}
         {% endwith %}
-        {% with heading='Records from other UK archives' group="group:nonTna" %}
+        {% with results=result_groups.nonTna%}
```
ababic, line 17: Tiny nitpick: It makes things a little easier to scan if there's consistant spacing between tag contents and the end of the tag (`%}`)

templates/search/featured_search.html
```diff
@@ -11,49 +11,28 @@
     <p class="search-results__explainer--desktop">This page displays a summary of catalogue and non-catalogue content that matches your search.</p>
     <p class="search-results__explainer--mobile">Catalogue and non-catalogue content that matches your search.</p>
     <div class="featured-search__results">
-        {% with heading='Records from The National Archives' group="group:tna" %}
+        {% with results=result_groups.tna%}
```
ababic: Nice use of `with` here! When 'variabalizing' things specifically for an include to pick up, it can be cleaner to do:
```
{% include 'template_name' with results=results_groups.tna %}
```

templates/search/featured_search.html
AshTNA: Thanks @ababic - fixed now :+1:

templates/search/blocks/featured-search__interpretive-card.html
```diff
@@ -1,22 +1,33 @@
...
-                <source media="(min-width: 1200px)" srcset="https://via.placeholder.com/348x208">
-                <img src="https://via.placeholder.com/348x208" alt="" class="featured-search__interpretive-card-image-fallback"> 
-            </picture>
-        </div>
-    </a>
+{% load search_tags %}
+{% load humanize %}
+
+{% if result.hits.hits.0 %}
+    {% with record=result.hits.hits.0 %}
```
ababic, line 5: Template logic definitely shouldn't be having to contend with elasticsearch-like structures... the API client should be containing that mess to stop it leaking out into the rest of the codebase. We can't do anything about that now though - later maybe :)

templates/search/featured_search.html
AshTNA: Thanks @ababic - I've changed them to reflect this now

ababic (approved):

PR #279, opened 2022-03-30
Df 208-Update client and form to support /search endpoint's new date parameters

TNA-Allan:
Update client and form to support /search endpoint's new date parameters
Removed DateField, startDate, endDate, buckets, topics, references
Added openingStartDate,openingEndDate,closingStartDate,closingEndDate

### DIFF
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -168,9 +168,6 @@ class KongClient:
         aggregations: Optional[list[Aggregation]] = None,
         filter_aggregations: Optional[list[str]] = None,
         filter_keyword: Optional[str] = None,
-        buckets: Optional[list[str]] = None,
-        topics: Optional[list[str]] = None,
-        references: Optional[list[str]] = None,
         offset: Optional[int] = None,
         size: Optional[int] = None,
     ) -> dict:
@@ -202,12 +199,6 @@ class KongClient:
             filter results set by aggregation
         filter_keyword:
             filter results by keyword
-        buckets:
-            Restrict results to given bucket(s)
-        topics:
-            Restrict results to given topic(s)
-        references:
-            Restrict results to given reference(s)
         offset:
             Offset for results. Mapped to 'from' before making request
         size:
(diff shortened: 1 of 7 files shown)


etna/ciim/client.py
```diff
@@ -164,19 +157,17 @@ def search(
         *,
         q: Optional[str] = None,
         web_reference: Optional[str] = None,
-        start_date: Optional[datetime] = None,
-        end_date: Optional[datetime] = None,
-        date_field: Optional[DateField] = None,
+        opening_start_date: Optional[datetime] = None,
+        opening_end_date: Optional[datetime] = None,
```
ababic, line 161: If the client expects a `datetime`, but the user only selects a `date` - we'll probably need to turn the 'end dates' into 'end of the day' datetimes to have the desired affect (e.g. `2021-03-30-23:59:99`). This will apply to both `opening_end_date` and `closing_end_date`

TNA-Allan: > If the client expects a `datetime`, but the user only selects a `date` - we'll probably need to turn the 'end dates' into 'end of the day' datetimes to have the desired affect (e.g. `2021-03-30-23:59:99`). This will apply to both `opening_end_date` and `closing_end_date`

Nice one Andy. Agree, but current ticket scope is for the parameter changes, thought to address this in another ticket DF-209 where date values are input individually.

etna/search/forms.py
```diff
@@ -115,13 +115,13 @@ class BaseCollectionSearchForm(forms.Form):
         ),
         required=False,
     )
-    start_date = forms.DateTimeField(
+    opening_start_date = forms.DateTimeField(
```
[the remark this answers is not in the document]
TNA-Allan: > Are they inputting a `date` or a `datetime`? Would a `DateField()` be more appropriate here?

API spec 1.0.0-SNAPSHOT (2022-03-23 09:25) says string($date-time), but the input from user is only date. Handling of time 00:00:00 to 23:59:59 can be done in the DF-209

etna/ciim/client.py
ababic: Sensible :) Works for me!

etna/search/forms.py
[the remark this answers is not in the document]
ababic: Cool beans :)

ababic (approved):
LGTM. Nice work @TNA-Allan :)

TNA-Allan:
> LGTM. Nice work @TNA-Allan :)

Thank you Andy

PR #282, opened 2022-04-01
DF-74: Custom data for analytics service (amends)

TNA-Allan:
- Datalayer changes for testing comments
- Included changes for digitized detail record, put TODO commented code in records.py - also relates to DF-220


etna/records/models.py
```diff
@@ -49,10 +49,10 @@ class Record(DataLayerMixin, APIModel):
     repo_archon_value: str = ""
     level_code: str = ""
     level: str = ""
-    association_reference_number: str = ""
-    association_summary_title: str = ""
```
ababic, line 53: I can confirm these are not referenced anywhere other than in tests

etna/records/models.py
```diff
@@ -49,10 +49,10 @@ class Record(DataLayerMixin, APIModel):
...
+    template_reference_number: str = ""
     template_summary_title: str = ""
-    data_source: str = ""
```
ababic, line 55: I can confirm this not referenced anywhere other than in tests

etna/records/tests/test_views.py
```diff
@@ -211,11 +211,126 @@ def test_record_renders_for_record_with_image(self):
...
+class TestDataLayerRecordDetail(TestCase):
+    def setUp(self):
+        private_beta_user = User(
+            username="private-beta@email.com", email="private-beta@email.com"
+        )
+        private_beta_user.set_password("password")
+        private_beta_user.save()
+        private_beta_user.groups.add(Group.objects.get(name="Beta Testers"))
+
+        self.client.login(email="private-beta@email.com", password="password")
```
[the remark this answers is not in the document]
TNA-Allan: 1. The page is for the end-user like Beta like Tester
2. Keeping is consistent with other test which use the same view.

TNA-Allan:
Thank you Andy for appropriate - renaming title.

etna/records/models.py
TNA-Allan: These fields are not in tests

etna/records/models.py
```diff
@@ -91,20 +91,49 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
         Override this method on subclasses to add data that is relevant to a
         specific record type.
         """
+
+        def get_customdimension11(self):
```
ababic, line 95: How would you feel about making these property methods and sticking with python's underscored naming as a rule? e.g:

```
@property
custom_dimension_11(self):
    pass
```

ababic (changes requested):
Just a couple of small things, but otherwise looking great. Thanks @TNA-Allan! It's great to have these fixtures in place to test the different levels. However, we'll need to be sure to make time to go through them all if further changes are made to the API.

etna/records/tests/test_views.py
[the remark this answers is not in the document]
ababic: Sure, but the 'page being protected' is really separate and temporary issue. These tests are for testing the view logic (not the auth protection logic), so the more generic you can keep this logic, the better.

etna/records/models.py
TNA-Allan: Thanks @ababic Done this change.

etna/records/tests/test_views.py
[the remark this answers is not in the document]
TNA-Allan: The view requires only a valid login, nothing specific for any kind of a user. Please see I've used self.login() instead. Would this help?

etna/records/models.py
```diff
@@ -83,6 +83,41 @@ def get_gtm_content_group(self) -> str:
         else:
             return self.__class__.__name__
 
+    @property
+    def _customdimension11(self) -> str:
```
ababic, line 87: Do you not like the format `custom_dimension_11`?

etna/records/tests/test_views.py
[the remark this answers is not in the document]
ababic: Yes, that should do the job :)

etna/records/models.py
TNA-Allan: Done :)

ababic (approved):
LGTM

PR #296, opened 2022-04-13
Fix/bucket count

AshTNA:
Hi @ababic 

This fixes the issues with bucket count and the buckets not showing on the search landing page.

Thanks :+1:


ababic (approved):
Thanks @AshTNA. This resolves the issue for me locally.

PR #297, opened 2022-04-14
DF-228: Fix 0 results showing for 'Find an archive' bucket

ababic:
Resolves ticket: https://national-archives.atlassian.net/browse/DF-228


etna/search/views.py
```diff
@@ -161,9 +163,11 @@ def catalogue_search(request):
                 Aggregation.CATALOGUE_SOURCE,
                 Aggregation.CLOSURE,
                 Aggregation.COLLECTION,
-                Aggregation.GROUP,
                 Aggregation.LEVEL,
                 Aggregation.TOPIC,
+                # Fetching more groups so that we receive a counts
+                # for any bucket/tab options we might be showing
+                f"{Aggregation.GROUP}:30",
             ],
```
TNA-Allan, line 171: Can we have it as a constant like 'AGGREGATION_SIZE' used in catalogue_search_long_filter_chooser than hardcode?

etna/search/views.py
```diff
@@ -294,9 +298,11 @@ def website_search(request):
...
```
TNA-Allan, line 306: Can we have it as a constant like 'AGGREGATION_SIZE' used in catalogue_search_long_filter_chooser than hardcode?

TNA-Allan (changes requested):

TNA-Allan:
Branch name convention does not contain ticket df-228

etna/search/views.py
ababic: I'm not sure that's worth it, personally? I think the comment does a good enough job of explaining what's happening - and using a constant in just this one case and not the others, where we're relying on the default (10)

ababic:
> Branch name by convention does not contain ticket df-228

My bad - shouldn't be a problem though, as the branch will be deleted after merging.

etna/search/views.py
TNA-Allan: Somewhat agree, but I will leave this to your best judgment :)

TNA-Allan (approved):
Looks good to me

etna/search/views.py
ababic: In this case, I think including 30 directly is slightly clearer.

PR #303, opened 2022-04-19
DF-229: Disable validation for 'collection' field in search

ababic:
Even though they are useful as a label source, the choice values we have for the `collection` field are not complete enough to be used for validation... a selected collection shouldn't be considered invalid just because it isn't in the hardcoded choice list.


jamesbiggs (approved):
All works fine and the changes to the filter texts have been made.

PR #304, opened 2022-04-20
Fix: Incorrect image block template name

ababic:


jamesbiggs (approved):
Fixes the template location

PR #306, opened 2022-04-20
DF-238: Filter held by

TNA-Allan:


etna/ciim/client.py
```diff
@@ -212,7 +231,9 @@ def search(
             "sortOrder": sort_order,
             "template": template,
             "aggregations": format_list_param(aggregations),
-            "filterAggregations": format_list_param(filter_aggregations),
+            "filterAggregations": format_list_param(
+                filter_aggregations, "filter_aggregations"
+            ),
```
ababic, line 236: Hi @TNA-Allan. As per the conversation on https://national-archives.atlassian.net/browse/DF-238, we want to remove the `format_list_param` function completely and just let `requests` handle the list of values as per the example at end of https://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls

etna/ciim/client.py
```diff
@@ -89,12 +93,27 @@ def format_list_param(items: Optional[list]) -> Optional[str]:
...
+        regex = r"([():\\/,&])"
+        subst = " "
+        items = [
+            i.split(":", 1)[0]
+            + ":"
+            + re.sub(regex, subst, i.split(":", 1)[1], 0, re.MULTILINE)
+            for i in items
+        ]
+    else:
+        items = [i.replace(",", "") for i in items]
```
ababic, line 114: With lists of values represented as `?param=1&param=2&param=3`, commas in values should no longer be an issue, so we can turn this function into a dedicated `prepare_filter_aggregations()` function.

etna/ciim/client.py
```diff
@@ -89,12 +93,27 @@ def format_list_param(items: Optional[list]) -> Optional[str]:
...
+        # reconstruct the required starting value that contains special value
+        # replace special chars by space
+        regex = r"([():\\/,&])"
+        subst = " "
+        items = [
+            i.split(":", 1)[0]
+            + ":"
+            + re.sub(regex, subst, i.split(":", 1)[1], 0, re.MULTILINE)
+            for i in items
+        ]
```
ababic, line 112: Do these transforms need to be made to every value, or just "heldBy" values?

ababic (changes requested):
@TNA-Allan please see the feedback below

etna/ciim/client.py
TNA-Allan: Every Value including heldBy

etna/ciim/client.py
TNA-Allan: There are issues found with the API response and is recorded in the same ticket. Hence created draft PR not to be merged.

etna/ciim/client.py
TNA-Allan: New changes, only heldBy values

etna/ciim/client.py
TNA-Allan: Done

etna/ciim/client.py
```diff
@@ -73,47 +73,40 @@ class Aggregation(str, enum.Enum):
...
-    else:
-        items = [i.replace(",", "") for i in items]
+    regex = r"([/():,&-])"
+    subst = " "
+    field_list_to_prepare = ["heldBy"]
+    filter_prepared_list = []
+
+    for item in items:
+        split_field_value = item.split(":", 1)
+        field, value = split_field_value[0], split_field_value[1]
```
ababic, line 102: You could shorten this to:

```suggestion
        field, value = item.split(":", 1)
```

TNA-Allan: Done TY :)

TNA-Allan: Done TY :)

etna/ciim/client.py
```diff
@@ -69,32 +70,42 @@ class Aggregation(str, enum.Enum):
...
+    subst = " "
+    field_list_to_prepare = ["heldBy"]
+    filter_prepared_list = []
+
+    for item in items:
+        field, value = item.split(":", 1)
+        if field in field_list_to_prepare:
+            filter_prepared = field + ":" + re.sub(regex, subst, value, 0, re.MULTILINE)
+        else:
+            filter_prepared = field + ":" + value
```
ababic, line 105: I think this is the right choice for the client code... Even though ':' can be problematic elsewhere, that is definitely something that should be handled by the API.

ababic (approved):
LGTM! Thanks @TNA-Allan

PR #307, opened 2022-04-21
DF-248:Make-catalogue-search-results-count-reflect-current-pagination

TNA-Allan:
1. Added page index
2. Removed bold/highlight settings


jamesbiggs (approved):
The pagination works on all pages

PR #308, opened 2022-04-21
UN 141 Update analytics service to ‘pick up’ keywords and push the information to Google Analytics dataLayer

jamesbiggs:


ababic:
Hi @jamesbiggs. You might want to try running `fab format` locally... it'll take care of any isort/black formatting issue for you :)

etna/insights/models.py
```diff
@@ -105,6 +106,31 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
         index.SearchField("insight_tag_names"),
     ]
 
+    def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
+        data = super().get_datalayer_data(request)
+        if customDimension4 := self.topic:
+            pass
+        else:
+            customDimension4 = ""
```
ababic, line 114: When the variable is something you want to access outside of the immediate conditional, there are clearer ways to write this sort of thing. For example, if there were more than two potential options, you could write like this:

```
some_variable = ""
if some_conditional:
    some_variable = some_value
elif some_other_conditional:
    some_variable = some_other_value
``` 

However, when the conditional logic is as simple as this, it's more readable to write things in a single line using an in-line conditional. For example:

```suggestion
        customDimension4 = self.topic.title if self.topic else ""
```

Also note: I'm also explicitly using `self.topic.title` as the value in the above example, as that is the string we want to display.

ababic: Since `data` already has a default value for all dimensions, you could simplify this a bit, so that you don't need the `data.update()` at the end... For example:
```suggestion
        if self.topic:
            data["customDimension4"] = self.topic.title
```

etna/insights/models.py
```diff
@@ -108,27 +108,15 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
-            customDimension5,
-            customDimension6,
-            customDimension7)
+        if self.topic:
+            data["customDimension4"] = self.topic.title
+        if self.topic:
+            data["customDimension5"] = self.sub_heading
+        if self.insight_tag_names:
+            semicolon_insight_tag_names = ";".join(self.insight_tag_names.split("\n"))
+            data["customDimension6"] = semicolon_insight_tag_names
```
ababic, line 117: We might as well do this in one line:
```suggestion
            data["customDimension6"] = ";".join(self.insight_tag_names.split("\n"))
```

etna/insights/models.py
```diff
@@ -108,27 +108,15 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
+        if self.time_period:
+            data["customDimension7"] = self.time_period
```
ababic, line 119: `time_period` is another `ForeignKey` to a different page model, so we do grab the page title here too, e.g.:
```suggestion
            data["customDimension7"] = self.time_period.title
```

etna/insights/models.py
```diff
@@ -108,27 +108,15 @@ class InsightsPage(HeroImageMixin, TeaserImageMixin, BasePage):
...
-            customDimension7 = ""
-        data.update(
-            customDimension4,
-            customDimension5,
-            customDimension6,
-            customDimension7)
+        if self.topic:
+            data["customDimension4"] = self.topic.title
+        if self.topic:
+            data["customDimension5"] = self.sub_heading
```
ababic, line 114: We can safely remove these 2 lines. The requirements are for "customDimension5" to show the 'subTopic' for a page, but we don't have a suitable value to display for this.

ababic (changes requested):
@jamesbiggs A bit more feedback for you below chap...

ababic (approved):

PR #310, opened 2022-04-22
DF-259: Consolidate search view logic into generic views

ababic:
Resolves: https://national-archives.atlassian.net/browse/DF-259

### Summary of changes:

- All search views are now class-based, and individual views simply override the thing they need to instead of redefining everything.
- Splitting the view logic up across multiple methods means that logic is easier to override in special cases, and also makes each part easier to explain and understand.
- I've extended our concept of 'Buckets', so that the objects in the context have everything they need in order to render themselves as needed. They can even contain the relevant lists of results in `FeaturedSearchView`... simplifying the data structure included in the context for that view.
- The flow of data from view -> form and back again is clearer.... default values are defined more clearly and always go through the form.
- Validation has been added to the `per_page` parameter so that only values between `20` and `200` are accepted (which can be easily adjusted). There was nothing there before to prevent users supplying a stupidly high number and putting undue pressure on both the app and API.
- Providing an invalid value for `group`, `per_page`, `sort_by`, `sort_order` or `display` will now result in a HTTP 400 (Bad Request) response.
- As much custom functionality as possible has been moved from forms, templates and template tags to the views, where it's much easier to find and amend (better separation of concerns).
- The authentication requirement for search views is now controlled by the `SEARCH_VIEWS_REQUIRE_LOGIN` setting, allowing it to be easily turned off in local development and tests (No need to create users / authenticate any longer)


templates/search/blocks/featured-search__record-creator.html
```diff
@@ -1,28 +1,23 @@
...
             <h3 class="featured-search__results-heading"><a href="#" class="featured-search__results-link">{{record|record_detail:"summaryTitle"|striptags }}</a></h3>
         </div>
-        {% endwith %}
-    {% endif %}
+    {% endfor %}
 
     <div class="featured-search__see-all-link">
-        <a class="featured-search__results-link" href="{% url 'search-catalogue' %}?q={{ form.q.value }}&group={{ results.aggregation }}">
-            See all {{ results.bucket.label }} ({{ results.hits.total.value|intcomma }})
+        <a class="featured-search__results-link" href="{% url 'search-catalogue' %}?q={{ search_query|iriencode }}&group={{ bucket.key }}">
```
gtvj, line 19: I've not come across the `iriencode` filter before. Glad to learn about it.

templates/search/blocks/featured-search__record-creator.html
```diff
@@ -1,28 +1,23 @@
 {% load search_tags %}
-{% load humanize %}
 <div class="featured-search__results-block">
-    <h2 class="featured-search__heading">{{ results.bucket.label }}</h2>
-    
-    {% if results.hits.hits %}
+    <h2 class="featured-search__heading">{{ bucket.label }}</h2>
 
-        {% with record=results.hits.hits.0 %}
+    {% for record in bucket.results|slice:":1" %}
```
gtvj, line 5: Quick Python know-how question: am I correct to interpret this use of `|slide:":1"` to mean just the first item?

templates/search/blocks/featured-search__record-creator.html
```diff
@@ -1,28 +1,23 @@
...
             </div>
             <h3 class="featured-search__results-heading"><a href="#" class="featured-search__results-link">{{record|record_detail:"summaryTitle"|striptags }}</a></h3>
         </div>
-        {% endwith %}
-    {% endif %}
+    {% endfor %}
 
     <div class="featured-search__see-all-link">
-        <a class="featured-search__results-link" href="{% url 'search-catalogue' %}?q={{ form.q.value }}&group={{ results.aggregation }}">
-            See all {{ results.bucket.label }} ({{ results.hits.total.value|intcomma }})
```
gtvj, line 24: Also wasn't aware of the `intcomma` filter - very handy. Good to know it exists.

templates/search/blocks/featured-search__interpretive-card.html
```diff
@@ -13,21 +10,19 @@ <h2 class="featured-search__heading">{{result.bucket.label}}</h2>
                         <source media="(max-width: 991px)" srcset="https://via.placeholder.com/268x160">
                         <source media="(max-width: 1199px)" srcset="https://via.placeholder.com/348x208">
                         <source media="(min-width: 1200px)" srcset="https://via.placeholder.com/348x208">
-                        <img src="https://via.placeholder.com/348x208" alt="" class="featured-search__interpretive-card-image-fallback"> 
+                        <img src="https://via.placeholder.com/348x208" alt="" class="featured-search__interpretive-card-image-fallback">
```
gtvj, line 13: Bit strange that we've got `placeholder.com` placeholders here.

etna/search/views.py
```diff
@@ -1,172 +1,208 @@
...
-    page_number = int(request.GET.get("page", 1))
-    offset = (page_number - 1) * per_page
+    # The source data for get_buckets()
+    bucket_list: BucketList = None
+    # Can be updated by the view for get_current_bucket_key() to pick up
+    current_bucket_key: str = None
+
+    def extract_group_buckets(
+        self, api_result: Union[None, Dict[str, Any]]
+    ) -> Sequence[Dict[str, Union[str, int]]]:
```
gtvj, line 65: Quick Python know-how question. Am I correct to understand this definition to be: 

* Stipulating that `extract_group_buckets` expects an `api_result` argument with a type of [Union](https://docs.python.org/3/library/typing.html#typing.Union). What confuses me a little is the presence of `None` in the Union type - but I'm sure this is my ignorance. 
* Stipulating that it returns a Sequence of Dicts (with the key of each being a string and the value being either a `str` or `int`)

Is that right? It would also be great if you could help me understand the `Union[None, Dict[str, Any]]`

etna/search/views.py
```diff
@@ -177,192 +213,398 @@ def catalogue_search(request):
...
+        form.
+        """
+        return {}
+
+    def form_valid(self, form: Form) -> HttpResponse:
+        """
+        When the form is valid, fetch results from the API, take any actions
+        based on the result, then render everything to a template.
+        """
+        self.api_result = api_result = self.get_api_result(form)
```
gtvj, line 282: Quick Python know-how question: I'm not familiar with this double assignment syntax. Grateful if you could point me to some docs that will help me understand what's happening here when we have `x = y = z`

etna/search/views.py
```diff
@@ -177,192 +213,398 @@ def catalogue_search(request):
...
+        context = self.get_context_data(form=form)
+        return self.render_to_response(context)
+
+    def get_meta_title(self) -> str:
+        """
+        Return a string to use the the <title> tag for this view.
+        """
+        title = self.base_title
+        if query := self.form.cleaned_data.get("q", ""):
+            title += ' "' + query.replace('"', "'") + '"'
```
gtvj, line 293: Quick Python know-how question here. Is there a reason we wouldn't use a [f-string](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) here?

gtvj (approved):

etna/ciim/constants.py
```diff
@@ -1,17 +1,34 @@
...
+from django.contrib.humanize.templatetags.humanize import intcomma
 
 
 @dataclass
 class Bucket:
     key: str
     label: str
+    result_count: int = None
+    is_current: bool = False
+    results: List[Any] = None
```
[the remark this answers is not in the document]
ababic: Hi @gtvj . Yes, exactly :) When used, the value for `results` is actually a list of dictionaries, but there's nothing about the design of `Bucket` that requires certain types of value... just that it is a list.

etna/search/forms.py
```diff
@@ -210,46 +204,10 @@ def clean(self):
                 )
         except TypeError:
             # Either one or both date fields are empty. No further validation necessary.
-            ...
```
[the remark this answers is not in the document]
ababic: I've never seen `...` or used it myself, hence changing it to something a bit more obvious. Flake8 seemed not to flag it as invalid though, which is kinda puzzling :S

etna/search/templatetags/search_tags.py
```diff
@@ -13,17 +13,6 @@
...
-    try:
-        buckets = api_response["aggregations"]["group"]["buckets"]
-    except KeyError:
-        return 0
-
-    return next(iter(a["doc_count"] for a in buckets if a["key"] == bucket_name), 0)
-
-
 @register.filter
 def record_detail(record: dict, key: str) -> str:
```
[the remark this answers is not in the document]
ababic: Yes, that's exactly it. There's no requirement to use type hints in Python, and you can really be as specific or as lenient as you want to be (e.g. defining your own types to really nail down the valid ranges of return values). I really like using them in a light-touch way, just for the self-documenting nature, and for the benefits it brings to things like 'Intellisense' in modern editors.

templates/search/blocks/featured-search__record-creator.html
ababic: Yes, that's right. If we change the number of items we want to show at some point, this new format will be easier to change. It's worth bookmarking [this page of the Django docs](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/) if working a lot with Django. There are some very handy tidbits in there, including [slice](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#slice)

etna/search/views.py
ababic: `typing.Union` essentially means "any one of the enclosed types". So, in this case, the value is expected to be `None` in some circumstances, and a dictionary in others. However, it always must always be provided.

etna/search/views.py
ababic: It's basically just a way of defining more than one variable at a time from the same value. In this case, we want a convenient `api_result` variable to use in the same method, but we also want to set `self.api_result` so that it becomes part of the view object's 'state', and can be accessed from other methods.

The above code is a one-line equivalent to:

```
api_result = self.get_api_result(form)
self.api_result = api_result
```

etna/search/views.py
ababic: f-string syntax here was triggering a formatting issue in VSCode, because of what is being replaced.

ababic: I suppose this could be split up into multiple lines, with an f-string used to combine everything on the second line... but, I figured this was simple enough to do on one line.

etna/ciim/constants.py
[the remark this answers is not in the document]
ababic: Would highly recommend taking a look at: https://docs.python.org/3/library/dataclasses.html

Dataclasses are a really clever (and efficient) way of defining simple data structures like this. Under the hood, these type annotations actually get enforced when you try to create an instance of `Bucket`.. taking them beyond mere guidance/documentation

ababic:
@TNA-Allan All of the things noted earlier should be addressed now. Could I request a re-review?

TNA-Allan:
> @TNA-Allan All of the things noted earlier should be addressed now. Could I request a re-review?

@ababic Just one query, the bucket counts are not showing. Otherwise all Good.

etna/search/views.py
```diff
@@ -1,172 +1,211 @@
...
+                    bucket.is_current = True
+                    break
+
+        return bucket_list
+
+    def get_context_data(self, **kwargs):
+        if self.bucket_list:
+            group_buckets = self.extract_group_buckets(
+                getattr(self, "api_request", None)
+            )
```
TNA-Allan, line 131: I guess typo, xhanging from `api_request ` to `api_result` should show Bucket counts.

ababic: Thanks for the diagnosis @TNA-Allan! This should be fixed in 499e04c

PR #319, opened 2022-04-28
CHORE: Update {% record_url %} tag to better account for edge cases

ababic:
The tag is currently using the wrong URL parameter names, so this PR fixes that first of all. However, even when that problem is resolved, some record values simply don't match up to the format we're expecting, so this also updates the tag to handle those situations better.

We need to look much more in-depth at the URLs and accepted patterns in Etna and the API anyway, but that will happen separately.


AshTNA (approved):
All good :+1:

PR #325, opened 2022-05-05
Feature/update featured record fe

JamesWChan:
Hi @AshTNA,

This PR updates the styling for the featured record component and featured records component.  The featured record component now looks like this:

image

This component was using an old template (record-embed-no-image) so I've renamed it to something more suitable. 

And the featured records component looks like this:

image

Would it be possible to merge if all's okay? Thank you 👍


AshTNA (approved):
All good :+1:

PR #328, opened 2022-05-09
DF-123: Sentry integration amends

ababic:


config/versioning.py
```diff
@@ -2,12 +2,19 @@
...
 
     head_path = os.path.join(base_dir, ".git", "HEAD")
 
-    with open(head_path, "r") as fp:
-        head = fp.read().strip()
+    try:
+        with open(head_path, "r") as fp:
+            head = fp.read().strip()
+    except FileNotFoundError:
+        return ""
```
TNA-Allan, line 17: Could we mark this as temporary or maybe  get_git_sha?

TNA-Allan (commented):
One query, otherwise all looks good

TNA-Allan (commented):
One query, otherwise all looks good

config/versioning.py
ababic: There's no way to do this lazily, unfortunately - as the config is needed to initialise the sentry client. I think leaving it as a blank string if it cannot be identified is a safe enough fallback.

Perhaps when we move to AWS, there's be something similar we can get from the environment too.

PR #331, opened 2022-05-11
Fix: KongBadRequestError raised from RecordChooserBlock.bulk_to_python()

ababic:
Resolves Sentry issue: https://sentry.io/organizations/the-national-archives/issues/3264948158/?project=6377027

- Avoid calling fetch_all() if there are no `iaid` values to search with
- Rejig code to improve readability

### DIFF
--- a/etna/records/blocks.py
+++ b/etna/records/blocks.py
@@ -55,13 +55,15 @@ class RecordChooserBlock(ChooserBlock):
         return value.iaid
 
     def bulk_to_python(self, values):
-        """Return the model instances for the given list of primary keys.
+        """Return a list of model instances for the given list of primary keys.
         The instances must be returned in the same order as the values and keep None values.
         """
-        _, records = self.target_model.api.fetch_all(iaids=values)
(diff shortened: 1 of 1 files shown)


etna/records/blocks.py
```diff
@@ -55,13 +55,15 @@ def get_prep_value(self, value):
...
     def bulk_to_python(self, values):
-        """Return the model instances for the given list of primary keys.
+        """Return a list of model instances for the given list of primary keys.
         The instances must be returned in the same order as the values and keep None values.
         """
-        _, records = self.target_model.api.fetch_all(iaids=values)
-        record_dict = {r.iaid: r for r in records}
-        record_list = tuple(record_dict.get(iaid) for iaid in values)
-        return record_list
+        if not values:
```
mozaki1, line 61: The question is why does the function gets called if the values not true?

ababic: I would guess it has something to do with the state of the page... maybe there are instances of the block that have lost their values due to the other bug we fixed.

mozaki1 (approved):

PR #333, opened 2022-05-11
Added back end for trigger/content warnings on Insights page

jamesbiggs:


templates/insights/insights_page.html
```diff
@@ -4,6 +4,7 @@
 {% load wagtailcore_tags %}
 {% load sections_tags %}
 
+
```
ababic, line 7: No need for an extra blank line here :)

ababic (changes requested):
Have tested locally, and everything works as expected. Just a few minor linting/formatting things to fix

ababic (approved):

PR #334, opened 2022-05-11
Remove title from featured_record.html and remove decorative from ima…

mozaki1:
…ge block , make direct inheritance from blocks.StructBlock

### DIFF
--- a/templates/insights/blocks/featured_record.html
+++ b/templates/insights/blocks/featured_record.html
@@ -17,7 +17,7 @@
             <h4 class="featured-record__standfirst">{{ value.title }}</h4>
             
             {% if value.record.title %}
-                <p class="featured-record__title">{{ 'title'|as_label }}: {{ value.record.title }}</p>
+                <p class="featured-record__title">{{ value.record.title }}</p>
             {% endif %}
 
             <div class="featured-record__metadata">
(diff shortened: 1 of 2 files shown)


templates/insights/blocks/featured_record.html
```diff
@@ -17,7 +17,7 @@
             <h4 class="featured-record__standfirst">{{ value.title }}</h4>
             
             {% if value.record.title %}
-                <p class="featured-record__title">{{ 'title'|as_label }}: {{ value.record.title }}</p>
+                <p class="featured-record__title">{{ value.record.title }}</p>
```
ababic, line 20: @mozaki1 Did you discuss with @JamesWChan about maintaining the label for screen-reader users?

mozaki1: I did and he is happy with that.

JamesWChan: Hi @ababic, Mo did ask me and I said there wouldn't be any accessibility concerns if it was no longer required i.e. there had been a change request that I wasn't aware of that meant the title was being removed in favour of something curated by the editors. I wasn't aware of the context of this and I see that the title is being retained and so, yes, it might be beneficial to retain this.

ababic (changes requested):
Just one tiny thing @mozaki1...

ababic (approved):
Thanks for fixing @mozaki1! LGTM :)

PR #336, opened 2022-05-12
Fix: Transformation function erroring when encountering topic data in a different format

ababic:
Resolves Sentry issue: https://sentry.io/organizations/the-national-archives/issues/3260325174/

### DIFF
--- a/etna/records/transforms.py
+++ b/etna/records/transforms.py
@@ -81,12 +81,15 @@ def transform_record_result(result):
             data["availability_delivery_surrogates"] = delivery.get("surrogate")
 
     if topics := source.get("topic"):
-        data["topics"] = [
-            {
-                "title": i["summary"]["title"],
-            }
-            for i in topics
-        ]
+        data["topics"] = []
+        for i in topics:
+            topic_title = ""
+            try:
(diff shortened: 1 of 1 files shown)


etna/records/transforms.py
```diff
@@ -81,12 +81,15 @@ def transform_record_result(result):
...
-                "title": i["summary"]["title"],
-            }
-            for i in topics
-        ]
+        data["topics"] = []
+        for i in topics:
+            topic_title = ""
+            try:
+                topic_title = i.get["name"][0]["value"]
+            except (KeyError, IndexError):
```
mozaki1, line 89: Are we not eating the exception here? Can we not check first and also this one:  topic_title = i.get("summary", {}).get("title", "") , logic in the exception rather than logging the actual exception ?

ababic: @mozaki1 This is a preferred pattern in Python called ["Ask forgiveness, not permission"](https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain): 

We want this code to fail if there is a `KeyError` (either "name" or "value" are not present) or an `IndexError` ("name" is present, but for some reason, empty).

If either of those things are true, we try the original pattern.

PR #337, opened 2022-05-13
UN 149: Trigger/content warning for specific content

JamesWChan:
Hi @AshTNA,

Could you have a look at this small PR? It just adds the frontend for the content warning component. I've put an `aria-label` on the element to communicate its purpose to screen readers. I haven't added `role="alert"` to this component because going by MDN's documentation:

"Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required."

I don't think this component needs to be highlighted to the user immediately and they will encounter it before they reach the content anyway.


AshTNA (approved):
All good :+1:

PR #338, opened 2022-05-16
DF-210: Make filters list in long filter page searchable

AshTNA:
Hi @JamesWChan 

Would you be able to merge this PR?

It relates to https://national-archives.atlassian.net/browse/DF-210

It adds a search component to the 'long filters' page on Search, which allows the list of filters to be narrowed down via a keyword. It works by:

- Grabbing the NodeList of filter`<li>` elements
- Converting the list into an Array
- Running `filter()` on this array to grab the narrowed down filters
- Rendering these narrowed down filters in the DOM

It will also always show any filters that the user has checked, so that they don't have to delete their keyword and then scroll through all the filters to see which ones are checked.

Accessibility wise, initially we were going to use `aria-live='polite'` to always update users on how many filters were being shown. However it was conflicting with the users' keyword being read out. The screen reader would try to read out how many filters were showing, but then it would cut out and read out the keyword the user had typed instead. Therefore, we are using `ariadescribed-by` on the fieldset to describe how many filters are remaining.

Access the component here: http://localhost:8000/search/catalogue/long-filter-chooser/collection/?q=&per_page=20&sort_order=asc&group=tna

It's been checked on Chrome, Firefox, IE11 and on mobile.

Thanks :+1:

image

### DIFF
--- a/sass/includes/search/_long-filters.scss
+++ b/sass/includes/search/_long-filters.scss
@@ -23,72 +23,75 @@
 
     }
 
-    &__form {
-        @media only screen and (min-width: $screen__md) {
-            display: flex;
+
+
+    &__filters {
+        display: flex;
+
+        @media only screen and (max-width: $screen__md) {
+            margin-top: 2rem;
+            margin-left: 2rem;
         }
 
-        &-filters {
-            @media only screen and (max-width: $screen__md) {
-                margin-top: 2rem;
-                margin-left: 2rem;
+        &-list {
+            padding-left: 0;
+            margin-bottom: 0;
+
+            @media only screen and (min-width: $screen__md) {
+                column-count: 3;
+                column-gap: 1rem;
             }
 
-            &-list {
-                padding-left: 0;
-                margin-bottom: 0;
-
-                @media only screen and (min-width: $screen__md) {
-                    column-count: 3;
-                    column-gap: 2rem;
-                }
-
-                &-item {
-                    list-style: none;
-                    margin-right: 4rem;
-                    width: 100%;
-                    max-width: 20rem;
-                    vertical-align: text-top;
-                }
-            }
-
-            &-checkbox {
-                &:focus {
-                    outline: 5px solid $color__focus-blue-outline-light-bg;
-                }
-                margin-right: 0.2rem;
-            }
-
-            &-label {
-                cursor: pointer;
-                max-width: 85%;
-                vertical-align: top;
-            }
-        }
-
-        &-search {
-            background-color: $color__grey-4;
-            border: 1px solid $color__grey-3;
-            padding: 1rem;
-            max-height: 12rem;
-            &-label {
-                display: block;
-            }
-            &-button {
-                @extend .tna-button;
-                display: block;
-                margin-top: 0.25rem;
-                cursor: pointer;
-            }
-            &-box {
+            &-item {
+                list-style: none;
+                margin-right: 4rem;
                 width: 100%;
-                max-width: 15rem;
-                margin-bottom: 0.5rem;
-                border: 1px solid $color__grey-3;
-                border-radius: 0.5rem;
-                height: 2.7rem;
-                padding: 1rem;
+                max-width: 20rem;
+                vertical-align: text-top;
             }
+        }
+
+        &-checkbox {
+            &:focus {
+                outline: 5px solid $color__focus-blue-outline-light-bg;
+            }
+            margin-right: 0.2rem;
+        }
+
+        &-label {
+            cursor: pointer;
+            max-width: 85%;
+            vertical-align: top;
+        }
+    }
+
+
+
+    &__search {
+        background-color: $color__grey-4;
+        border: 1px solid $color__grey-3;
+        padding: 1rem;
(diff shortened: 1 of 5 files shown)


AshTNA:
I need to get the editorial checked over, so if you approve the changes please don't merge it yet

sass/includes/search/_long-filters.scss
```diff
@@ -11,83 +11,101 @@
...
         }
 
     }
 
-    &__form {
-        @media only screen and (min-width: $screen__md) {
+
+
+    &__filters {
+        @media only screen and (min-width: 60rem) {
```
JamesWChan, line 31: Could this use the `screen__lg` variable? Or does it need to be 60rem (`screen__lg` is 62rem)?

sass/includes/search/_long-filters.scss
```diff
@@ -11,83 +11,101 @@
...
             }
+        }
 
-            &-label {
-                cursor: pointer;
-                max-width: 85%;
-                vertical-align: top;
+        &-checkbox {
+            &:focus {
+                outline: 5px solid $color__focus-blue-outline-light-bg;
```
JamesWChan, line 59: Would we want to change this from `px` to `rem`?

sass/includes/search/_long-filters.scss
```diff
@@ -11,83 +11,101 @@
...
+        &-button {
+            @extend .tna-button--dark;
+            display: block;
+            margin-top: 0.25rem;
+            font-weight: normal;
+            cursor: pointer;
+        }
+        &-box {
+            width: 100%;
+            @media only screen and (min-width: 62rem) {
```
JamesWChan, line 97: Could we use the `screen__lg` variable here?

sass/includes/search/_long-filters.scss
```diff
@@ -11,83 +11,101 @@
...
-                max-width: 15rem;
-                margin-bottom: 0.5rem;
-                border: 1px solid $color__grey-3;
-                border-radius: 0.5rem;
-                height: 2.7rem;
-                padding: 1rem;
+            margin-left: -0.2rem;
+
+            &:focus {
+                outline: 5px solid $color__focus-blue-outline-light-bg;
```
JamesWChan, line 108: Would we want to change this from px to rem?

scripts/src/modules/search/search-long-filters.js
```diff
@@ -0,0 +1,96 @@
...
+        while ($longFiltersList.firstChild) {
+            $longFiltersList.removeChild($longFiltersList.firstChild);
+        }
+
+        for(const filter of narrowedDownFilters) {
+            $longFiltersList.appendChild(filter);
+        }
+
+        let narrowedFiltersLength = narrowedDownFilters.length;
+        let longFiltersCountText = `Showing ${narrowedFiltersLength} out of ${totalFiltersLength} filters`;
```
JamesWChan, line 81: Quick question - you've declared this variable on line 45; does it need redeclaring here or could it just be re-assigned?

scripts/src/modules/search/search-long-filters.js
```diff
@@ -0,0 +1,96 @@
...
+    const totalFiltersLength = longFiltersArray.length;
+    let longFiltersCountText = `Showing ${totalFiltersLength} out of ${totalFiltersLength} filters`;
+    $filterCount.innerText = longFiltersCountText;
+
+    $longFiltersFieldset.insertBefore($filterCount, $longFiltersFieldset.childNodes[2]);
+
+    $longFiltersContainer.insertBefore($searchDiv, $longFiltersContainer.childNodes[0]);
+
+    const handleSearch = function(e) {
+        e.preventDefault();
```
JamesWChan, line 53: Could this be simplified to:

`if(e.key === “Enter”) { 
      e.preventDefault(); 
  }`
  
  And then replace line 94 with something like `$searchBox.addEventListener('keypress', handleSearch);`?
  
Then I think `disableSearchSubmit` wouldn't be needed. Just my opinion though so if it's preferable to keep it as it is then that's cool with me.

JamesWChan:
Hi @AshTNA,

Just left a couple of comments, everything is working well so they're more questions than anything. Also I spotted a small issue on IE11 and Safari with the search box display, screenshot below:

image

scripts/src/modules/search/search-long-filters.js
JamesWChan: Actually, this might not be a good idea. Just realised that this would result in calling the `handleSearch` function twice every time a key is pressed.

AshTNA: Thanks @JamesWChan  - I also think the `preventDefault` has to be on the `$parentForm` element, as with forms it will consider any Enter keypress while inside the form as a submit, so it can't be on the `$searchBox` itself. The `id` checking is so that the submit buttons and other form elements can still trigger a submission by pressing enter.

AshTNA:
Thanks @JamesWChan - everything should be fixed. Feel free to approve then I'll get the editorial checked before merging :+1:

JamesWChan (approved):
All good 👍

PR #341, opened 2022-05-18
Remove heading from promoted list block and remove the sub section block

mozaki1:


etna/insights/tests/test_blocks.py
```diff
@@ -146,21 +100,6 @@ def test_headings_rendered_as_h3(self):
...
         response.render()
         content = response.content.decode()
-        for heading_text in (
-            "This should render as a h3",
-            "Section 1A",
-            "Section 1B",
-        ):
+        for heading_text in ("This should render as a h3",):
             with self.subTest(heading_text):
                 self.assertContainsHeading(content, heading_text, 3)
```
ababic, line 105: Hi @mozaki1 we should remove the `for` and `subTest` lines here now, and just do:
```
self.assertContainsHeading(content, "This should render as a h3", 3) 
```

ababic (changes requested):
Just one tiny thing on the tests, but otherwise looking good! thanks @mozaki1

etna/insights/tests/test_blocks.py
mozaki1: Done

PR #342, opened 2022-05-18
UN 197: Insights page mobile menu scrolling issue

JamesWChan:
Hi @AshTNA,

Would it be possible to have a look at this PR? It implements the bug fix that we discussed earlier. If all's okay, would it be possible to merge (but don't delete the branch)? Thank you 👍

### DIFF
--- a/scripts/src/modules/insights/accordion_functionality/convert_px_property_to_number.js
+++ b/scripts/src/modules/insights/accordion_functionality/convert_px_property_to_number.js
@@ -1,3 +1,3 @@
 export default function convert_px_property_to_number(property) {
-    return parseInt(property.slice(0, 2))
+    return parseInt(property.replace("px", ""));
 }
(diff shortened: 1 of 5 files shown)


scripts/src/modules/insights/accordion_functionality/convert_px_property_to_number.js
```diff
@@ -0,0 +1,3 @@
+export default function convert_px_property_to_number(property) {
+    return parseInt(property.slice(0, 2))
```
AshTNA, line 2: Should this instead do a string replace of `px`  with an empty string, and then parse that as an int? This would future proof it as you aren't reliant on the number always having the same amount of digits. 

Would moving everything to REM also affect this function? With string replace you could replace both `px` or `rem` with an empty string. I'm guessing JavaScript scroll always works in pixels though so accounting for REM might not be necessary.

AshTNA (commented):
Looks good, though could you respond to the comment I've added. Thanks

scripts/src/modules/insights/accordion_functionality/convert_px_property_to_number.js
JamesWChan: Thanks @AshTNA, I've changed that function to use `replace`. I don't think we need to worry about `rem` so I believe it's fine as it is but it can be easily fixed if it turns out to be an issue later on.

AshTNA (approved):
All good :+1:

PR #344, opened 2022-05-22
Create about app

mozaki1:


etna/generic_pages/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class GenericPagesConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "etna.generic_pages"
+    verbose_name = "Generic_Pages"
```
ababic, line 7:
```suggestion
    verbose_name = "Generic pages"
```

etna/generic_pages/models.py
```diff
@@ -3,13 +3,13 @@
...
 from etna.core.models import BasePage
 
-from .blocks import AboutPageStreamBlock
+from .blocks import GeneralPageStreamBlock
 
 
-class AboutPage(BasePage):
-    """About Page"""
+class GeneralPage(BasePage):
+    """GeneralPage"""
```
ababic, line 10: This docstring is not useful in it's current state, so it'd probably be better to remove it

etna/generic_pages/tests/test_blocks.py
```diff
@@ -5,22 +5,22 @@
```
ababic, line 5: Maybe `test_models.py` would be a better name for this module?

ababic (changes requested):

etna/generic_pages/tests/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertTemplateUsed(response, "generic_pages/general_page.html")
+
+    def test_paragraph_rendered_as_div(self):
+        url = self.general_page.get_url()
+        response = self.client.get(url)
+        response.render()
+        content = response.content.decode()
+        for paragraph_text in ("This is a paragraph",):
+            with self.subTest(paragraph_text):
+                self.assertContainsParagraph(content, paragraph_text)
```
ababic, line 54: If the paragraph block is just a regular `StructBlock`, what value are we getting out of this test?

etna/generic_pages/tests/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+        paragraph_texts = []
+        for match in regex.finditer(subject):
+            paragraph_texts.append(match.groups()[0].strip())
+
+        self.assertIn(paragraph_text, paragraph_texts)
+
+    def test_view_url_accessible_by_name(self):
+        url = "/general-page/"
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
```
ababic, line 39: What is it that we're testing here? It seems a little unclear.

etna/generic_pages/tests/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_view_uses_correct_template(self):
+        url = self.general_page.get_url()
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "generic_pages/general_page.html")
```
ababic, line 45: This test is far better, as it's much more obviously related to the test page :)

templates/generic_pages/general_page.html
```diff
@@ -0,0 +1,20 @@
...
+
+
+
+{% block content %}
+
+    {% include 'includes/generic-intro.html' %}
+
+    <div class="container pt-4">
+        {% for block in page.body %}
+            <div class="row">
```
ababic, line 14: Should this `<div>` be here? Maybe it should be on the outside of the `{% for %}` loop?

etna/generic_pages/tests/test_blocks.py
mozaki1: Hi Andy, so basically we can use the page name or the model name to do something like http://localhost:8000/general-page/

etna/generic_pages/tests/test_blocks.py
mozaki1: Hi Andy , I have commented on the above test :)

etna/generic_pages/tests/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+    def assertContainsParagraph(self, subject: str, paragraph_text: str):
+        regex = re.compile(
+            "<div" + r"[\s\w'\=\-\"\.]*>\s*(.*)\s*<\/div" + ">",
+            re.MULTILINE,
+        )
+        paragraph_texts = []
+        for match in regex.finditer(subject):
+            paragraph_texts.append(match.groups()[0].strip())
+
+        self.assertIn(paragraph_text, paragraph_texts)
```
ababic, line 34: As noted already: The paragraph block is a basic `StructBlock`, so there's no reason it wouldn't render as expected. The tests in `etna/insights/tests/test_blocks.py` are only there to test the behaviour of `SectionDepthAwareStructBlock`

etna/generic_pages/tests/test_blocks.py
ababic: But all this test is really doing is calling `self.client.get(url)` with a hardcoded value. The test after this one is doing the  exact same thing: calling `self.client.get(url)`... the only difference is that it gets the `"/general-page/"` string from the `get_url()` method beforehand.

mozaki1: This is correct so maybe on the test_view_uses_correct_template , I can remove the   self.assertEqual(response.status_code, 200) , however I believe it is a good test for self.general_page.get_url()

etna/generic_pages/tests/test_blocks.py
mozaki1: So the test is "test_paragraph_rendered_as_div" the test is testing the paragraph render in a div tag,  unless we know for sure this is the case always then maybe not needed.

etna/generic_pages/tests/test_blocks.py
ababic: I would suggest removing this test completely, and just leaving the second one

etna/generic_pages/tests/test_blocks.py
ababic: It makes sense for the tests in insights because we need to confirm that the correct heading tag is used. Whether or not the paragraph appears in a div is entirely dependent on what we put in the template, so it isn't really offering any value.

templates/generic_pages/general_page.html
mozaki1: So every Paragraph block has its own div tag.

etna/generic_pages/tests/test_blocks.py
mozaki1: This test is now removed

etna/generic_pages/tests/test_blocks.py
mozaki1: This test is now removed

etna/generic_pages/tests/test_blocks.py
mozaki1: This test is now removed

etna/generic_pages/tests/test_blocks.py
mozaki1: Done

etna/generic_pages/models.py
mozaki1: Done

etna/generic_pages/apps.py
mozaki1: Done

templates/generic_pages/general_page.html
ababic: Okay, but was this a requirement? It's a little unusual to wrap individual blocks in extra HTML, as each block type has its own template to control the output.

mozaki1: Yes I can see this :                                                                                                                                                           <div class="row">               
<div class="content-block">
    <div class="container">
        <p data-block-key="38tqm">General here</p>
    </div>
  </div>
</div>

mozaki1: I will remove it , just to mention we may have the same issue on the home page

ababic (approved):
Excellent work! Thanks @mozaki1!

PR #345, opened 2022-05-23
CHORE: Replace transformation function with lazily-evaluated properties

ababic:
Removes the long/fragile transformation function in favour of enhancing the `Record` class.

These changes are helpful because they:

1. Remove the most fragile of the steps from our over-complicated serialisation process
2. Break a long/complicated piece of code into smaller, more-testable chunks
3. They make the `Record` class the one single place for managing how data is converted to more useful structures
4. They bring us a step closer to using `Record` instances everywhere to represent records (e.g. even search results), which will allow us to remove all kinds of crazy special-case code for dealing with raw dicts


etna/records/models.py
```diff
@@ -182,12 +182,20 @@ def hierarchy_level3_summary_title(self) -> str:
     @cached_property
     def next_record(self) -> Union[Dict[str, str], None]:
         if record := self.get("@next"):
-            return {"iaid": extract(record, "@admin.id")}
+            ref_num = self.get("identifier.0.reference_number")
```
ababic, line 185: @jamesbiggs1 I don't think `self.get()` is the right thing to use here... that will give you the current record's reference number, when what you want is the one for the next record.

etna/records/models.py
```diff
@@ -182,12 +182,20 @@ def hierarchy_level3_summary_title(self) -> str:
...
+            return dict(
+                iaid=extract(record, "@admin.id"),
+                reference_number=ref_num,
+            )
 
     @cached_property
     def previous_record(self) -> Union[Dict[str, str], None]:
         if record := self.get("@previous"):
-            return {"iaid": extract(record, "@admin.id")}
+            ref_num = self.get("identifier.0.reference_number")
```
ababic, line 194: @jamesbiggs1 I don't think `self.get()` is the right thing to use here... that will give you the current record's reference number, when what you want is the one for the previous record.

jamesbiggs: Hi Andy, cheers for pointing this out - I didn't even realise... I was having an issue referencing the @next.identifier, so I thought I did it right when it worked! I've fixed this now, cheers :)

etna/ciim/utils.py
```diff
@@ -1,7 +1,8 @@
 import re
 
-from typing import Dict
+from typing import Any, Dict
 
+from django.template import TemplateSyntaxError, Variable, VariableDoesNotExist
```
TNA-Allan, line 5: 1. This is not standard in Django tough, is this taken from here https://github.com/django/django/blob/main/django/template/base.py

ababic: I'm not sure what you mean @TNA-Allan? I'm importing from:
https://github.com/django/django/blob/main/django/template/__init__.py

etna/records/models.py
```diff
@@ -1,78 +1,261 @@
...
+            if fail_loudly:
+                raise VariableDoesNotExist(str(e))
+            return default
 
-    @classmethod
-    def from_api_response(cls, response: dict) -> Record:
-        return cls(**transform_record_result(response))
+    @cached_property
+    def template(self) -> Dict[str, Any]:
+        return self.get("@template.details", self.get("@template.results", {}))
```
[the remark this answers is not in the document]
ababic: 1. Because, as you know, there are [two possible templates ](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/ciim/client.py#L51-L58) we can request. The `Record` class prefers the `details` template, because it's more complete. But, if that isn't available, the `results` template still usually contains the important stuff, hence why we fall back to that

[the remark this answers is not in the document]
ababic: 2. I'm not sure what we gain from knowing about the `template` param that was used. Either the template data is present or it isn't. Because this is a cached property, the work to figure out the return value is only done once... the first time it is requested. On every subsequent request to `self.template`, the cached return value is used.

etna/ciim/utils.py
TNA-Allan: What I mean by its not standard that its not documented in Django Documentation to use `TemplateSyntaxError, Variable, VariableDoesNotExist`. I think it usage is on special needs only to extend a functionality. 

I notice a difference in accessing the api response path as a result of using `django.template`.

1. Is there anything which did not function as required in order to use this functionality of django.template?
2. Can this bit be substituted with anything else,  so that we not want any impact on our code if ever there are changes to the internal working of the template?

etna/ciim/utils.py
```diff
@@ -14,6 +15,31 @@ def underscore_to_camelcase(word, lower_first=True):
...
+        # that is not an issue here, so let's bypass that validation
+        variable_node = Variable("source")
+        variable_node.var = variable_str
+        variable_node.lookups = tuple(variable_str.split("."))
+    try:
+        return variable_node.resolve({"source": source})
+    except VariableDoesNotExist:
+        if fail_loudly:
+            raise
+        return default
```
TNA-Allan, line 40: Can we have tests around VariableDoesNotExist:=, fail_loudly=True and default?

etna/ciim/utils.py
```diff
@@ -14,6 +15,31 @@ def underscore_to_camelcase(word, lower_first=True):
...
+    Utilises Django template engine's variable resolution logic to extract
+    values from unpredictable nested data structures using a simple '.'
+    separated string to indicate keys, attributes, methods and indexes to be
+    traversed or called.
+    """
+    variable_str = f"source.{key}"
+    try:
+        # attempt regular initialization to trigger validation
+        variable_node = Variable(variable_str)
+    except TemplateSyntaxError:
```
TNA-Allan, line 29: Can we have tests around `TemplateSyntaxError`?

etna/ciim/utils.py
```diff
@@ -14,6 +15,31 @@ def underscore_to_camelcase(word, lower_first=True):
...
+        # the template engine doesn't like segments that start with '_', but
+        # that is not an issue here, so let's bypass that validation
+        variable_node = Variable("source")
+        variable_node.var = variable_str
+        variable_node.lookups = tuple(variable_str.split("."))
+    try:
+        return variable_node.resolve({"source": source})
```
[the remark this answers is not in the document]
ababic: Isn't this self explanatory? We're creating an object called `variable_node`, then updating a couple of its attribute values directly to bypass the validation in `Variable.__init__()`

etna/ciim/utils.py
ababic: Yes

etna/ciim/utils.py
ababic: Yes

etna/records/models.py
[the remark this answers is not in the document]
TNA-Allan: Its pointless keeping something like `@template.results` when 
1. Not sure what its used for,
2. `@template.results` is  also available in the response as the input param `template=results` is not used in the code.
3. Gives wrong impression that the results response is available.

etna/records/models.py
```diff
@@ -1,78 +1,261 @@
...
+    @cached_property
+    def iaid(self) -> str:
+        try:
+            return self.template["iaid"]
+        except KeyError:
+            pass
+        admin_id = self.get("@admin.id", fail_loudly=True)
+        if re.match(IAIDConverter.regex, admin_id):
+            return admin_id
+        raise VariableDoesNotExist(f"@admin.id '{admin_id}' is not a valid iaid.")
```
TNA-Allan, line 64: 1. `VariableDoesNotExist(f"@admin.id '{admin_id}' is not a valid iaid.")` The Attribute and message talks about different things - a Variable that does not exist and Invalid Data. I think Invalid Data must be handled separately.
2. In order to see `VariableDoesNotExist` on browser, `/catalogue/id/C1286726/` I tried commenting out `#if re.match(IAIDConverter.regex, admin_id):  #return admin_id`. But it gets swallowed by the `utils.extract` and does not display.  Alternatively I tried giving a dummy iaid in  `Record.__init__ method` (`["@admin"]["id"]="C12345"`) but `VariableDoesNotExist` does not show in the Browser

etna/records/models.py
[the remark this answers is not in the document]
ababic: @TNA-Allan But it's not pointless... we know that it can be used and can be present in the representations. The ultimate fallback is `{}` (a blank dict), so it's okay if it isn't there.... but we can certainly use it if it is.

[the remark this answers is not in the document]
TNA-Allan: > The ultimate fallback is {} (a blank dict)
The pointlessness is not about blank dict, also it was not in question.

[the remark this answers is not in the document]
TNA-Allan: > but we can certainly use it if it is.
Surely, as long we use it when its needed, not otherwise.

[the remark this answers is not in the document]
ababic: I'm honestly lost here @TNA-Allan. Can you articulate what you want to change and why?

etna/ciim/utils.py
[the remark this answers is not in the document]
ababic: Have re-written this now, so you don't have to understand any of the stuff from the `django.template` module

etna/ciim/utils.py
ababic: This is irrelevant now, as `VariableDoesNotExist` is not used

etna/ciim/utils.py
ababic: This is irrelevant now, as `VariableDoesNotExist` is not used

etna/records/models.py
ababic: 1. The method now raises `ValueError` when the value is not a valid `iaid`
2. We're now raising an `ValueExtractionError` instead of `VariableDoesNotExist`. Also, I have explained in person that this is necessary to avoid using the value as an `iaid` when it is not suitable.

etna/records/models.py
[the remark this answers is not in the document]
TNA-Allan: > I'm honestly lost here @TNA-Allan. Can you articulate what you want to change and why?
@ababic  The change is to remove @template.results and several reasons are mentioned in the my point above. Happy do discuss further if you're still lost.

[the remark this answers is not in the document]
ababic: @TNA-Allan But I explained in my first reply about this:
> There [two possible templates](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/ciim/client.py#L51-L58) we can request from Kong

While this is true of the API, wouldn't you find it odd if you requested the `results` template, but the Record instance couldn't access that data, even if the API provided it?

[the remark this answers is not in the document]
TNA-Allan: > wouldn't you find it odd if you requested the results template

See below if this helps why we do not want it-

> Not sure what its used for,
@template.results is available in the response as the input param template=results but this param is not used in the code.
Gives wrong impression that the results response is available.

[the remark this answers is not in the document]
TNA-Allan: We could definitely use it whenever it's param is used in the code.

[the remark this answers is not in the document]
ababic: Try to detach this code in your mind from how it is currently being used currently. You should see this as a generic interface for interacting with record data. While it's true that it makes the values easier to access in views, that is not the limit of it's usefulness... maybe you'll need to write a management command or write a new view soon that does use `template=results` - You shouldn't have to worry about whether that data will be respected or not at the time... you would just expect it to work.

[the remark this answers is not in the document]
ababic: With respect, only catering for 'right now' is the very reason changes like this are needed. Sometimes you've got to pull your mind away from the immediate requirements and develop interfaces that make sense in their own right. When you allow two interfaces to bend too much to the whims of the other, you quickly end up in a mess.

etna/records/models.py
```diff
@@ -1,78 +1,314 @@
...
+    format_description_markup,
+    format_link,
+)
+from .converters import IAIDConverter
+
+logger = logging.getLogger(__name__)
 
 
-@dataclass
 class Record(DataLayerMixin, APIModel):
```
TNA-Allan, line 27: @ababic Isn't it Best Practice to calculate the data and then pass them to the view as in the `dataclass` implementation of Record. With the current implementation using property methods each template tag has to resolve to the requested value by extracting using _Ask for Forgiveness_ implementations. Although its done in memory and not querying the database is the performance benefit of this implementation is an improvement over `dataclass` when accessing data through template tags?

etna/ciim/utils.py
```diff
@@ -14,6 +14,58 @@ def underscore_to_camelcase(word, lower_first=True):
...
+
+    try:
+        for bit in lookups:
+            try:  # dictionary lookup
+                current = current[bit]
+                # ValueError/IndexError are for numpy.array lookup on
+                # numpy < 1.9 and 1.9+ respectively
+            except (TypeError, AttributeError, KeyError, ValueError, IndexError):
+                try:  # attribute lookup
+                    current = getattr(current, bit)
```
TNA-Allan, line 45: `current = getattr(current, bit)` Could not find a scenario satisfying this assignment, Can we a test case for it?

etna/ciim/utils.py
```diff
@@ -14,6 +14,58 @@ def underscore_to_camelcase(word, lower_first=True):
...
+                except (TypeError, AttributeError) as e:
+                    # Reraise if the exception was raised by a @property
+                    if bit in dir(current):
+                        raise
```
TNA-Allan, line 49: `bit in dir(current)` Could not find a scenario satisfying this assignment, Can we a test case for it?

etna/records/models.py
ababic: And we're using _Ask for Forgiveness_ implementations here too... nothing has changed there - it's just that the work is split into separate units, so a failure in one unit doesn't kill everything else.

etna/ciim/utils.py
ababic: @TNA-Allan This happens whenever the initial dict lookup fails. In the tests, you'll see that we're looking up attributes from date objects in lots of places... that is where this code will be working

etna/ciim/utils.py
ababic: Yes we can

etna/records/models.py
TNA-Allan: The number of fields to access are same for both implementations. I thought it would be an unnecessary overhead of evaluating all the template tags to get the value, than pre-populating them in a `dataclass` without overhead to get the value. The new instance initialise for every attribute tag accessed, so I am not sure if that is cheaper than pre-populating those tag attributes only once. But as you are saying so, it must be the case.

ababic: I would encourage you to measure the rendering speed of the detail pages before and after these changes for a comparisson

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+    @cached_property
+    def title(self) -> str:
+        try:
+            return self.template["summaryTitle"]
+        except KeyError:
+            return self.get("summary.title", default="")
+
+    @cached_property
+    def closure_status(self) -> str:
+        return self.get("access.conditions", default="")
```
TNA-Allan, line 178: The path is `availability.access.condition.value` (existing code is incorrect) and template path is `@template.details.accessCondition` see `iaid=C7860693`
Perhaps using `return self.get("availability.access.condition.value", self.template.get("accessCondition", ""))`

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+
+    @cached_property
+    def arrangement(self) -> str:
+        return format_description_markup(self.get("arrangement.value", default=""))
```
TNA-Allan, line 182: `arrangement` also has a template path `@template.details.arrangement`

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+
+    @cached_property
+    def legal_status(self) -> str:
+        return self.get("legal.status", default="")
```
TNA-Allan, line 186: `legal_status` also has template path '@template.details.legalStatus`

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+    def media_reference_id(self) -> str:
+        return self.get("multimedia.@admin.id", default="")
+
+    @cached_property
+    def catalogue_source(self) -> str:
+        return self.get("source.value", default="")
+
+    @cached_property
+    def description(self) -> str:
+        descriptions = self.get("description", default=())
```
TNA-Allan, line 206: Could we use `default=[]` since in the response it is a list.

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+            return self.template["sourceUrl"]
+        except KeyError:
+            pass
+
+        locations = self.get("source.@location", ())
+        if len(locations) == 1:
+            return locations[0]["value"]
+        for item in locations:
+            if item.get("public", False):
+                return item["value"]
```
TNA-Allan, line 152: Could not find this path `source.@location` in api response or spec using stream=interpretive. 
Can we have a test case for this.

etna/records/models.py
```diff
@@ -1,78 +1,316 @@
...
+    def origination_date(self) -> str:
+        return self.template.get("dateCreated", "")
+
+    @cached_property
+    def level(self) -> str:
+        return self.get("level.value", self.template.get("level", ""))
+
+    @cached_property
+    def level_code(self) -> int:
+        return self.get("level.code", None)
```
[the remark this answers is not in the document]
ababic: I have changed this to an int, as I can't see a valid reason for converting it to a string

etna/records/models.py
ababic: Hi @TNA-Allan. These are present for insights results:
{{BASE_URL}}/api/v1/data/searchAll?filterAggregations=group:insight&size=10

ababic: Have removed in favour of just looking for `self.template["sourceUrl"]`

etna/records/models.py
```diff
@@ -1,78 +1,319 @@
...
+        return self.get("source.value", default="")
+
+    @cached_property
+    def description(self) -> str:
+        descriptions = self.get("description", default=())
+        if len(descriptions) == 1:
+            return format_description_markup(descriptions[0]["value"])
+        for item in descriptions:
+            if item.get("type", "") == "description":
+                return format_description_markup(item["value"])
```
TNA-Allan, line 214: We have `@template.details.description` available to use, perhaps as default, but I'm not sure if it picks up the correct description when there is more than one type="description". With len=1, the type = "scope and content" so far as I seen.

templates/includes/records/hierarchy-local.html
```diff
@@ -4,32 +4,33 @@
     <div class="container">
         <div class="row">
             <div class="col-4 col-md-2">
-                {% if page.previous_record %}
+                {% if page.previous_record.has_reference_number %}
```
TNA-Allan, line 7: I don't think this is right. The Previous and next records should show irrespective of having reference numbers or not.

ababic: @TNA-Allan I have been told otherwise... we only want to show the links when there is a reference number, as we want what is displayed to be consistent.

templates/includes/records/hierarchy-local.html
```diff
@@ -4,32 +4,33 @@
...
                     <p class="hierarchy-local__nav-item">
                         <span class="hierarchy-local__nav-indicator hierarchy-local__nav-indicator--left"></span>
                         <span class="hierarchy-local__nav-label">{{ 'previous_record'|as_label }}</span>
-                        <a href="{% record_url page.previous_record %}" data-link="Previous" class="link--bold"><span class="sr-only">{{ 'previous_record'|as_label }}</span> {{ page.previous_record.iaid }}</a>
+                        <a href="{% record_url page.previous_record %}" data-link="Previous" class="link--bold"><span class="sr-only">{{ 'previous_record'|as_label }}</span> {{ page.previous_record.reference_number }}</a>
```
TNA-Allan, line 11: > @TNA-Allan I have been told otherwise... we only want to show the links when there is a reference number, as we want what is displayed to be consistent.

The underlying `href url` shows `iaid` from `record_url`, but in [previous version before using record_url](https://github.com/nationalarchives/ds-wagtail/blob/566d39a2ec8bd3543eda73a7f36c117dd8649a78/templates/includes/records/hierarchy-local.html) of this template it shows underlying `href url` as `reference number` when it has a `reference number`.

etna/records/models.py
ababic: Have decided to ditch that code and just use `highlight.@template.details.description` or `@template.details.description`

templates/includes/records/hierarchy-local.html
ababic: The link may work regardless, but we want the link text to always be a reference number... so that users always know what is being shown.

templates/includes/records/hierarchy-local.html
TNA-Allan: >@TNA-Allan I have been told otherwise... we only want to show the links when there is a reference number, as we want what is displayed to be consistent.

My concern is about underlying URL and not what is displayed in this context. This condition does not allow to show 'iaid' if `reference number` is not available. Was this thought about?

etna/records/models.py
```diff
@@ -97,30 +352,23 @@ def custom_dimension_11(self) -> str:
...
-                + " - "
-                + self.hierarchy_level3_summary_title
-            )
-        else:
-            return ""
+        for ancestor in self.hierarchy:
+            if ancestor.level_code == 3:
+                if ancestor.has_reference_number():
+                    return f"{ancestor.reference_number} - {ancestor.title}"
+                return ancestor.title
```
TNA-Allan, line 364: As in the previous version of transforms we don't want to return a value if only one of them is present. Its both or blank.

etna/records/models.py
```diff
@@ -97,30 +352,23 @@ def custom_dimension_11(self) -> str:
...
 
     @property
     def custom_dimension_14(self) -> str:
-        if self.template_reference_number and self.template_summary_title:
-            return self.template_reference_number + " - " + self.template_summary_title
-        else:
-            return ""
+        if self.has_reference_number():
+            return f"{self.reference_number} - {self.title}"
+        return self.title
```
TNA-Allan, line 371: As in the previous version of transforms we don't want to return a value if only one of them is present. Its both or blank.

etna/records/models.py
```diff
@@ -1,78 +1,333 @@
...
+                raise ValueExtractionError(str(e))
+            return default
+
+    @cached_property
+    def template(self) -> Dict[str, Any]:
+        return self.get("@template.details", default=self.get("@template.results", {}))
+
+    @cached_property
+    def highlights(self) -> Dict[str, Any]:
+        return self.get("highlight", default={})
```
TNA-Allan, line 67: Where is this property used? What API query will return data for highlight?

etna/records/models.py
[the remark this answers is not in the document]
TNA-Allan: I differ on this at this time, but going with your reasoning.

TNA-Allan:
> @TNA-Allan I have rewritten `extract()` again to make things easier to follow. Also, because we're not attempting to access attributes until the end, we can just raise any errors at that point (so we don't need as much 'special-case' handling).

Thank you :)

etna/records/templatetags/records_tags.py
```diff
@@ -21,33 +23,60 @@ def record_url(
...
+        ref = None
+
+    try:
+        # Use Record property if available
+        url = record.url
+    except AttributeError:
+        # 'record' is likely just a dict
+        url = record.get("sourceUrl")
+    except ValueExtractionError:
+        url = None
```
TNA-Allan, line 51: Can we have test cases for the `record_url` template tags?

etna/records/models.py
ababic: This is supplied when you do a text search... it brings back values with `<mark></mark>` applied to indicate where the terms matched for that result

etna/records/templatetags/records_tags.py
ababic: We can, but that's got to be out of scope for this PR, surely?

TNA-Allan: I was looking at test cases within the scope - the new elements introduced when compared with the previous version 
iaid = record.iaid, ref = record.reference_number, url = record.url

templates/includes/records/hierarchy-local.html
TNA-Allan: Hi @ababic Example `iaid=C34131` does not have _reference numbers_ in _@next/@previous_. In this case we need the `iaid`.

etna/records/templatetags/records_tags.py
ababic: I haven't really got time to write tests for record_url if they don't already exist. Happy for there to be a ticket created for that though.

templates/includes/records/hierarchy-local.html
ababic: @TNA-Allan I can't find the ticket in Jira any longer, but it's one that was assigned to @jamesbiggs. If I remember rightly, detail on the ticket was pretty light anyway. Not showing the link unless there is a reference number is a detail that came out of a conversation he had with @sdwilkes. You'll just have to trust me on this one.

etna/records/models.py
ababic: Resolved in 77f2f69

templates/includes/records/hierarchy-local.html
ababic: Have reverted the changes to this template... Will submit those changes as a separate PR

etna/records/templatetags/records_tags.py
TNA-Allan: >if they don't already exist. 

May be misleading as -

>I was looking at test cases within the scope - the new elements introduced when compared with the previous version
iaid = record.iaid, ref = record.reference_number, url = record.url

It would be better in this ticket as its only related to the changes you have made. If you want to take it on a separate ticket,  then please go ahead please address it there cc @sdwilkes  @ACTNA 

This is an important part of this development, if its a separate ticket it would helpfull to see other areas of `record_url` addressed which was previously not.

ababic: I'll write some tests, but this PR is not the place for them. The new PR should be merged first, then this one rebased on top, so you can see that the changes haven't caused the tests to fail.

etna/records/models.py
```diff
@@ -1,78 +1,340 @@
...
+        if parent_data := self.get("parent.0", default=None):
+            return Record(parent_data)
+
+    @cached_property
+    def hierarchy(self) -> Tuple["Record"]:
+        return [
+            Record(item)
+            for item in self.get("@hierarchy.0", default=())
+            if item.get("identifier")
+        ]
```
TNA-Allan, line 286: Return `list []` can be updated to match type hint `Tuple`

etna/records/models.py
```diff
@@ -1,78 +1,340 @@
...
+        if prev := self.get("@previous", default=None):
+            return Record(prev)
+
+    @cached_property
+    def topics(self) -> Tuple[Dict[str, str]]:
+        return_value = []
+        for item in self.get("topic", default=()):
+            topic_title = ""
+            try:
+                topic_title = item.get["name"][0]["value"]
```
TNA-Allan, line 304: Was not able to find the API response for topic, which API returns this path?

etna/records/templatetags/records_tags.py
```diff
@@ -21,33 +23,60 @@ def record_url(
...
     returning more than one record
     """
-    if iaid := record.get("iaid"):
+    try:
+        # Use Record property if available
+        iaid = record.iaid
+    except AttributeError:
+        # 'record' is likely just a dict
+        iaid = record.get("iaid")
+    except (ValueExtractionError, ValueError):
```
TNA-Allan, line 32: `ValueError` is a particular case for exception handling. 
`ValueError` cases where `iaid` are not yet brought into Etna i.e. does not match catalogue `IAIDConverter` regex - example iaid for related article for blog post.
It would be nice to have some comment on it.

etna/records/models.py
ababic: The aim was to replicate what the transform function is doing. See original code here: https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/records/transforms.py#L85-L94

etna/records/templatetags/records_tags.py
ababic: @TNA-Allan see here for changes needed to get the existing record_url tests to pass (just initialising the `Record` objects differently to match the updated behaviour)... https://github.com/nationalarchives/ds-wagtail/pull/345/commits/753c2de6eff35131774e7ecc097192e405e59717

etna/records/models.py
TNA-Allan: @ababic Thanks Andy for coming along with me to show this test.

TNA-Allan (approved):
Thanks Andy, it looks good. Great piece of work. :)

PR #346, opened 2022-05-24
CHORE: Test configuration tweaks

ababic:
Test code is very often difficult enough to understand, so the more we can do to centralise test configuration and simplify the test module code, the better.

Essentially, we don't want our unit tests to ever be making live requests to Kong - so, it makes sense that this configuration is always set when testing.


ababic:
@TNA-Allan / @mozaki1. This has been rebased to address merge conflicts, and is now ready for review

ababic:
@mozaki1 Re-requesting a review on this one.

mozaki1 (approved):

PR #347, opened 2022-05-24
DF-319: Add settings to support private/public toggling of specific views

ababic:
Relates to: https://national-archives.atlassian.net/browse/DF-319 

As we move toward a partial launch (where we're still a little unclear on what will / not be made public), I was asked to look at how we might control this sort of thing more easily. These changes allow us to make certain areas public or private on demand, by simply tweaking the environment configuration.

### DIFF
--- a/etna/records/tests/test_views.py
+++ b/etna/records/tests/test_views.py
@@ -19,16 +19,6 @@ User = get_user_model()
     KONG_CLIENT_BASE_URL="https://kong.test",
 )
 class TestRecordDisambiguationView(TestCase):
-    def setUp(self):
-        private_beta_user = User(
-            username="private-beta@email.com", email="private-beta@email.com"
-        )
-        private_beta_user.set_password("password")
-        private_beta_user.save()
-        private_beta_user.groups.add(Group.objects.get(name="Beta Testers"))
-
-        self.client.login(email="private-beta@email.com", password="password")
-
     @responses.activate
     def test_no_matches_respond_with_404(self):
         responses.add(
@@ -99,16 +89,6 @@ class TestRecordDisambiguationView(TestCase):
     KONG_CLIENT_BASE_URL="https://kong.test",
 )
 class TestRecordView(TestCase):
-    def setUp(self):
-        private_beta_user = User(
-            username="private-beta@email.com", email="private-beta@email.com"
-        )
-        private_beta_user.set_password("password")
-        private_beta_user.save()
-        private_beta_user.groups.add(Group.objects.get(name="Beta Testers"))
-
-        self.client.login(email="private-beta@email.com", password="password")
-
     @responses.activate
     def test_no_matches_respond_with_404(self):
         responses.add(
@@ -217,13 +197,6 @@ class TestRecordView(TestCase):
     KONG_CLIENT_BASE_URL="https://kong.test",
 )
 class TestDataLayerRecordDetail(WagtailTestUtils, TestCase):
-    def setUp(self):
-        """These tests are for testing the view logic (not the auth protection logic).
-        Therefore keeping more generic."""
-        # To bypass view-level protection
-        super().setUp()
-        self.login()
-
     @responses.activate
     def test_datalayer_level1(self):
         import json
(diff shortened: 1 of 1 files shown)


etna/core/tests/test_decorators.py
```diff
@@ -0,0 +1,71 @@
...
+
+import responses
+
+from ...ciim.tests.factories import create_response
+
+CONDITIONALLY_PROTECTED_URLS = (
+    reverse_lazy("search-catalogue"),
+    reverse_lazy("image-browse", kwargs={"iaid": "C140"}),
+    reverse_lazy("details-page-machine-readable", kwargs={"iaid": "C140"}),
+)
```
TNA-Allan, line 14: Generic - Is there any time its too much to include in the test. Is it fine to have some than all - for example urls in this list? Of course, I imagine its really helpful in getting some testcases here than it not being there at all. Thanks for that.

etna/core/decorators.py
```diff
@@ -0,0 +1,24 @@
...
+    redirect_field_name=REDIRECT_FIELD_NAME,
+    login_url=None,
+):
+    """
+    Decorator for views that conditionally require login based on the value
+    of a Django project setting (usually controlled via env vars),
+    redirecting to the log-in page if necessary.
+    """
+    actual_decorator = user_passes_test(
+        lambda u: not getattr(settings, setting_name) or u.is_authenticated,
```
TNA-Allan, line 18: Does the callable 'user' instead of 'u' makes it easier to understand.

TNA-Allan (changes requested):

etna/core/context_processors.py
```diff
@@ -0,0 +1,9 @@
+from django.conf import settings
+
+
+def feature_flags(request):
+    return {
+        "IMAGE_VIEWER_REQUIRE_LOGIN": settings.IMAGE_VIEWER_REQUIRE_LOGIN,
+        "RECORD_DETAIL_REQUIRE_LOGIN": settings.RECORD_DETAIL_REQUIRE_LOGIN,
+        "SEARCH_VIEWS_REQUIRE_LOGIN": settings.SEARCH_VIEWS_REQUIRE_LOGIN,
+    }
```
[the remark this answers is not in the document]
ababic: Not currently, but it's quite possible that you might want to do something based on the user's current authentication state and whether a login is required. When taking into account that this is literally just a dictionary of setting values that are already in memory, the overhead is tiny. It's consistency that I feel is more important here than anything - we make these settings available everywhere as a pattern, and stick to it.

etna/core/decorators.py
```diff
@@ -0,0 +1,24 @@
+from django.conf import settings
+from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.contrib.auth.decorators import user_passes_test
+
+
+def setting_controlled_login_required(
+    function=None,
+    setting_name=None,
+    redirect_field_name=REDIRECT_FIELD_NAME,
```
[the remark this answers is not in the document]
ababic: This is basically copying the Django version as closely as possible: https://github.com/django/django/blob/main/django/contrib/auth/decorators.py#L43-L57

etna/core/tests/test_decorators.py
ababic: I don't have any solid rules for where to put URLs, only that using `reverse_lazy` is generally more flexible than `reverse`, because it doesn't require all of the URLs to have been registered yet. These could just as well live on the test case class, or elsewhere. In this case, I felt like having them as a global better highlighted their importance to the rest of the module.

etna/core/context_processors.py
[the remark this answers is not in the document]
ababic: See https://github.com/nationalarchives/ds-wagtail/pull/351 for a better example

etna/core/decorators.py
ababic: I agree :)

etna/core/tests/test_decorators.py
TNA-Allan: Going with 'not having any solid rules'.

config/settings/base.py
```diff
@@ -344,3 +343,13 @@
 CACHE_CONTROL_STALE_WHILE_REVALIDATE = int(
     os.getenv("CACHE_CONTROL_STALE_WHILE_REVALIDATE", 30)
 )
+
+# Feature flags: simple boolean settings that allow us to toggle features on/off in
```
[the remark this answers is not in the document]
ababic: @TNA-Allan Have decided to leave the 'feature flags' bit out of this PR and just use standard settings/env vars

TNA-Allan (approved):
Looks good. Thank you.

PR #340, opened 2022-05-18
DF-253: Website results -  add correct URLS to links

AshTNA:
Hi @ababic 

This completes https://national-archives.atlassian.net/browse/DF-253

Would you be able to review this implementation of Website result URLs? I'm guessing you might have a better way of implementing this.

Thanks :+1:

PR #348, opened 2022-05-24
DF-253: Use {% record_url %} everywhere for record links

ababic:
This is an extension of the work started by @AshTNA in #340, and a pre-requisite for one of the things we're looking to try out ahead of Insights launch.


AshTNA (approved):
all good :+1:

PR #349, opened 2022-05-25
UN 197: Insights page mobile menu scrolling issue

JamesWChan:
Hi @AshTNA,

Would it be possible to have a look at this PR? It addresses the scrolling issue on mobile that we've been discussing. I'm still using `px` here but I will refactor this to use `rem` in another PR. Thank you 👍

### DIFF
--- a/scripts/src/insights.js
+++ b/scripts/src/insights.js
@@ -15,6 +15,7 @@ import open_first_section from "./modules/insights/accordion_functionality/open_
 import remove_aria_roles from "./modules/insights/accordion_functionality/remove_aria_roles";
 import remove_event from "./modules/insights/remove_event";
 import set_active from "./modules/insights/navigation/set_active";
+import set_heading_positions from "./modules/insights/accordion_functionality/set_heading_positions";
 import throttle from "./modules/throttle";
 
 document.addEventListener('DOMContentLoaded', () => {
@@ -23,13 +24,16 @@ document.addEventListener('DOMContentLoaded', () => {
     audio_tracking();
     video_tracking();
     add_unique_ids();
+});
 
+window.addEventListener('load', () => {
     // Initialise jquery
     window.$ = $;
 
     const $sectionHeadings = $(".section-separator__heading");
     const $sectionContents = $(".section-content");
     const $jumplinks = $(".jumplink");
+    let headingPositions;
 
     // These booleans are used to detect if certain enhancements (e.g. click event listeners) have been applied in order to 
     // prevent events from being added more than once.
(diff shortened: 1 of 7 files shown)


scripts/src/modules/insights/accordion_functionality/set_heading_positions.js
```diff
@@ -0,0 +1,29 @@
...
+    sectionHeadings.each(function(index) {
+        if($(this).attr("id") === $(sectionHeadings[0]).attr("id")) {
+            positions.push($(this).offset().top);
+        }
+        else {
+            const prevHeading = sectionHeadings[index - 1];
+
+            // This accounts for any padding, margin, and borders that have been applied to the heading in order to correctly
+            // calculate its height.
+            const sectionHeadingPadding = convert_px_property_to_number($(prevHeading).css("padding-top")) + convert_px_property_to_number($(prevHeading).css("padding-bottom"));
```
AshTNA, line 18: Could you use jQuery's outerHeight here? It looks like that would calculate the height with padding, border, and margin for you:

https://www.w3schools.com/jquery/html_outerheight.asp

https://api.jquery.com/outerheight/

scripts/src/insights.js
```diff
@@ -15,48 +15,56 @@ import open_first_section from "./modules/insights/accordion_functionality/open_
...
-    audio_tracking();
-    video_tracking();
-    add_unique_ids();
-
     // Ids are added to sections for ARIA purposes.
     add_section_ids($sectionHeadings, $sectionContents);
-
-    if($(window).width() < 768 && !mobileEnhancementsApplied) {
+    
+    if($(window).width() < 769 && !mobileEnhancementsApplied) {
```
AshTNA, line 46: A minor thing, but using `<= 768` might be more readable wherever you're doing this, especially when fixing the 768px bug in the other comment (though I appreciate it will be changed to REM anyway)

AshTNA (changes requested):
Hi @JamesWChan 

Nice work using the heading positions to fix the bug. I've spotted one thing to change and suggested a minor change. 

There's also an issue at exactly `768px`. It looks like the desktop Javascript doesn't work, but the mobile JavaScript is being applied to the desktop styles, as when you click a heading it tries to slide. If you're struggling when you try to fix this, I had a similar issue with my JS/SASS, and Gwyn fixed it by making all the SASS classes use mobile styles by default, then only used media queries to apply desktop styles. So there were no  `max-width` media queries, only `min-width: desktop` media queries. But it might be simpler to fix your issue, so you can ignore this way of fixing it.

Thanks :+1:

JamesWChan:
Thank you @AshTNA, I've addressed those comments but I'm going to delete this branch and raise a new PR because the merge conflicts are messy and it would be simpler just to create a new branch and put my fixes on top.

PR #352, opened 2022-05-26
Fix/insights fe fixes

JamesWChan:
Hi @AshTNA,

Would it be possible to merge this PR? It adds a fixed height (`100vh`) and `overflow: scroll` to the desktop navigation in case it gets too long for the viewport. It also converts places where I've used `px` to `rem`. Thank you 👍


AshTNA (approved):
All good :+1:

PR #355, opened 2022-05-27
SSO: Auth0 integration

ababic:
Adds basic integration with Auth0, allowing users from the connected Auth0 tenant to authenticate using the Auth0 login page.

If the user HASN'T authenticated with the app before, successful authentication results in the creation of a basic Django user with an unusable password`*` and no special privileges. This user then becomes the active user for the remainder of the app session.

If the user HAS authenticated with the app before, successful authentication results in the existing Django user being found (via a linked `IDPProfile` object with a matching `provider_user_id` value), and becoming the active user for the remainder of the app session

`*` Setting an 'unusable password' in Django prevents the user from logging in via Django's regular email/password mechanisms, and also prevents them from initialising a password reset from Django - meaning they can only authenticate via Auth0.


etna/auth0/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    path("login/", views.login, name="account_login"),
+    path("logout/", views.logout, name="account_logout"),
+    path("signup/", views.register, name="account_signup"),
+    path("authorize/", views.authorize, name="account_authorize"),
+]
```
ababic, line 10: NOTE: The intention here is to mimic `allauth` where possible in terms of names / paths, so that you can always use `django.urls.reverse()` or the `{% url 'name_of_url' %}` tag to link to the key views, regardless of whether AUTH0 is configured/enabled.

PR #357, opened 2022-05-30
Feature/access 157 all iaid patterns

mozaki1:
1. Changing Regex to accept the required iaid format.
2. Add unit tests , testing iaid GUID format and long format like

### DIFF
--- a/etna/records/tests/test_iaid.py
+++ b/etna/records/tests/test_iaid.py
@@ -1,0 +1,16 @@
+from django.test import TestCase
+from django.urls import resolve
+
+
+class TestiaidFromats(TestCase):
+    def test_iaid_long_mixed_format(self):
+        longformat = "3717ee38900740728076a61a398fcb84"
+        resolver = resolve("/catalogue/id/" + longformat + "/")
+        self.assertEqual(resolver.view_name, "details-page-machine-readable")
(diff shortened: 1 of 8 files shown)


etna/records/tests/test_iaid.py
```diff
@@ -0,0 +1,16 @@
+from django.test import TestCase
+from django.urls import resolve
+
+
+class TestiaidFromats(TestCase):
```
TNA-Allan, line 5:
```suggestion
class TestIaidFormats(TestCase):
```

TNA-Allan: Could we have Test case for iaid=A13530124, C2341693, D431198, F257629 as per the ticket.
Also add N14562581

TNA-Allan (changes requested):

etna/records/tests/test_iaid.py
mozaki1: Yes , I also wanted to do that, but I found it hard to give a good name to these tests, and I don't want to put them all in a for loop also, can you please suggest the names for those tests?

TNA-Allan: As discussed, lets go with your suggestion.

TNA-Allan (approved):
Looks Good.

PR #358, opened 2022-05-30
UN 181: Create global dismissible 'about beta' message

JamesWChan:
Hi @gtvj,

Would it be possible to have a look at this PR? It adds the beta banner to the site which can be dismissed using a JS created button. Clicking the dismiss button will set a cookie so that the banner remains dismissed for the remainder of the session. 

I've tested in Chrome, Firefox, Edge, Safari, and IE11 (and in mobile and high zoom levels).

Thank you 👍


JamesWChan:
Hi @gtvj,

No worries 👍  Agreed, this isn't urgent so this PR can stay open.

Re the backend work: the ticket is now with Mo but I believe he needs to check with Andy (when he's back from leave) regarding where the banner is to be implemented (there was a discussion around whether it should go in alerts or in site settings. I'm not familiar with the different approaches though so I can't elaborate further).

gtvj:
Thanks @JamesWChan

@mozaki1 - grateful if you could let me know what's planned. It does seem to me that the cleanest way to bring in the CMS integration is via this PR.

PR #360, opened 2022-06-07
Moving iaid check above ref check in records_tags.py

mozaki1:


etna/records/templatetags/records_tags.py
```diff
@@ -16,18 +16,19 @@ def record_url(record: Union[Record, Dict[str, Any]]) -> str:
...
+            return reverse("details-page-machine-readable", kwargs={"iaid": iaid})
+        except NoReverseMatch:
+            pass
     if ref := record.get("reference_number", record.get("referenceNumber")):
         try:
             return reverse(
                 "details-page-human-readable", kwargs={"reference_number": ref}
             )
         except NoReverseMatch:
             pass
```
TNA-Allan, line 31: Perhaps commenting as to why its important to have this sequence of conditions and not otherwise. Is it important to give preference of Iaid over Reference in a situation where Reference returns a one record, do all links show Iaid? If not, then the if conditions must be handled exclusively for each url

ababic: I assumed, probably because of the name, that we were preferring the 'human readable' URLs wherever the reference number is available. This is a significant change, as those URLs are unlikely to be seen after this, and the disambiguation view isn't really needed any longer either. We need some input from @sdwilkes here as to what the expected experience is for the user.

TNA-Allan: @sdwilkes , we discussed today standup to go ahead with this PR to make Iaid as priority and if necessary to handle exclusively Disabiguation page with multiple records in a separate PR. So going ahead with Approval.

TNA-Allan: @mozaki1 would it be ok to have a comment something like `#Handling of Iaid as priority to allow Iaid in disambiguation pages when returning more than one record`

PR #366, opened 2022-06-09
Add configurable banner to settings 

mozaki1:


templates/includes/beta-banner.html
```diff
@@ -1,9 +1,11 @@
...
-        This beta site contains new services and features currently in development. Many of these features are works in 
-        progress and are being updated regularly.  You can help us improve them by providing feedback as you use the site.
-    </p>
+     {% if settings.core.SiteSettings.banner_text %}
+        <p class="beta-banner__text">
+            {{ settings.core.SiteSettings.banner_text|richtext }}
+        </p>
+     {% endif %}
     <div class="beta-banner__options">
         <a class="beta-banner__link" href="/">Find out more</a>
```
JamesWChan, line 10: Would it be possible to make this link text editable as well?

etna/core/models/settings.py
```diff
@@ -12,15 +13,23 @@
...
-    panels = [
-        InlinePanel(
-            "main_menu_items_rel",
-            heading="main menu items",
-            label="item",
-            min_num=1,
-            max_num=6,
-        ),
-    ]
+    stand_first = models.CharField(max_length=200, blank=False, null=True)
```
ababic, line 16: As per guidance in the [Django docs](https://docs.djangoproject.com/en/4.0/ref/models/fields/#null)), `null=True` is best avoided on `CharField`s (also true for `URLField` and `RichTextField`).

ababic: Also the name `stand_first` is a little ambiguous. I would suggest `beta_banner_standfirst` for this field, and add `verbose_name="standfirst"` to avoid repeating the 'beta banner' text in the field labels (we'll use a `MultiFieldPanel` to group them under a heading)

etna/core/models/settings.py
```diff
@@ -12,15 +13,23 @@
...
+    banner_link = models.URLField(blank=True, null=True, help_text="Banner URL")
```
ababic, line 17: We should lose the `null=True` here, and again look to make the field name less ambiguous. I think `beta_banner_link` would be good. I would question the usefulness of the `help_text` here too. Is that supposed to be `verbose_name` instead of `help_text`?

etna/core/models/settings.py
```diff
@@ -12,15 +13,23 @@
...
+                "main_menu_items_rel",
+                heading="main menu items",
+                label="item",
+                min_num=1,
+                max_num=6,
+            ),
+        ]
+        + [FieldPanel("stand_first")]
+        + [FieldPanel("banner_link")]
+        + [FieldPanel("banner_text")]
```
ababic, line 31: I'm a little bit confused about the joining of multiple lists here. Usually panels would just be a single list. Generally speaking, it usually makes for a better interface to have fixed fields first, with inline panels at the end (as they hog quite a bit of screen real-estate). We can also use `MultiFieldPanel` to better group the fields together, like so:

```suggestion
    panels = [
        MultiFieldPanel(
            heading="beta banner",
            children=[
                FieldPanel("beta_banner_standfirst"),
                FieldPanel("beta_banner_text"),
                FieldPanel("beta_banner_link"),
            ]
        ),
        InlinePanel(
            "main_menu_items_rel",
            heading="main menu items",
            label="item",
            min_num=1,
            max_num=6,
        ),
    ]
```

etna/core/models/settings.py
```diff
@@ -12,15 +13,23 @@
...
-            "main_menu_items_rel",
-            heading="main menu items",
-            label="item",
-            min_num=1,
-            max_num=6,
-        ),
-    ]
+    stand_first = models.CharField(max_length=200, blank=False, null=True)
+    banner_link = models.URLField(blank=True, null=True, help_text="Banner URL")
+    banner_text = RichTextField(blank=True, null=True)
```
ababic, line 18:
```suggestion
    beta_banner_text = RichTextField(blank=True, verbose_name="text")
```

templates/includes/beta-banner.html
```diff
@@ -1,10 +1,19 @@
...
+         <div class="beta-banner__standfirst">
+         {{ settings.core.SiteSettings.stand_first }}
+         </div>
+     {% endif %}
+
+     {% if settings.core.SiteSettings.banner_text %}
+        <div class="beta-banner__text">
+            {{ settings.core.SiteSettings.banner_text|richtext }}
+        </div>
+     {% endif %}
```
ababic, line 13:
```suggestion
     <div class="beta-banner__standfirst">
         {{ settings.core.SiteSettings.beta_banner_standfirst|default:">What does the 'beta' label mean?" }}
     </div>
     <div class="beta-banner__text">
         {% if settings.core.SiteSettings.beta_banner_text %}
             {{ settings.core.SiteSettings.beta_banner_text|richtext }}
         {% else %}
             <p>This beta site contains new services and features currently in development. Many of these features are works in progress and are being updated regularly.  You can help us improve them by providing feedback as you use the site.</p>
         {% endif %}
     </div>
```

templates/includes/beta-banner.html
```diff
@@ -1,10 +1,19 @@
...
+    {% if settings.core.SiteSettings.banner_link %}
+        <div class="beta-banner__options">
+            <a class="beta-banner__link" href="{{ settings.core.SiteSettings.banner_link }}">Find out more</a>
+        </div>
+    {% endif %}
```
ababic, line 18:
```suggestion
    <div class="beta-banner__options">
        <a class="beta-banner__link" href="{{ settings.core.SiteSettings.beta_banner_link }}">Find out more</a>
    </div>
```

ababic (changes requested):

ababic:
@mozaki1 Could we maybe remove these new migrations and regenerate a single one when we have all the fields we need? Ideally, it would be good to see a single `0002_add_beta_banner_fields.py` migration. 

To do this locally, you would first run the following to reverse the current migrations:

```
python manage.py migrate core 0001
```

Then you can safely remove the existing files and regenerate using the `-n`  option:

```
python manage.py makemigrations core -n add_beta_banner_fields
```

templates/includes/beta-banner.html
```diff
@@ -1,10 +1,14 @@
...
-    </div>
+
+         <div class="beta-banner__standfirst">
+         {{ settings.core.SiteSettings.beta_banner_standfirst }}
+         </div>
+        <div class="beta-banner__text">
+            {{ settings.core.SiteSettings.beta_banner_text|richtext }}
+        </div>
+        <div class="beta-banner__options">
+            <a class="beta-banner__link" href="{{ settings.core.SiteSettings.beta_banner_link }}">Find out more</a>
```
JamesWChan, line 11: Could we move `{{ settings.core.SiteSettings.beta_banner_link }}` to the link text i.e. replace “Find out more”?

With regards to the `href`, it can stay as “/” for now because I’m not too sure where the link is going to go; it’s either going to go to a survey or the about page (as per a conversation I had with Lauren yesterday) and I need to confirm.

JamesWChan: @ababic - if the link takes the user to a third party website (as could be the case with the survey), we'll need to add `rel="noopener noreferrer"` and potentially `target="_blank"` to the `<a>` tag. Eventually, once the feedback mechanism has been completed, I believe the idea would be to always point to the About page thereafter. So I was thinking I could just implement the above to avoid creating additional potential backend complexity. But if it's straightforward to conditionally render the link attributes then happy to go with your suggestion.

templates/includes/beta-banner.html
```diff
@@ -1,9 +1,11 @@
+{% load wagtailcore_tags %}
 <div class="beta-banner" aria-label="About The National Archives' beta">
```
gtvj, line 2: I'm not sure that `aria-label` is needed here because the `<div>` itself is not an interactive element.

JamesWChan: Thanks @gtvj, I can remove this. I did see this pattern used elsewhere in Etna so we might want to look at those too as well.

gtvj:
This is a general question about the overall approach here but I'd be interested to know why we've opted to implement this with `SiteSettings` rather than implementing the banner as a [Snippet](https://docs.wagtail.org/en/stable/topics/snippets.html). I'd imagine there will ultimately be many such banners introduced to the site (even just for the different phases such as Alpha) and wonder if `SiteSettings` might get a bit crowded and confusing.

ababic:
There are also compelling reasons for extending SiteSettings for this purpose:
- The relevant model instance is already being fetched for every request, so we can access these field values without any additional database queries
- Site settings are generally easier for editors to grasp. It's more self-explanatory "here are some site-wide things that don't change very often". Adding multiple snippet models adds a degree of 'horizontal complexity' to the CMS, the Snippet list becomes longer, and editors have to remember what each thing does.

In all likelihood, these fields will be removed when it's time to introduce an alpha banner instead, so I'm not particularly worried about the form becoming unwieldy. Even if we do end up adding more fields, with some clever use of MultiFieldPanel and Tabs, it doesn't have to be messy.

templates/includes/beta-banner.html
ababic: @JamesWChan it should be relatively simple to write a filter to detect if the URL goes to an external site, and conditionally add those bits in if so.

etna/core/models/settings.py
```diff
@@ -12,7 +13,35 @@
...
 class SiteSettings(BaseSetting, ClusterableModel):
+    beta_banner_standfirst = models.CharField(
+        max_length=200,
+        default="What does the 'beta' label mean?",
+        verbose_name="standfirst",
+    )
+    beta_banner_link = models.URLField(blank=True, verbose_name="link to")
+    beta_banner_link_text = models.CharField(
+        max_length=200,
+        default="Find out more ?",
```
ababic, line 24: where did this question mark come from, @mozaki1? probably best that we lose it:
```suggestion
        default="Find out more",
```

etna/core/models/settings.py
```diff
@@ -12,7 +13,35 @@
...
+        verbose_name="linktext",
```
ababic, line 25: `verbose_name` is what ultimately becomes the field label in the form. While `standfirst` is a recognised term, I'm not sure the same can be said for `"linktext"`, so perhaps we should change to:
 ```suggestion
        verbose_name="link text",
```

File: templates/includes/beta-banner.html [hunk 12]
```diff
@@ -1,10 +1,15 @@
...
-    </div>
+
+         <div class="beta-banner__standfirst">
+         {{ settings.core.SiteSettings.beta_banner_standfirst }}
+         </div>
+        <div class="beta-banner__text">
+            {{ settings.core.SiteSettings.beta_banner_text|richtext }}
+        </div>
+        <div class="beta-banner__options">
+            <a class="beta-banner__link" href="{{ settings.core.SiteSettings.beta_banner_link }}">
```
ababic, line 11: @JamesWChan Could the `rel="noopener noreferrer"` be added with JS? That feels like more of a front-end concern than a backend one. This should be easy to detect with something like:

```
const isExternalURL = (url) => new URL(url).origin !== location.origin;
```

etna/core/models/settings.py
```diff
@@ -12,7 +13,35 @@
 
 @register_setting(icon="list-ul")
 class SiteSettings(BaseSetting, ClusterableModel):
+    beta_banner_standfirst = models.CharField(
+        max_length=200,
+        default="What does the 'beta' label mean?",
+        verbose_name="standfirst",
+    )
+    beta_banner_link = models.URLField(blank=True, verbose_name="link to")
```
ababic, line 21: @mozaki1 If we need to accommodate bot internal (without scheme/domain) or external (with scheme/domain) here, we'll want to go with `models.CharField` instead. `models.URLField()` fails validation when the url doesn't include the scheme/domain.

templates/includes/beta-banner.html
JamesWChan: @ababic - it can’t I’m afraid. As the attribute would be part of the core experience (i.e. not an enhancement), we can’t rely on JS to deliver it.

etna/core/models/settings.py
mozaki1: Hi Andy, based on  https://docs.djangoproject.com/en/4.0/ref/forms/widgets/#django.forms.URLInput

Summary below :

1. Normalizes to: A string.
2. Uses [URLValidator] to validate that the given value is a valid URL.
3. Error message keys: required, invalid
4. Has three optional arguments max_length, min_length, and empty_value which work just as they do for [CharField]

Number 2 may be relevant, which is not available in CharField , and since we know this is a URL , maybe this is a better option? after all this is what is recommended.

ababic: @mozaki1 try sticking in `/about-us/` and see what happens :)

templates/includes/beta-banner.html
ababic: But `noreferrer` is an SEO nicety thing, right? It's not important to the core user experience (which is what progressive enhancement is all about).  And `noopener` carries some security benefits, but is probably unnecessary in this case, as the target will always be trusted.

JamesWChan: As per Section 9 of the GDS Service Manual:

“Service teams should actively identify security and privacy threats to the service, and have a robust, proportionate approach to securing information and managing fraud risks.”

Linking to a third party is a potential security threat. Yes, the third party we link to would most likely be fine, which is why I'm suggesting a proportionate approach (i.e. adding just a small bit of code which you've said should be simple to implement).

This would mean that we'd be doing everything we can to make the core experience of the highest quality (which is also what progressive enhancement is all about).

Source - https://www.gov.uk/service-manual/service-standard/point-9-create-a-secure-service

ababic: I'm keen to identify the _right_ approach here @JamesWChan. Is this only a problem with the banner link, or a wider issue? Is a backend solution capable of solving this problem everywhere it needs to be solved? I'm all for writing code, I've just seen enough 'temporary' solutions on this project to last me a life time.

etna/core/models/settings.py
```diff
@@ -12,7 +13,37 @@
...
+        ),
+    )
     panels = [
+        MultiFieldPanel(
+            heading="BETA banner",
+            children=[
+                FieldPanel("beta_banner_standfirst"),
+                FieldPanel("beta_banner_link"),
+                FieldPanel("beta_banner_link_text"),
+                FieldPanel("beta_banner_text"),
```
ababic, line 44: @mozaki1 One last/tiny thing: I find it helps editors to see fields in the order they are used in the output. Since the link/button comes at the end, a better order might be:
```suggestion
                FieldPanel("beta_banner_text"),
                FieldPanel("beta_banner_link"),
                FieldPanel("beta_banner_link_text"),
                
```

ababic (approved):
LGTM - thanks @mozaki1

PR #367, opened 2022-06-15
Display Insights on Collection Explorer topic and time period pages

mozaki1:


etna/collections/models.py
```diff
@@ -124,6 +126,16 @@ def results_pages(self):
     ]
     subpage_types = ["collections.TopicExplorerPage", "collections.ResultsPage"]
 
+    def get_context(self, request):
```
ababic, line 129: @mozaki1 Did this work without errors when you viewed one of these pages?

etna/collections/models.py
```diff
@@ -124,6 +126,16 @@ def results_pages(self):
...
+        context = super().get_context(request)
+        context["insights_topics"] = (
+            InsightsPage.objects.filter(topic=self)
+            .live()
+            .order_by("title")
+            .specific()[:3]
```
ababic, line 135: Are the insights displayed with images? Could it be worth using `select_related()` to optimise things?

templates/collections/time_period_explorer_page.html
```diff
@@ -14,8 +14,23 @@
                 {% endfor %}
             </div>
         </div>
+        {% if FEATURE_RELATED_INSIGHTS_ON_EXPLORE_PAGES %}
+        <div class="row">
+            <div class="col-md-12">
+                 {% for insights_page in insights_time_periods %}
+                {% with page=insights_page.specific %}
```
ababic, line 21: This `with` isn't needed, as the queryset returns instances of `InsightPage` already

etna/collections/models.py
```diff
@@ -205,6 +217,16 @@ def results_pages(self):
     ]
     subpage_types = ["collections.TimePeriodExplorerPage", "collections.ResultsPage"]
 
+    def get_context(self, request):
+        context = super().get_context(request)
+        context["insights_time_periods"] = (
+            InsightsPage.objects.filter(time_period=self)
+            .live()
+            .order_by("title")
+            .specific()[:3]
```
ababic, line 226: You only need to use `specific()` when starting the queryset from the generic `Page` model manager, e.g. `Page.objects.all()`. This queryset will already be 'specific', because you're using the `InsightsPage` model manager.

etna/collections/models.py
ababic, line 135: You only need to use specific() when starting the queryset from the generic Page model manager, e.g. Page.objects.all(). This queryset will already be 'specific', because you're using the InsightsPage model manager.

templates/collections/time_period_explorer_page.html
```diff
@@ -14,8 +14,23 @@
...
+                     <li>
+                       <a href="{{ page.url }}">{{ page.title }}</a>
```
ababic, line 23: Always use the `{% pageurl %}` tag to generate page urls, where possible:
```suggestion
                       <a href="{% pageurl page %}">{{ page.title }}</a>
```

etna/collections/models.py
```diff
@@ -205,6 +217,16 @@ def results_pages(self):
     ]
     subpage_types = ["collections.TimePeriodExplorerPage", "collections.ResultsPage"]
 
+    def get_context(self, request):
```
ababic, line 220: Can you think of an alternative to overriding `get_context()`, that might be neater?

templates/collections/topic_explorer_page.html
```diff
@@ -14,7 +15,23 @@
...
         </div>
-    </div>
+
+    {% if FEATURE_RELATED_INSIGHTS_ON_EXPLORE_PAGES %}
+        <div class="row">
+            <div class="col-md-12">
+                 {% for insights_page in insights_topics %}
+                     {% with page=insights_page.specific %}
+                     <li>
+                       <a href="{{ page.url }}">{{ page.title }}</a>
```
ababic, line 25: Always use the `{% pageurl %}` tag to generate page urls, where possible:
```suggestion
                       <a href="{% pageurl page %}">{{ page.title }}</a>
```

etna/collections/models.py
```diff
@@ -205,6 +217,16 @@ def results_pages(self):
...
+        context = super().get_context(request)
+        context["insights_time_periods"] = (
```
ababic, line 222: Maybe `related_insights` would be a sensible name?

templates/collections/topic_explorer_page.html
```diff
@@ -14,7 +15,23 @@
                 {% endfor %}
             </div>
         </div>
-    </div>
+
+    {% if FEATURE_RELATED_INSIGHTS_ON_EXPLORE_PAGES %}
+        <div class="row">
+            <div class="col-md-12">
+                 {% for insights_page in insights_topics %}
+                     {% with page=insights_page.specific %}
```
ababic, line 23: Another unnecessary `{% with %}` here - the pages are already specific

etna/collections/models.py
mozaki1: Yes it did :)

etna/collections/models.py
mozaki1: This is what Gasman says : 
https://stackoverflow.com/questions/32626815/wagtail-views-extra-context

ababic: Did you know you can access all fields values, methods and properties from the page directly with something like `{{ page.attribute_name }}`? If you broke the queryset stuff out into a separate method like this:

```
    @cached_property
    def related_insights(self):
        return InsightsPage.objects.filter(time_period=self).live().order_by("title")[:3]
```

You could easily access this from the template like so:

```
{% for obj in page.related_insights %}
    ...
{% endfor %}
```

ababic: ^ This also allows you to access the queryset for other purposes too (i.e. not just when rendering the page).

mozaki1: Yes, it makes sense :)

ababic:
@mozaki1 Is there a separate FE task for styling these bits up? Remember to check the design to see if any images are going to be needed, and if so, optimise the queryset appropriately.

PR #372, opened 2022-06-21
UN-213: Front end tweaks

JamesWChan:
Hi @matt-blair,

Would it be possible to have a look at this PR? It implements some Insights tweaks that were requested by the team. It's mostly just CSS changes but I have changed the implementation of the featured insights image to be a `<div>` with a background as opposed to using the `<picture>` element with several `<source>`s. I did this because the `<picture>` implementation was dependent on explicitly defining image dimensions in pixels and I don't think it responds well depending on the screen size. I found going with the `<div>` with a background approach was more responsive (happy to have a call to demo and explain this further, it's a bit tricky to explain over text). Thank you 👍

### DIFF
--- a/sass/includes/_featured-insight.scss
+++ b/sass/includes/_featured-insight.scss
@@ -32,12 +32,11 @@
 
     @media (max-width: $screen__md) {
         display: block;
-        &__image, &__description {
+        &__description, &__image{
             width: 100%;
-
-            img {
-                width: 100%;
-            }
+        }
+        &__image {
+            height: 21.875rem;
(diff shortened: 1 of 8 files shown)


matt-blair (approved):
Thanks @JamesWChan. I've checked the code and we've been through this on a call to talk through the rationale behind the choices around handling images responsively via CSS. Happy with this approach.

JamesWChan:
I discussed the image implementation with @gtvj and he was also okay with the approach so I'm going to merge this

PR #376, opened 2022-06-24
CHORE: Add tests for record_url tag

ababic:
Pre-requisite of #345, so that we can see the impact those changes have on `{% record_url %}` results.

### DIFF
--- a/etna/records/tests/test_templatetags.py
+++ b/etna/records/tests/test_templatetags.py
@@ -6,7 +6,10 @@ from etna.records.templatetags.records_tags import record_url
 
 class TestRecordURLTag(SimpleTestCase):
 
+    # A sample search result respresentation, as encountered in search views
+    # (where no up-front transformation is applied)
     record_search_hit = {
+        "_score": 1,
         "_source": {
             "@template": {
                 "details": {
@@ -18,7 +21,10 @@ class TestRecordURLTag(SimpleTestCase):
         },
     }
 
+    # A sample search result respresentation, as encountered in search views
+    # (where no up-front transformation is applied)
     interpretive_search_hit = {
+        "_score": 1,
         "_source": {
             "@template": {
                 "details": {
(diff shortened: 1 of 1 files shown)


etna/records/tests/test_templatetags.py
```diff
@@ -0,0 +1,133 @@
...
+        reference_number="2515/300/1",
+        title="Test",
+    )
+
+    record_instance_no_reference = Record(
+        iaid="e7e92a0b-3666-4fd6-9dac-9d9530b0888c",
+        reference_number="",
+        title="Test",
+    )
+
```
TNA-Allan, line 55: `record_search_hit`, `interpretive_search_hit` gives us an eye at the API response. 
Ideally, would appreciate if the same done with these. Maybe extract the partial bit from a `record json`. _The names of those variables arn't sufficient enough to know from where the partial bit is taken from._
Alternatively, a `#comment` mentioning the path followed for the partial bit.

TNA-Allan: > Happy to add a few comments if you think it'll help
Please do then. Thanks Andy.

ababic: Are these sufficient @TNA-Allan? [a0de236](https://github.com/nationalarchives/ds-wagtail/pull/376/commits/a0de236c88b4961f77ccb9bbd14ad924fcb70c3c)

TNA-Allan: `record_instance_no_reference` ... `reference_number=""`
Does API return value with empty reference?

ababic: Well, what this is really aiming to replicate is what is currently prepared by `transform_record_result()`. Looks like the value is typically `None` when not found in the representation, so I can change this to reflect that.

TNA-Allan: OK, please do

TNA-Allan (approved):
Thanks Andy. Looks Good

PR #377, opened 2022-06-24
ACCESS-176: Replace iaids with reference numbers in record hierarchical nav

ababic:
Dependent on #345 (must be merged first)

Implemented as per approach agreed on Slack thread: https://torchbox.slack.com/archives/C027G6UR8N4/p1656075785897039


templates/includes/records/hierarchy-local.html
```diff
@@ -29,7 +35,13 @@
                     <p class="hierarchy-local__nav-item">
                         <span class="hierarchy-local__nav-indicator hierarchy-local__nav-indicator--right"></span>
                         <span class="hierarchy-local__nav-label">{{ 'next_record'|as_label }}</span>
-                        <a href="{% record_url page.next_record %}" data-link="Next" class="link--bold"><span class="sr-only">{{ 'next_record'|as_label }}</span> {{ page.next_record.iaid }}</a>
+                        <a href="{% record_url page.next_record %}" data-link="Next" class="link--bold">
+                            {% if page.next_record.has_reference_number %}
+                                <span class="sr-only">{{ 'next_record'|as_label }}</span> {{ page.next_record.reference_number }}
+                            {% else %}
+                                Next {{ page.level.lower|default:"record" }}
```
TNA-Allan, line 42: Similar to Previous comment - Next label is already present above `{{ 'next_record'|as_label }}`

templates/includes/records/hierarchy-local.html
```diff
@@ -8,7 +8,13 @@
                     <p class="hierarchy-local__nav-item">
                         <span class="hierarchy-local__nav-indicator hierarchy-local__nav-indicator--left"></span>
                         <span class="hierarchy-local__nav-label">{{ 'previous_record'|as_label }}</span>
-                        <a href="{% record_url page.previous_record %}" data-link="Previous" class="link--bold"><span class="sr-only">{{ 'previous_record'|as_label }}</span> {{ page.previous_record.iaid }}</a>
+                        <a href="{% record_url page.previous_record %}" data-link="Previous" class="link--bold">
+                            {% if page.previous_record.has_reference_number %}
+                                <span class="sr-only">{{ 'previous_record'|as_label }}</span> {{ page.previous_record.reference_number }}
+                            {% else %}
+                                Previous {{ page.level.lower|default:"record" }}
```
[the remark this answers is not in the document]
ababic: It's important that the 'Previous' text is part of the link, as that will be read out to screen readers

templates/includes/records/hierarchy-local.html
ababic: It's important that the 'Next' text is part of the link, as that will be read out by screen readers

templates/includes/records/hierarchy-local.html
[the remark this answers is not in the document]
TNA-Allan: > It's important that the 'Previous' text is part of the link, as that will be read out to screen readers

1. Why do we not use `{{ 'previous_record'|as_label }}` instead of Hard coded label?
2. Do you want to also add it to existing code where reference number is used?
3. `page.level.lower` could we just retain the casing returned in the response.

Same comments apply for `next record' as they are both similar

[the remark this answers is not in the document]
ababic: 1. Have updated to use `{{ 'previous_record'|as_label }}` and `{{ 'next_record'|as_label }}`
2. No, the better approach is to hide the 'hidden on mobile' label, and always have "Next thing" or "Previous thing" as the visible link text
3. GDS tend to prefer title casing, and since these are standard archival terms, and not people or place names, they shouldn't be capitalised. Waiting for confirmation on Slack thread.

etna/records/models.py
```diff
@@ -174,6 +174,13 @@ def title(self) -> str:
             pass
         return self.get("summary.title", default="")
 
+    @cached_property
+    def is_tna(self):
```
TNA-Allan, line 178: Could we have a test case for this new property?

ababic: Unit tests added in b57a855 :)

TNA-Allan (approved):
Thank you @ababic

PR #378, opened 2022-06-27
Add cookie consent frontend

Puththiran:


templates/base.html
```diff
@@ -70,5 +78,7 @@
         {% endblock %}
         <script src="{% static 'scripts/global_search.js' %}"></script>
         <script src="{% static 'scripts/hamburger_menu.js' %}"></script>
+        <script src="{% static 'cookie-consent/ds-cookie-consent.js' %}"></script>
+
```
ababic, line 82: This code is cookie related, so might need to be wrapped in a conditional.

templates/base.html
```diff
@@ -47,9 +50,14 @@
...
     <body class="{% block body_class %}{% endblock %}">
+        {% if FEATURE_COOKIE_BANNER_ENABLED %}
+            {% include 'includes/cookie-consent.html' %}
+        {% endif %}
         <a href="#maincontent" class="skip-to-content-link sr-only sr-only-focusable" data-link="Skip to main content">Skip to main content</a>
 
-        {% include 'includes/gtm-no-script.html' %}
+        {% if FEATURE_COOKIE_BANNER_ENABLED %}
+            {% include 'includes/gtm-no-script.html' %}
+        {% endif %}
```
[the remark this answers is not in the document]
Puththiran: corrected, this is now shown only if the usage is set to true.

templates/base.html
Puththiran: corrected, this is now included to show based on the value FEATURE_COOKIE_BANNER_ENABLED

etna/core/templatetags/sections_tags.py
```diff
@@ -20,3 +25,20 @@ def jumplinks(page):
...
+    """
+    try:
+        usage = False
+        if "cookies_policy" in record:
+            cookie_str = unquote(record["cookies_policy"])
+            # Use Record property if available
+            usage = json.loads(cookie_str)["usage"]
+    except JSONDecodeError:
+        usage = False
+    return usage
```
ababic, line 44: I'm not sure `section_tags.py` is the right place for this particular functionality. I would suggest adding a new module for this, e.g. `cookie_tags.py`

etna/core/templatetags/sections_tags.py
```diff
@@ -20,3 +25,20 @@ def jumplinks(page):
...
+
+@register.filter
+def cookie_usage(record: dict) -> bool:
+    """
+    Return the True/False based on cookie usage value
+    if no cookie set it will return False
+    """
+    try:
+        usage = False
+        if "cookies_policy" in record:
```
ababic, line 38: A little tweak here will cater for when the value is `None`
```suggestion
        if record and "cookies_policy" in record:
```

ababic:
@Puththiran nice work! the only thing left here is to add a few unit tests. We don't necessarily need to use the tag in a template and render it to check the result (it is just a regular python function when all is said and done). You may find these examples helpful for inspiration: https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/records/tests/test_templatetags.py

etna/core/tests/test_cookie_tags.py
```diff
@@ -0,0 +1,27 @@
...
+
+class TestCookieUsePermittedTag(SimpleTestCase):
+
+    empty_cookie = ""
+    none_cookie = None
+    json_usage_true_cookie = '{"usage":true,"settings":false,"essential":true}'
+    json_unicoded_usage_true_cookie = (
+        "%7b%22usage%22%3atrue%2c%22settings%22%3afalse%2c%22essential%22%3atrue%7d"
+    )
+    json_usage_false_cookie = '{"usage":false,"settings":false,"essential":true}'
```
ababic, line 14: Not a bad start, but we should probably try to trigger all of those exceptions we're handling, to make sure it behaves as expected, e.g.
```suggestion
    json_usage_false_cookie = '{"usage":false,"settings":false,"essential":true}'
    invalid_json_cookie = 'NOT_JSON'
    unexpected_json_format = '["item_one", "item_two"]'
```

ababic (changes requested):

etna/core/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class CookieMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        cookie_name = "cookies_policy"
+        if not request.COOKIES.get(cookie_name, None):
```
ababic, line 14:
```suggestion
        if settings.COOKIE_BANNER_ENABLED and not request.COOKIES.get(cookie_name):
```

etna/core/middleware.py
```diff
@@ -0,0 +1,24 @@
+from datetime import datetime, timedelta
+from urllib.parse import quote
+
+from django.conf import settings
+
+
+class CookieMiddleware:
```
ababic, line 7: We can afford to be a little more descriptive with the naming here:

```suggestion
class SetDefaultCookiePreferenceCookieMiddleware:
```

etna/core/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
```
ababic, line 12: We should avoid doing anything when `FEATURE_COOKIE_BANNER_ENABLED` is `False`:

```suggestion
        response = self.get_response(request)
        if not settngs.FEATURE_COOKIE_BANNER_ENABLED:
            return response
```

ababic (changes requested):
Thanks @Puththiran. Just a couple of small things, and we should be there!

ababic (approved):

PR #381, opened 2022-06-29
DF-345:Changes for MAINTENANCE_MODE functionality

TNA-Allan:


etna/core/middleware.py
```diff
@@ -0,0 +1,43 @@
...
+from etna.ciim.utils import get_date_for_retry_after_header
+
+
+def get_client_ip(request) -> str:
+    x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+    if x_forwarded_for:
+        ip = x_forwarded_for.split(",")[0]
+    else:
+        ip = request.META.get("REMOTE_ADDR")
+    return ip
```
ababic, line 13: Any reason why this method is added here, but `get_date_for_retry_after_header()` is added to utils?

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,76 @@
...
+    def setUp(self):
+
+        private_beta_user = User(
+            username="private-beta@email.com", email="private-beta@email.com"
+        )
+        private_beta_user.set_password("password")
+        private_beta_user.save()
+        private_beta_user.groups.add(Group.objects.get(name="Beta Testers"))
+
+        self.client.login(email="private-beta@email.com", password="password")
```
ababic, line 51: Why is a test user needed here, Allan?

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,76 @@
...
+User = get_user_model()
+
+
+@override_settings(
+    MAINTENANCE_MODE=True,
+    MAINTENENCE_MODE_ENDS="2011-11-04T00:05:23+04:00",
+)
+class TestMaintenanceMode(TestCase):
+    @responses.activate
+    @prevent_request_warnings
```
ababic, line 22: What role is `responses` playing here?

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,76 @@
...
+        response = self.client.get("/catalogue/id/C123456/")
+        self.assertEquals(response.status_code, 503)
+        self.assertEquals(
+            response.headers["Retry-After"], "Fri, 04 Nov 2011 00:05:23 GMT"
+        )
+
+    @prevent_request_warnings
+    def test_503_maintenance_override_for_home_page(self):
+        response = self.client.get("/")
+        self.assertEquals(response.status_code, 503)
```
ababic, line 33: How come the `@prevent_request_warnings` decorator is needed here? I'm not sure I follow.

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,76 @@
...
+            responses.GET,
+            "https://kong.test/data/fetch",
+            json=create_response(
+                records=[
+                    create_record(iaid="C123456"),
+                ]
+            ),
+        )
+
+        response = self.client.get("/catalogue/C123456/")
```
ababic, line 70: I think this test is far less clear for for using the record detail view as the view to test. Could we maybe just stick with using the homepage to test with? I think we can assume that the Middleware works at a higher-level than individual views, so the value of testing more than one type of view isn't hugely valuable.

etna/ciim/utils.py
```diff
@@ -175,3 +177,43 @@ def convert_sort_key_to_index(sort):
...
+        previous_logging_level = logger.getEffectiveLevel()
+        logger.setLevel(logging.CRITICAL)
+
+        # trigger original function that would throw warning
+        original_function(*args, **kwargs)
+
+        # lower logging level back to previous
+        logger.setLevel(previous_logging_level)
+
+    return new_function
```
ababic, line 219: I'm not sure about the need for this function, are you able to explain? Even if definitely required, i'm not sure why we're putting either of these functions in the `ciim` app?

etna/core/middleware.py
```diff
@@ -0,0 +1,43 @@
...
+                if settings.MAINTENENCE_MODE_ENDS:
+                    kwargs.update(
+                        {
+                            "headers": {
+                                "Retry-After": get_date_for_retry_after_header(
+                                    settings.MAINTENENCE_MODE_ENDS
+                                )
+                            }
+                        }
+                    )
```
ababic, line 38:
```suggestion
                try:
                    end_datetime = datetime.datetime.fromisoformat(settings.MAINTENENCE_MODE_ENDS)
                except ValueError:
                    end_datetime = None
                if end_datetime:
                    kwargs.update(
                        {
                            "headers": {
                                "Retry-After": end_datetime.strftime(HTTP_HEADER_FORMAT)
                            }
                        }
                    )
```

TNA-Allan: Would it be fine to retain the `if settings.MAINTENENCE_MODE_ENDS` as I think it may be unnecessary overhead for every request to evaluate `end_datetime`

etna/ciim/utils.py
TNA-Allan: This function is used in tests to suppress messages like 503 `Service Unavailable: /`, Not Found, to make the test output clean.  Do you suggest any other location to add it?

etna/core/middleware.py
TNA-Allan: get_date_for_retry_after_header() is removed and refactored in middleware.py

etna/core/tests/test_middleware.py
TNA-Allan: Test case removed, irrelevant now.

etna/core/tests/test_middleware.py
TNA-Allan: Removed and updated test case for home page.

.github/workflows/main.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Check compliance with isort
         run: isort -c .
       - name: Install Black
-        run: python -m pip install black==22.1.0
+        run: python -m pip install black==22.6.0
```
ababic, line 29: Should we move this change to a separate PR?

pyproject.toml
```diff
@@ -21,7 +21,7 @@ wagtail-font-awesome-svg = "0.0.2"
 django-birdbath = "^0.0.5"
 
 [tool.poetry.dev-dependencies]
-black = "^22.1.0"
+black = "^22.6.0"
```
ababic, line 24: Should we move this change to a separate PR?

.github/workflows/main.yml
TNA-Allan: Sure

etna/ciim/utils.py
ababic: Maybe we can just update the logger config in the test settings? There's a complex example here that overrides several Django loggers: https://docs.djangoproject.com/en/4.0/topics/logging/.

etna/core/middleware.py
```diff
@@ -0,0 +1,60 @@
...
+        Maintenanc Mode is overridden with MAINTENENCE_MODE_ALLOW_IPS
+        """
+        if settings.MAINTENANCE_MODE:
+            # check override maintenance mode
+            if get_client_ip(request) not in settings.MAINTENENCE_MODE_ALLOW_IPS:
+                kwargs = {"template": "503.html", "status": 503}
+                if maintenance_mode_ends := settings.MAINTENENCE_MODE_ENDS:
+                    # Evaluate only if config is set
+                    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date, example "Wed, 21 Oct 2015 07:28:00 GMT"
+                    HTTP_HEADER_FORMAT = "%a, %d %b %Y %H:%M:%S GMT"
```
ababic, line 35: Let's move this to the global level (at the top of the module) - it will help simplify the code here.

etna/core/middleware.py
```diff
@@ -0,0 +1,60 @@
...
+                    if end_datetime:
+                        kwargs.update(
+                            {
+                                "headers": {
+                                    "Retry-After": end_datetime.strftime(
+                                        HTTP_HEADER_FORMAT
+                                    )
+                                }
+                            }
+                        )
```
ababic, line 55: For one-off additions it's simpler to do:
```suggestion
                        kwargs["headers"] = {
                            "Retry-After": end_datetime.strftime(HTTP_HEADER_FORMAT)
                        }
```

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,32 @@
...
+@override_settings(
+    KONG_CLIENT_BASE_URL="https://kong.test",
+    MAINTENANCE_MODE=True,
+    MAINTENENCE_MODE_ALLOW_IPS="123.4.5.6",
+)
+class TestMaintenanceModeOverrrice(TestCase):
+    def _mocked_ip(request):
+        return "123.4.5.6"
+
+    @patch("etna.core.middleware.get_client_ip", _mocked_ip)
```
ababic, line 28:
```suggestion
    @patch("etna.core.middleware.get_client_ip", return_value="123.4.5.6")
```

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,32 @@
...
+    @prevent_request_warnings
+    def test_503_maintenance_mode_for_home_page(self):
+        response = self.client.get("/")
+        self.assertEquals(response.status_code, 503)
+
+
+@override_settings(
+    KONG_CLIENT_BASE_URL="https://kong.test",
+    MAINTENANCE_MODE=True,
+    MAINTENENCE_MODE_ALLOW_IPS="123.4.5.6",
```
ababic, line 22: Isn't this setting value usually a list? https://github.com/nationalarchives/ds-wagtail/pull/381/files#diff-f7a16228131ae92457c22657bb33458207a7578b8cdb6ff796153916fa67a22dR323

etna/core/tests/test_middleware.py
```diff
@@ -0,0 +1,32 @@
...
+)
+class TestMaintenanceModeOverrrice(TestCase):
+    def _mocked_ip(request):
+        return "123.4.5.6"
+
+    @patch("etna.core.middleware.get_client_ip", _mocked_ip)
+    def test_503_maintenance_mode_override_for_home_page(self):
```
ababic, line 29: Should we also add a 'negative' test to show that a 503 is still received when the IP is NOT in `settings. MAINTENENCE_MODE_ALLOW_IPS`?

etna/core/middleware.py
ababic: Yes, that'd be fine. In that case, I would probably suggest:

```
                end_datetime = None
                if settings.MAINTENENCE_MODE_ENDS:
                    try:
                        end_datetime = datetime.datetime.fromisoformat(settings.MAINTENENCE_MODE_ENDS)
                    except ValueError:
                        end_datetime = None
                if end_datetime:
                    kwargs["headers"] = {
                        "Retry-After": end_datetime.strftime(HTTP_HEADER_FORMAT)
                    }
```

etna/core/middleware.py
ababic, line 35: One thing I completely forgot to highlight is timezones! The isoformat date may include a timezone indicator, in which case, it will have a timezone other than 'GMT', so hardcoding that doesn't make sense. We should use `%Z` in the string to get a dynamic value here.

etna/core/tests/test_middleware.py
TNA-Allan: Set it to list of sting

etna/core/tests/test_middleware.py
TNA-Allan: Added test `test_maintenance_mode_enforced_when_ip_not_in_allowed_list`

etna/core/middleware.py
TNA-Allan: Done

TNA-Allan: Done

etna/core/tests/test_middleware.py
TNA-Allan: Updated to use mock return value in test body.

ababic (approved):

PR #385, opened 2022-07-01
DF-209: Split datefield into three seperate fields

TNA-Allan:
The record opening date From, To fields are split into individual parts - DD MM YYY 

Inputs allowed -
- YYYY
- MM YYYY
- DD MM YYYY
- Either FROM or TO dates
- Both FROM and TO dates

The missing parts are auto-filled -

- FROM: YYYY -> 01/01/YYYY
- FROM: MM/YYY -> 01/MM/YYYY
- TO: YYYY -> 31/12/YYYY
- TO: MM/YYY -> [28/29/30/31]/MM/YYYY


etna/search/forms.py
```diff
@@ -198,20 +237,238 @@ class BaseCollectionSearchForm(forms.Form):
         required=False,
     )
 
+    def _get_compressed_date_or_none_on_error(
```
ababic, line 240: As much as possible, we should leave the implementation of the method up to the method, and not try to explain it in the name:
```suggestion
    def _get_compressed_date(
```

etna/search/forms.py
```diff
@@ -198,20 +237,238 @@ class BaseCollectionSearchForm(forms.Form):
...
+        self, start_day: str, start_month: str, start_year: str
```
ababic, line 241: There doesn't appear to be anything about this method that is specific to 'start' dates VS 'end' dates or some other date, so maybe less verbose variable names would be appropriate here?
```suggestion
        self, day: str, month: str, year: str
```

etna/search/forms.py
```diff
@@ -198,20 +237,238 @@ class BaseCollectionSearchForm(forms.Form):
...
+        Args:
+
+        prefix_field_with_type:
+        example when field is 'opening_start_day', prefix_field is 'opening', prefix_field_with_type is 'opening_start'
+        """
+        return (
+            self.cleaned_data.get(prefix_field_with_type + "_day"),
+            self.cleaned_data.get(prefix_field_with_type + "_month"),
+            self.cleaned_data.get(prefix_field_with_type + "_year"),
+        )
```
ababic, line 273: This feels a little bit like code smell, as there's no way to enforce this naming convention for fields. Is it not possible to have this wrapped up in a single form field, where this behaviour can be controlled (as in the examples provided). Surely we could just copy this code exactly and tweak the validation logic to meet our needs?:
Field: https://github.com/wildfish/crispy-forms-gds/blob/master/src/crispy_forms_gds/fields.py#L11-L170
Widget: https://github.com/wildfish/crispy-forms-gds/blob/master/src/crispy_forms_gds/widgets.py#L5-L59

PR #387, opened 2022-07-04
Upgrade black

TNA-Allan:
Upgrade black


ababic (approved):
Nice and simple! Thanks @TNA-Allan.

PR #388, opened 2022-07-05
DF-209: Split datefield (componentised)

ababic:
Supersedes #385


etna/core/tests/test_fields.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_end_date_default_day31_month12_for_empty_day_empty_month(self):
+        input_day = ""
+        input_month = ""
+        input_year = "2001"
+        date = datetime.date(year=int(input_year), month=12, day=31)
+        field = DateInputField(
+            required=False, default_day=END_OF_MONTH, default_month=12
+        )
+        value = field.clean([input_day, input_month, input_year])
+        self.assertEqual(value, date)
```
ababic, line 85: Suggestions:
- Avoid unnecessary variabalisation: If a value is only used once, variabalising it makes the code more difficult to scan
- Try to limit tests to 'one thing' per test, or use docstrings to explain what is being tested where that is not possible
- Where there is a pattern to what you are testing, loop over variations of inputs / expected outputs and use `subTest`... there's no better way to indicate "this same thing is being tested each time"

etna/search/tests/test_forms.py
```diff
@@ -4,12 +4,89 @@
...
-    def test_start_date_after_end_date_is_invalid(self):
+    def test_empty_opening_start_date_is_valid(self):
         form = CatalogueSearchForm(
             {
                 "group": "tna",
-                "opening_start_date": "2000-01-01",
-                "opening_end_date": "1900-01-01",
+                "opening_end_date_0": "01",
+                "opening_end_date_1": "12",
+                "opening_end_date_2": "2002",
```
ababic, line 13:
```suggestion
                "opening_start_date_0": "01",
                "opening_start_date_1": "12",
                "opening_start_date_2": "2002",
```

etna/search/tests/test_forms.py
```diff
@@ -4,12 +4,89 @@
...
+
+        self.assertTrue(is_valid)
+
+    def test_empty_opening_end_date_is_valid(self):
+        form = CatalogueSearchForm(
+            {
+                "group": "tna",
+                "opening_start_date_0": "01",
+                "opening_start_date_1": "01",
+                "opening_start_date_2": "2001",
```
ababic, line 27:
```suggestion
                "opening_end_date_0": "01",
                "opening_end_date_1": "01",
                "opening_end_date_2": "2001",
```

etna/core/tests/test_fields.py
TNA-Allan: Thanks @ababic updated.

etna/search/tests/test_forms.py
TNA-Allan: The case was correct, but Not relevant, code changed to subTest.

etna/search/tests/test_forms.py
```diff
@@ -4,12 +4,89 @@
...
+                "opening_start_date_1": "12",
+                "opening_start_date_2": "1999",
+                "opening_end_date_1": "01",
+                "opening_end_date_2": "2000",
+            }
+        )
+
+        is_valid = form.is_valid()
+
+        self.assertTrue(is_valid)
```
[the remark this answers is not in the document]
TNA-Allan: @ababic , Thanks for this

> These tests are overkill.
Had to retain some due to its nature of using both fields during validation. For example in `test_various_opening_date_inputs_is_valid`
Others moved to `test_fields.py` and added `test_validators.py`

etna/core/tests/test_fields.py
```diff
@@ -0,0 +1,132 @@
...
+                "Entered date must include a Month.",
+            ),
+            (
+                "input year is empty ddmm<empty>",
+                {"input_day": "01", "input_month": "01", "input_year": ""},
+                "Entered date must include a Year.",
+            ),
+        ):
+            with self.subTest(label):
+                field = DateInputField(
```
ababic, line 100: Field could be defined outside of the loop, but it's fine here

etna/core/tests/test_fields.py
```diff
@@ -0,0 +1,132 @@
...
+                "Entered date must be a real date, for example 23 9 2017.",
+            ),
+            (
+                "input day is empty <empty>mmyyyy",
+                {"input_day": "", "input_month": "01", "input_year": "2007"},
+                "Entered date must be a real date, for example 23 9 2017.",
+            ),
+        ):
+            with self.subTest(label):
+                field = DateInputField(required=False)
```
ababic, line 128: Field could be defined outside of the loop, but it's fine here

etna/core/tests/test_validators.py
```diff
@@ -0,0 +1,62 @@
+from django.core.exceptions import ValidationError
+from django.test import SimpleTestCase
+
+from etna.core.fields import DateInputField
+
+
+class DateFieldValidatorTest(SimpleTestCase):
```
ababic, line 7: From the name of the module, I would expect this file to contain unit tests for the validators themselves, but we are testing the field again here. I would just move this test case to `test_fields.py` and remove this module.

etna/search/tests/test_forms.py
```diff
@@ -4,31 +4,79 @@
...
 
 class CatalogueSearchFormTest(SimpleTestCase):
-    def test_start_date_after_end_date_is_invalid(self):
+    def test_various_opening_date_inputs_is_valid(self):
+        """tests inputs accross both date fields with various input formats for form validity"""
+        for label, value in (
+            (
+                "input end date with empty start date",
+                {
+                    "input": {
```
ababic, line 13: Not sure what we're gaining from the multi-level dicts here. If all we need is data for the form, we should lose the outer dicts (with the `"input"` key)

etna/search/tests/test_forms.py
```diff
@@ -4,31 +4,79 @@
 
 
 class CatalogueSearchFormTest(SimpleTestCase):
-    def test_start_date_after_end_date_is_invalid(self):
+    def test_various_opening_date_inputs_is_valid(self):
+        """tests inputs accross both date fields with various input formats for form validity"""
+        for label, value in (
```
ababic, line 9:
```suggestion
        for label, form_data in (
```

etna/search/tests/test_forms.py
```diff
@@ -4,31 +4,79 @@
...
+                        "opening_start_date_2": "1999",
+                        "opening_end_date_0": "31",
+                        "opening_end_date_1": "12",
+                        "opening_end_date_2": "1999",
+                    }
+                },
+            ),
+        ):
+            with self.subTest(label):
+                form = CatalogueSearchForm(value["input"])
```
ababic, line 62:
```suggestion
                form = CatalogueSearchForm(form_data)
```

etna/core/fields.py
```diff
@@ -0,0 +1,236 @@
...
+                required=default_day is None,
+            ),
+            forms.CharField(
+                label=_("Month"),
+                error_messages={"incomplete": "Entered date must include a Month."},
+                validators=[
+                    PositiveIntegerStringValidator(
+                        min=1,
+                        max=12,
+                        msg="Entered date must be a real date, for example 23 9 2017.",
```
ababic, line 67: These are component-specific errors, so making them all generic makes the field less helpful. What was wrong with the original wording?

ababic: There's definitely no reason the 'Day', 'Month' or 'Year' to be capitalised in any of these messages

etna/core/fields.py
```diff
@@ -0,0 +1,236 @@
...
+                field_value = None
+            if field_value in self.empty_values:
+                if self.required and self.require_all_fields:
+                    # Raise a 'required' error if the MultiValueField is
+                    # required and all sub-fields are required.
+                    raise ValidationError(
+                        self.error_messages["required"], code="required"
+                    )
+
+                if (
```
ababic, line 143: This new condition prevents a default month value from being respected below

etna/core/fields.py
```diff
@@ -0,0 +1,236 @@
...
+                    i == 1
+                    and self.default_values[i]
+                    and value[0]
+                    and value[2]
+                    and not value[1]
+                ):
+                    # Validate empty month input (overrid default month) when day, year are input
+                    msg = "Entered date must include a Month."
```
ababic, line 151: Why are we giving month special treatment here? If no default is specified for month, `field.required` will be `True`, and a widget-specific incomplete error will be raised below.

etna/core/fields.py
```diff
@@ -153,14 +187,24 @@ def clean(self, value):
...
+            raise ValidationError(e)
 
         try:
             out = self.compress(clean_data)
         except ValueError:
-            # The specified day must be invalid for the month, so add an error message
-            # to the 'day' sub-field widget before raising
-            msg = "Enter a valid day for the month you entered"
+            # When Day, Month belong to correct range and Year is 0, then error-value "year 0 is out of range"
+            msg = "Entered date must be a real date, for example 23 9 2017."
```
ababic, line 209: What's the benefit of making this generic? Surely that just makes it difficult to see what is wrong?

etna/core/tests/test_fields.py
TNA-Allan: Done - its outside the loop.

etna/core/tests/test_fields.py
TNA-Allan: Done - its outside the loop.

etna/core/tests/test_validators.py
TNA-Allan: As mentioned in the doc string, it validates the overridden default messages. I've split the test case for each field validator.

etna/search/tests/test_forms.py
TNA-Allan: Done - removed input key

etna/search/tests/test_forms.py
TNA-Allan: Same as above Done - removed input key

etna/search/tests/test_forms.py
TNA-Allan: Same as above Done - removed input key

etna/core/fields.py
TNA-Allan: This the requirement.

etna/core/fields.py
TNA-Allan: It does respect default month  see 
 `etna.core.tests.test_fields.DateFieldTest.test_various_inputs_with_dtf_default_for_end_date`

etna/core/fields.py
TNA-Allan: This is the requirement for those error handling.

TNA-Allan (approved):
Looks good :)

PR #390, opened 2022-07-06
Highlight card changes

gtvj:
Changes include:

- Swapping the order of catalogue data for Wagtail supplied data
- Swapping record.title (which does not seem to return anything) for
  record.description and truncating it to different word lengths
  depending on whether it is before or after the image
- Using striptags to prevent the XML from appearing

### DIFF
--- a/templates/includes/card-group-record-summary-no-image.html
+++ b/templates/includes/card-group-record-summary-no-image.html
@@ -20,7 +20,7 @@
             </p>
         </a>
         <div class="card-group-record-summary__body">
-            <p>{{ record.description | striptags | escape }}</p>
+            <p>{{ record.description|striptags }}</p>
         </div>
     </div>
 </li>
--- a/templates/includes/card-group-record-summary-no-image.html
+++ b/templates/includes/card-group-record-summary-no-image.html
@@ -8,7 +8,7 @@
 
 <li class="col-sm-12 col-md-6 col-lg-4">
     <div class="card-group-record-summary">
-        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{% firstof description_override|escape record.title|striptags|escape %}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
+        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{% firstof description_override|safe record.title|striptags %}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
             <p class="card-group-record-summary__heading">
                 {% if record.reference_number %}
                 <span class="sr-only">
--- a/templates/includes/card-group-record-summary-no-image.html
+++ b/templates/includes/card-group-record-summary-no-image.html
@@ -8,19 +8,19 @@
 
 <li class="col-sm-12 col-md-6 col-lg-4">
     <div class="card-group-record-summary">
-        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{{ record.title }}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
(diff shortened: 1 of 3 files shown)


templates/includes/card-group-record-summary.html
```diff
@@ -10,7 +10,7 @@
                     {{ record.reference_number }} -
                 </span>
                 {% endif %}
-                {{ record.title }}
+                {{ description_override|default:record.description|safe }}
```
ababic, line 13: We shouldn't ever mark anything outside of our control as safe (e.g. user input / api content), so would be good to update this. Maybe `firstof` would work here?

```suggestion
                {% firstof description_override|safe record.description|striptags|escape %}
```

TNA-Allan:
@gtvj , the 'results_page.html' also has `card-group-record-summary-no-image.html'. Would you like to update it as well similar to `card-group-record-summary.html'

gtvj:
Thanks @ababic - I've introduced the changes you proposed.

templates/includes/card-group-record-summary-no-image.html
```diff
@@ -8,19 +8,19 @@
...
-        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{{ record.title }}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
+        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{% firstof description_override|safe record.description|striptags|linebreaks %}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
             <p class="card-group-record-summary__heading">
                 {% if record.reference_number %}
                 <span class="sr-only">
                     {{ record.reference_number }} -
                 </span>
                 {% endif %}
-                {{ record.title }}
+                {% firstof description_override|safe record.description|striptags|linebreaks %}
```
ababic, line 18: Sorry @gtvj, I hadn't quite realised that we're already in a `<p>` tag here, so we don't need `|linebreaks`.

Just want to check that the intention is definitely to use `record.description` here instead of `record.title`? descriptions can be pretty darn long in some cases... it seems strange to be using the full value as a 'sort of heading', but then always displaying a truncated version below.

gtvj: All sorted and reverted back to `record.title`. I made this change because I wasn't getting anything back for `record.title` and in most cases it's simply a truncated `record.description` anyway. Still, it's using `record.title` now (with `record.description` appearing below).

templates/includes/card-group-record-summary-no-image.html
```diff
@@ -8,19 +8,19 @@
 
 <li class="col-sm-12 col-md-6 col-lg-4">
     <div class="card-group-record-summary">
-        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{{ record.title }}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
+        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{% firstof description_override|safe record.title|striptags|linebreaks %}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
```
ababic, line 11: We need to treat the values a little differently when using them as a HTML element attribute:

```suggestion
        <a href="{% record_url record is_editorial=True %}" class="card-group-record-summary__link" data-card-type="card-group-record-summary-no-image" data-card-title="{% firstof description_override|escape record.title|striptags|escape %}"{% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %} target="_blank"{% endif %}>
```

templates/includes/card-group-record-summary-no-image.html
```diff
@@ -8,19 +8,19 @@
...
                 </span>
                 {% endif %}
-                {{ record.title }}
+                {% firstof description_override|safe record.title|striptags %}
                 {% if FEATURE_RECORD_LINKS_GO_TO_DISCOVERY %}<span class="sr-only">(link opens in a new window)</span>{% endif %}
             </p>
         </a>
         <div class="card-group-record-summary__body">
-            <p>{{ description_override|default:record.description|striptags|truncatewords:40 }}</p>
+            <p>{{ record.description | striptags | escape }}</p>
```
ababic, line 23: I'm unsure of the need for the `| escape` addition here. If we aren't marking the value as 'safe', it should be escaped automatically here.
```suggestion
            <p>{{ record.description|striptags }}</p>
```

gtvj:
Hi @TNA-Allan - would you be OK to review this PR, please? I've addressed the issue with the failing test.

PR #391, opened 2022-07-06
UN-213: Frontend tweaks

JamesWChan:
Hi @matt-blair,

Could you have a look at this PR? It primarily addresses the featured insight image issue we've been discussing (also, as the `object-fit` property is not supported in IE11, I've taken your suggestion of omitting the image and making the yellow textbox full width. Had a quick chat with Gwyn this morning about the approach and he was in agreement). It also updates the `featured-record` component with some more spacing and the new caption style when there is an image. I've tested in Chrome, Firefox, Safari, Edge, and IE11 and didn't spot any issues.

@gtvj - just tagging you because I was asked to make the featured insight image a link and you may want check the implementation. I've copied the card implementation for this and the image link will be ignored by screen readers (by applying `aria-hidden="true"`) and keyboard users (by removing it from the tab order using `tabindex="-1"`)

Thank you 👍

### DIFF
--- a/sass/includes/_featured-insight.scss
+++ b/sass/includes/_featured-insight.scss
@@ -2,11 +2,23 @@
     display: flex;
     margin-bottom: 4rem;
 
-    &__image {
+    &__image-link {
+        display: inline-block;
         width: 50%;
+        z-index: 999;
+
+        img {
+            height: 100%;
+            width: 100%;
+            object-fit: cover;
+        }
 
         figure {
             margin-bottom: 0;
+        }
+
+        &:hover {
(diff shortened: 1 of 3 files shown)


matt-blair (commented):
Hi @JamesWChan,

Just looked through this and can't see any immediate issues. 

Can I ask whether this approach worked with the focal point in Wagtail in the end? Just curious to know.

Matt.

JamesWChan:
Great, thank you @matt-blair. Yeah it works with the focal point; thanks for your input yesterday, I think it is a solid solution 👍

PR #393, opened 2022-07-07
Fix/insights fe fixes

JamesWChan:
Hi @matt-blair,

Could you have a look at this PR? Just some more FE changes that have been requested:

- Adding a fixed height and `overflow: scroll` to transcripts (otherwise they display the full length of the transcript which can increase the length of the page quite a bit)
- Inserting the transcript into the tab order because IE11 keyboard users can't scroll through the transcript if it isn't focusable
- Hiding the scrollbar on elements that have `overflow: scroll` on them (this was causing vertical and horizontal scrollbars to appear unnecessarily)
- Reducing the `z-index` of the media component (e.g. podcasts) so that it doesn't overlap with the cookie consent banner

Happy to hop on a call and go through them in more detail if you'd like. Thank you 👍

### DIFF
--- a/sass/includes/_jumplinks.scss
+++ b/sass/includes/_jumplinks.scss
@@ -4,6 +4,12 @@
   padding: 2.5rem 0 0 1rem;
   height: 100vh;
   overflow: scroll;
+  -ms-overflow-style: none;  /* Hide jumplinks scrollbar on Internet Explorer 10+ and Edge */
+  scrollbar-width: none;  /* Hide jumplinks scrollbar on Firefox */
+
+  &::-webkit-scrollbar { 
+    display: none;  /* Hide jumplinks scrollbar on Safari and Chrome */
+  }
 
   &__heading {
     font-size: 2rem;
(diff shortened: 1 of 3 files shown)


matt-blair (commented):
Thanks @JamesWChan. 

I've just spotted a potential issue with hiding the scrollbars in that there may not be any affordance to suggest the content inside is scrollable. This is the case for both the transcript and the left hand panel. It could possibly have an impact for those who aren't gesture-based scrolling or scrolling using a mouse wheel as they won't have an element to select and drag. You also don't get a sense of where you are or how big the container is if you don't have a scrollbar. 

A possible solution might be re-thinking the design of the list and the transcript text into a dedicated container so that the overflow only attaches to the list part of the left hand panel side rather than the whole left hand panel. This way you could design something a little more subtle to accommodate the scrollbar in browsers that render it showing by default. The other option would be to have a bit more of a think about how that left hand panel might work but this would obviously take a bit of trial and error.

Maybe a potential option would be to just use overflow-y with overflow-x hidden in the first instance to avoid introducing a potential usability issue from the start and iterate from there?

Happy to have a chat about it. Not an immediate one to easily solve, but certainly something to be aware of in case people encounter issues.

matt-blair (approved):

PR #395, opened 2022-07-10
DF-333:Update 'Collection' and 'Held by' filters to use 'OR' rather than the default 'AND' operator

TNA-Allan:
Make collection, heldby fields to use OR operator for multiple filter values

### DIFF
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -75,8 +75,10 @@ class Aggregation(str, enum.Enum):
 
 def prepare_filter_aggregations(items: Optional[list]) -> Optional[str]:
     """
-    Filter format in items: 'field:value'
+    Filter format in items: 'field:value', 'field:value:or'
     Prepares i.e. removes/replaces special chars from a filter fields' value to be passed to the api
+    When using filter with multiple values, specific fields require OR operator to be specified,
+    otherwise AND is used by default.
 
     Example:
     before-prepare: "heldBy:Birmingham: Archives, Heritage and Photography Service"
@@ -94,6 +96,7 @@ def prepare_filter_aggregations(items: Optional[list]) -> Optional[str]:
     subst = " "
     field_list_to_prepare = ["heldBy"]
     filter_prepared_list = []
+    fields_using_or_operator = ["heldBy", "collection"]
 
     for item in items:
         field, value = item.split(":", 1)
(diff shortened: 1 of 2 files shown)


ababic:
Sorry @TNA-Allan, but I'm not sure that hardcoding the treatment of specific filter aggregations is the way to go (it's quite possible that other views could be introduced in future that want to apply multiple AND filters for the same field). I think it should be up to the Django view to add ":or" to each of the form values instead.

ababic:
@TNA-Allan I would suggest [BaseFilteredSearchView.get_api_filter_aggregations()](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/search/views.py#L403-L425) as the place for this... perhaps using an `or_filters = []` attribute on the class to control it.

TNA-Allan:
> Sorry @TNA-Allan, but I'm not sure that hardcoding the treatment of specific filter aggregations into the client is the way to go (it's quite possible that other views could be introduced in future that want to apply multiple AND filters for the same field). I think it should be up to the Django view to add ":or" to each of the form values instead.

I'd like to understand what is hardcoding the treatment means
since the code is placed in `def prepare_filter_aggregations` where these fields are treated specifically
using configured `fields_using_or_operator = ["heldBy", "collection"]`
the `":or"` is appended at one place.

Besides this as the method name states, it does `prepare` the filter. Also it takes care of `:` which replaces values containing spaces and not in the `:or` appended.

Can you please give an example of applying multiple AND filters for the same field?

ababic:
@TNA-Allan my advice would be to try and step back from the problem a little and see each component for what it is...

The clients job is to make an API request based on the options passed to it's methods. It's a 'convenience' interface t

TNA-Allan:
> Now, if I were to ask you "Allan, the search view appears to be doing something funny when I use this combination of filters. Can you look into it?", where would you look first?

I'm thinking of fat models, skinny views, although this is not the same  :)

> However unlikely it is that the user intends to do an multiple `AND` search for these particular fields is redundant. It should be apparent from the view code (the first place you should look when someone reports a problem) how the values are being treated and passed to the API client.

Using of AND in this case is meaningless as the data returned with AND is mutually exclusive, therefore 0 records returned when AND applied.

`def prepare_filter_aggregations` is in client will require further treatment due to the `:or` appearing at the end. Should this method qualify to be moved over to the view and processed over here [BaseFilteredSearchView.get_api_filter_aggregations()](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/search/views.py#L403-L425)

etna/ciim/client.py
```diff
@@ -107,8 +110,24 @@ def prepare_filter_aggregations(items: Optional[list]) -> Optional[str]:
...
+        if sum((item.split(":", 1)[0].count(field) for item in items)) > 1:
+            # append or to the value for each field
+            updated_list_for_or_operator = []
+            for item in filter_prepared_list:
+                if item.split(":", 1)[0] == field:
+                    updated_list_for_or_operator.append(item + ":or")
+                else:
+                    updated_list_for_or_operator.append(item)
+
+            filter_prepared_list = updated_list_for_or_operator
```
ababic, line 129: Is it not possible to do this in the above loop? And can we not just always add `:or` whether multiple values have been supplied or not, or does that cause an error?

TNA-Allan:
> Where there are _horrible API-specific implementation details_, 

_Search your feelings Luke_  :) 

> Perhaps we can add a `or_filter_fields` option or something to `KongClient.search()`, and that can pass both values on to the `prepare_filter_aggregations()`?

I understood previously -
> What I mean is: this behaviour doesn't belong on the client.

I thought lets keep it simple and leave the `prepare_filter_aggregations()` do what its doing.

etna/ciim/client.py
TNA-Allan: Don't want to "mash" adding `:or` whilst preparing the filter to remove spaces in the above loop. But as a role of preparing the filter, added after remove spaces, to keep it simple and target only those configured fields in var  `fields_using_or_operator`

ababic (approved):

PR #409, opened 2022-07-13
Refactor cookie consent JS implementation

JamesWChan:
Hi @gtvj,

I've hardcoded the correct `gtm-script` into the cookie consent JS for now (loaded it as `src="/static/scripts/gtm_script.js"`). I'm working on moving all the cookie consent modules into `ds-wagtail` and building them there but I'm having some issues. So I've pushed this as a quick fix (works on localhost for me) in case I run out of time. Thank you 👍  

Edit: have now refactored the cookie consent JS implementation so that all the code for cookie consent exists as modules within this repository and is built by webpack. This is to make cookie consent easier to manage for this application.


JamesWChan:
Refactored the cookie consent JS implementation as per GJ's suggestion

PR #422, opened 2022-07-20
DF-501: Consistantly use `Record` objects to represent records

ababic:
The long-awaited sequel to #345 

The `APIManager` is gone... There is now only `KongClient`, and its methods return more useful values... with lazy conversion to boot.

With `Record` instances used consistently to represent records, we can remove a bunch of horrible code for dealing with different types of values in different places, and have a single component that is responsible for data conversion.

Please thank the following bits of code for their service: 
1. The (necessarily) over-complicated `{% record_url %}` tag
2. Those special custom template filters in `search_tags.py` for extracting record details out of dicts
3. `APIManager` and all of the tests for it
4. Unnecessary methods on `APIModel` to create `APIManager` instances


TNA-Allan (commented):
Hi @ababic, some [CI steps](https://github.com/nationalarchives/ds-wagtail/actions/runs/4143278942/jobs/7164951815) are failing.

TNA-Allan (commented):
[Ensure fake API responses have the same ['hits']['total']['value'] in both the 'agreggations' and 'results' responses, as the latter is now used by ResultList](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/b4a3af37cde9ae9ebff7bca19ab20aca1f0e23e6)
I think we should not fake the original response, but rely on original values. Better if we get the values at real time at some point.

etna/records/models.py
```diff
@@ -118,24 +99,13 @@ def reference_number(self) -> str:
...
-        raw data for this record. Otherwise `False`.
-        """
-        try:
-            self.reference_number
-        except ValueExtractionError:
-            return False
-        else:
-            return True
+    def reference_prefixed_title(self):
+        return f"{self.reference_number or 'N/A'} - {self.summary_title}"
```
TNA-Allan, line 105: How about calling it  `reference_prefixed_summary_title` - due to return containing `self.summary_title`

etna/search/templatetags/search_tags.py
```diff
@@ -7,98 +7,12 @@
...
-            value = record["_source"]["@template"]["details"][key]
-        except KeyError:
-            return ""
-
-    if not value:
-        return ""
-    if isinstance(value, (list, tuple)):
-        value = value[0]
-
-    return mark_safe(bleach.clean(value, tags=["mark"], strip=True))
```
TNA-Allan, line 49: This code is removed, but did not find it in [`def highlights`](https://github.com/nationalarchives/ds-wagtail/blob/19df21c949fde726b210e9da49f732a4442fea38/etna/records/models.py#L66) Search results will show without cleaning. Are we raising it to be cleaned in the API response with `mark` retained?

templates/search/blocks/search_results__list-card--grid.html
```diff
@@ -10,13 +9,12 @@
 {% endif %}
 
 {% comment %} TODO downloadable, and record variables {% endcomment %}
-{% load search_tags records_tags static %}
 
 <li class="search-results__list-card--grid">
-    {% if record|record_detail:"digitised" %}
+    {% if record.is_digitised %}
```
TNA-Allan, line 14: The record|record_detail:"digitised" [implementation is removed.](https://github.com/nationalarchives/ds-wagtail/pull/422/files#diff-04206e540f8f7fda4cf059782e16ee16e502465b2ac29d463c00238a146b504dL19-L29)
We could add it to [current model implementation](https://github.com/nationalarchives/ds-wagtail/blob/19df21c949fde726b210e9da49f732a4442fea38/etna/records/models.py#L206) to so that it shows up in search results.

templates/search/blocks/featured-search__results-block.html
```diff
@@ -7,10 +7,12 @@ <h2 class="featured-search__heading">{{ bucket.label }}</h2>
             <li class="featured-search__results-list-item">
                 <a href="{% record_url record %}" class="featured-search__results-link">
                     <h3 class="featured-search__results-heading">
-                        {{ record|record_detail:"summaryTitle"|striptags }}
+                        {{ record.summary_title }}
                     </h3>
                 </a>
-                <p>Catalogue reference: {{ record|record_detail:"referenceNumber"}}</p>
+                {% if record.reference_number %}
```
TNA-Allan, line 13: New `if` condition is added.  Shouldn't `Catalogue reference` show irrespective of the availability of `reference_number`?

etna/records/models.py
```diff
@@ -242,6 +240,18 @@ def description(self) -> str:
             return format_description_markup(raw)
         return ""
 
+    @property
+    def raw_content(self) -> str:
+        for item in self.get("source.content.items", ()):
+            return item.get("value", "")
+        return ""
```
TNA-Allan, line 247: The response for content is `"source": {"content": ["string"],...`. [Previous implementation](https://github.com/nationalarchives/ds-wagtail/blob/36f31a94d555fba18c8275572d9224e4c340998b/etna/search/templatetags/search_tags.py#L84) considered `detail[0]`

etna/records/models.py
```diff
@@ -242,6 +240,18 @@ def description(self) -> str:
...
+
+    @cached_property
+    def content(self) -> str:
+        if raw := self.raw_content:
+            return format_description_markup(raw)
```
TNA-Allan, line 252: This implementation caters for links in response. Its not same as [previous implementation](https://github.com/nationalarchives/ds-wagtail/blob/36f31a94d555fba18c8275572d9224e4c340998b/templates/search/blocks/featured-search__interpretive-card.html#L24) `{{ record|interpretive_detail:"content"|safe|striptags|truncatewords:40 }}` 
Are there links in the response?

etna/search/views.py
```diff
@@ -301,9 +299,14 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
 
         custom_dimension9 = self.form.cleaned_data.get("q") or "*"
 
+        result_count = self.get_result_count()
+        if result_count > 10000:
+            result_count = 10001
```
TNA-Allan, line 304:
```suggestion
```
issue: As per issue related to `customMetric1` with consistency for this PR its agreed to show `customMetric1=10000`  for values `gte 10000` , otherwise actual value. Also mentioned here [DF-334](https://national-archives.atlassian.net/browse/DF-334)

etna/search/views.py
```diff
@@ -793,20 +751,21 @@ def get_buckets(self) -> Dict[str, Bucket]:
...
+
+    def get_result_count(self):
+        """
+        Overrides BaseSearchView.get_result_count() to return the combined
+        totals from all buckets.
+        """
+        total = 0
+        for bucket in self.get_buckets().values():
+            total += bucket.result_count
+        return total
```
TNA-Allan, line 771:
```suggestion
        for bucket in self.get_buckets().values():
            total += bucket.result_count
        # custom value for customMetric1
        return 10000 if total >= 10000 else total
```
Over here, the actual counts are displayed, but need to set `customMetric1` as per agreed changes mentioned in [DF-334](https://national-archives.atlassian.net/browse/DF-334)

etna/search/tests/test_views.py
```diff
@@ -606,348 +595,587 @@ def test_current_bucket(self):
...
+                "customDimension9": "*",
+                "customDimension10": "",
+                "customDimension11": "",
+                "customDimension12": "",
+                "customDimension13": "",
+                "customDimension14": "",
+                "customDimension15": "",
+                "customDimension16": "",
+                "customDimension17": "",
+                "customMetric1": 10001,
```
TNA-Allan, line 698:
```suggestion
                "customMetric1": 10000,
```
to resolve and align with issue of consistency for `customMetric1` in this PR.

etna/search/tests/test_views.py
```diff
@@ -606,348 +595,587 @@ def test_current_bucket(self):
...
```
TNA-Allan, line 758:
```suggestion
                "customMetric1": 10000,
```
to resolve and align with issue of consistency for `customMetric1` in this PR.

etna/search/tests/test_views.py
```diff
@@ -606,348 +595,587 @@ def test_current_bucket(self):
...
```
TNA-Allan, line 848:
```suggestion
                "customMetric1": 10000,
```
to resolve and align with issue of consistency for `customMetric1` in this PR.

etna/search/tests/test_views.py
```diff
@@ -606,348 +595,587 @@ def test_current_bucket(self):
...
```
TNA-Allan, line 878:
```suggestion
                "customMetric1": 10000,
```
to resolve and align with issue of consistency for `customMetric1` in this PR.

etna/records/models.py
TNA-Allan, line 105:
```suggestion
    def reference_prefixed_summary_title(self):
        return f"{self.reference_number or 'N/A'} - {self.summary_title}"
```

TNA-Allan (changes requested):
Hi @ababic - some changes required. In order to use value for `customMetric1` consistently it was discussed in  [DF-334](https://national-archives.atlassian.net/browse/DF-334)

etna/search/templatetags/search_tags.py
ababic: I will investigate further into whether further work is needed here.

In regards to cleaning of values: I feel we really need to the API to decide on types of values it will serve, enforce that, and go from there. I don't see any good reason why any of the fields we are using would contain HTML (they just 'do'  - and we're having to 'hack around it')

templates/search/blocks/search_results__list-card--grid.html
ababic: @TNA-Allan According to the spec, `digitised` should be present as a boolean at the top level, which is where the `is_digitised` looks. I would expect that's true in a 'search results' context too.

ababic: Have played around, and it looks like there is no 'digitised' value available ANYWHERE in result content when specifying `template=details` (like we appear to be doing [here](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/search/views.py#L200) and [here](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/search/views.py#L401)), so I guess this may have never worked!

TNA-Allan: Check with this one on develop  /search/catalogue/?group=digitised

ababic: Ah, so what the template includes is also dependant on the filters applied to the search? That's 'fun' :D

ababic: Resolved in c1db32a :)

templates/search/blocks/featured-search__results-block.html
ababic: Resolved in 9672286

etna/records/models.py
ababic: Good shout. I've reviewed the spec, found some examples, and adjusted this property accordingly in f1f640e.

etna/records/models.py
ababic: Thanks Allan. This was an oversight on my part. I have swapped the `format_description_mark()` treatment here for a bleach-based `strip_tags()` function, which should be safer than the django tag (as outlined here: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#striptags).

Fixed in d6720a6

etna/search/templatetags/search_tags.py
ababic: The one interpretive field that was being used in output has now been updated to strip any HTML in a similar way (see d6720a6)

etna/records/models.py
ababic: Applied in 165b072

templates/search/blocks/featured-search__results-block.html
```diff
@@ -7,10 +7,10 @@ <h2 class="featured-search__heading">{{ bucket.label }}</h2>
             <li class="featured-search__results-list-item">
                 <a href="{% record_url record %}" class="featured-search__results-link">
                     <h3 class="featured-search__results-heading">
-                        {{ record|record_detail:"summaryTitle"|striptags }}
+                        {{ record.summary_title }}
```
TNA-Allan, line 10: 1) Existing implementation over here has `striptags`
2) noticed some inconsistencies in existing usage of summary title in templates with striptags, Ex: using [striptags](https://github.com/nationalarchives/ds-wagtail/blob/39bf313563744d381acd4c3c8002e20ab42df750/templates/includes/card-group-record-summary-no-image.html#L11) ; without [`striptags` ](https://github.com/nationalarchives/ds-wagtail/blob/39bf313563744d381acd4c3c8002e20ab42df750/templates/includes/records/record-title.html#L6), there should be something in place to get the data cleaned from the api response. Until then could the newly introduced `strip_html` method in this ticket be used to strip out tags from `summary_title` attribute?

etna/records/models.py
```diff
@@ -206,7 +205,9 @@ def legal_status(self) -> str:
 
     @cached_property
     def is_digitised(self) -> bool:
-        return self.get("digitised", default=False)
+        return self.get(
+            "digitised", default=self.get("template.digitised", default=False)
+        )
```
TNA-Allan, line 210:
```suggestion
        return self.get(
            "digitised", default=self.get("@template.details.digitised", default=False)
        )
```

TNA-Allan, line 210:
```suggestion
        return self.get(
            "digitised", default=self.get("@template.details.digitised", default=False)
        )
```

templates/search/blocks/featured-search__results-block.html
ababic: The API needs to be doing the stripping during ingestion really. But, using `strip_html` should work in the meantime

etna/search/tests/test_views.py
ababic: Resolved in a0ccd87

etna/search/tests/test_views.py
ababic: Resolved in a0ccd87

etna/records/models.py
ababic: Resolved in 70dad48

ababic: Resolved in 70dad48

templates/search/blocks/featured-search__results-block.html
ababic: As of 9a88688, `strip_html` is always applied to summary_title values, so we no longer need to use `|striptags` in templates

etna/search/tests/test_views.py
ababic: Resolved in [a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9)

etna/search/tests/test_views.py
ababic: Resolved in [a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9)

etna/search/views.py
ababic: Resolved in [a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9)

ababic:
@TNA-Allan I **think** everything you raised has now been resolved. Would you mind having another glance over to make sure?

TNA-Allan:
> @TNA-Allan I **think** everything you raised has now been resolved. Would you mind having another glance over to make sure?

@ababic 
Just took a **glance**
1) Could no find Resolved in [a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9)
2) The fake API responses are still showing.
Did you push these changes?

ababic:
Hi @TNA-Allan. Sorry for the confusion:

[a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9) was replaced by [0533fed](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/0533fed61d85eb3f9baff8a21f113dfb083c7680)

> The fake API responses are still showing.

I'm not following what you mean here, sorry. Could you clarify?

ababic:
@TNA-Allan If you're referring to your comment from a while ago:
> I think we should not fake the original response, but rely on original values. Better if we get the values at real time at some point.

I don't think we ever agreed this should be done as part of these changes. With respect, I think calling the API directly would be a huge mistake, as we'd be introducing some major flakiness into our test suite.

etna/search/templatetags/search_tags.py
TNA-Allan: Please see example `/search/catalogue/?filter_keyword=historic&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=&q=royal&per_page=20&sort_order=asc&display=list&group=tna`

In the above tag implementation data extracted here for highlight with <mark> is outside of `_source` i.e. `highlight@template.details.<key>` , therefore not getting <mark> in the output. For the same reason, I think the data extracted for `description` is not getting cleaned.

TNA-Allan:
> Hi @TNA-Allan. Sorry for the confusion:
> 
> [a0ccd87](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/a0ccd87f3aca7f9f87788eaccc56a67f9f38fdb9) was replaced by [0533fed](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/0533fed61d85eb3f9baff8a21f113dfb083c7680)
> 
> > The fake API responses are still showing.
> 
> I'm not following what you mean here, sorry. Could you clarify?

Hi @ababic , 
This is related to my previous comment on your commit re-
> [Ensure fake API responses have the same ['hits']['total']['value'] in both the 'agreggations' and 'results' responses, as the latter is now used by ResultList](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/b4a3af37cde9ae9ebff7bca19ab20aca1f0e23e6)
>  I think we should not fake the original response, but rely on original values. 

Those json files  are original responses from the API, we already have a factory method to create fake responses `factories->create_record`. Therefore without any changes made to these json files, the tests should work.

etna/search/tests/test_views.py
```diff
@@ -614,348 +603,587 @@ def test_current_bucket(self):
...
+                "customDimension9": "*",
+                "customDimension10": "",
+                "customDimension11": "",
+                "customDimension12": "",
+                "customDimension13": "",
+                "customDimension14": "",
+                "customDimension15": "",
+                "customDimension16": "",
+                "customDimension17": "",
+                "customMetric1": 10,
```
TNA-Allan, line 916:
```suggestion
                "customMetric1": 10000,
```

TNA-Allan:
Observed a [warning in CI test ](https://github.com/nationalarchives/ds-wagtail/actions/runs/4324410086/jobs/7549208390)

`/home/runner/work/ds-wagtail/ds-wagtail/etna/records/tests/test_blocks.py:7: RemovedInWagtail50Warning: Importing from wagtail.tests.utils.form_data is deprecated. Use wagtail.test.utils.form_data instead. See https://docs.wagtail.org/en/stable/releases/3.0.html#changes-to-module-paths
  from wagtail.tests.utils.form_data import nested_form_data, rich_text, streamfield`
  
It was not there previously for this PR.

etna/records/views/records.py
```diff
@@ -24,26 +24,26 @@ def record_disambiguation_view(request, reference_number):
...
-    if len(records) == 1:
-        record = records[0]
+    if len(result) == 1:
+        record = result.hits[0]
         return record_detail_view(request, record.iaid)
 
-    paginator = APIPaginator(count, per_page=per_page)
-    record_results_page = Page(records, number=page_number, paginator=paginator)
+    paginator = APIPaginator(result.total_count, per_page=per_page)
+    page = Page(result, number=page_number, paginator=paginator)
```
TNA-Allan, line 41: @ababic PR https://github.com/nationalarchives/ds-wagtail/pull/607 suggests `record_results_page`. Other Ok'd names :)

etna/ciim/client.py
```diff
@@ -134,6 +152,64 @@ def prepare_filter_aggregations(items: Optional[list]) -> Optional[str]:
     return filter_prepared_list
 
 
+class ResultList:
+    """
+    Wrapper class for JSON-decoded list of search results, that lazily converts
```
TNA-Allan, line 157: re : list of _search results_ 
Its covering results from various api's for `fetch` and `search`
Suggestion: Wrapper class for JSON-decoded list of _various api search, fetch results_, that lazily converts

etna/ciim/client.py
```diff
@@ -294,7 +397,11 @@ def search(
                 created_end_date, supplementary_time=time.max
             )
 
-        return self.make_request(f"{self.base_url}/data/search", params=params).json()
+        resp = self.make_request(f"{self.base_url}/data/search", params=params).json()
+        filter_aggregations_response, results_response = resp["responses"]
+        return self.resultlist_from_response(
+            results_response, filter_aggregations_response.get("aggregations")
+        )
```
TNA-Allan, line 404:
```suggestion
        resp = self.make_request(f"{self.base_url}/data/search", params=params).json()
        _, results_response = resp["responses"]
        return self.resultlist_from_response(
            results_response, results_response.get("aggregations")
        )
```

TNA-Allan: @ababic , there is an issue when passing `filter_aggregations_response.get("aggregations")` as the checkbox values in the `Refine by results` shore more than desired, response  depends on `results_response.get("aggregations")` the second part of the api response.
Ex url: /search/catalogue/?filter_keyword=ufo&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=&q=royal&per_page=20&sort_order=asc&display=list&group=tna

Obviously the impact of using `ResultList.filter_aggregations` may change. You might want to revisit this part.

etna/ciim/client.py
```diff
@@ -134,6 +152,64 @@ def prepare_filter_aggregations(items: Optional[list]) -> Optional[str]:
...
+    `total_count`:
+        The total number of results available
+
+    `aggregations`:
+        A dict of "aggregation" values from the results themselves
+
+    `filter_aggregations`:
+        A dict of "aggregation" values populated from the separate 'filter aggregations'
+        response from the API. For endpoints that do not include such data in their
+        response (e.g. 'searchAll/' and 'searchUnified/'), the value will be `None`.
```
TNA-Allan, line 171: @ababic There is a separate handling and meaning of filter_aggregations - its as search parameter `filterAggregations` . It looks **confusing** as the value assigned appears to be from a api response. Please look at this along with the comment of using `filter_aggregations_response` in `KongClient.search`. I think its not required, but other parts may be impacted.

ababic:
Hi @TNA-Allan, In answer to your points:

> Those json files are original responses from the API, we already have a factory method to create fake responses factories->create_record. Therefore without any changes made to these json files, the tests should work.

Okay, I think I understand now. While I'm with you that we could probably refactor our tests to work without these fixtures, that is a pretty big undertaking, and one I'd like to avoid in this PR. Happy for that to be a separate ticket in Jira (and work on it).

For now, I will undo all changes to any fixtures, and address any fallout from test failures.

etna/ciim/client.py
ababic: The `filter_aggregations` attribute has now been removed, and the `bucket_counts` attribute added in its place.

For the 'search' endpoint, this value comes from `["aggregations"]["groups"]["buckets"]` in first ES response, but in all other cases, it is taken from the same response that contains `["hits"]`

etna/search/templatetags/search_tags.py
ababic: Just investigating this issue now

ababic: So, it seems there are a few different issues here:
1. Highlight data is being currently being looked for in the wrong place (`raw_data["_source"]["highlight"]`). This can be fixed by updating `Record.__init__()` to look in the correct place in `raw_data` (`raw_data["highlight"]`), and dropping the `Record.highlights` cached_property.
2 `Record.description` is being used as plain text in templates, so really, the raw values should be passed through `strip_html()` 
3. The data in 'highlights' (when we actually pull it in correctly) includes partial tag HTML in it's list of 'marked strings', which make them impossible to clean up. Realistically, the only solution here is for K-Int to remove any HTML from these values on indexing, so that highlights come back as plain text with `mark` tags added (as expected)

ababic: The first issue above is resolved by 2581415 and 34a7ca6

ababic: The second is resolved by ce8389f

etna/ciim/client.py
ababic: @TNA-Allan I believe this is now resolved. `ResultList.filter_aggregations` has been replaced by `ResultList.bucket_counts`, which is populated from `ResultList.aggregations["groups"]["buckets"]` for all cases except for `search()`, which instead takes `["aggregations"]["groups"]["buckets"]` value from the first/separate ES response. This reflects the old behaviour from [BucketsMixin.get_group_buckets()](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/search/views.py#L66-L77), which can now be removed.

TNA-Allan: @ababic , this update looks good in terms of accessing the first/separate ES response, but the above url run does not show the filtered records (169) for `collection` and other attributes which uses the second ES response for `search`. This [filter_aggregations](https://github.com/nationalarchives/ds-wagtail/blob/bfb2e907bb81e4da5a308cba95dcab4c251a7623/etna/search/views.py#L455) attributes looks missed replacing with `aggregations`

ababic: Good catch @TNA-Allan! I thought I was going crazy there for a minute, as the failure was completely silent. I'll take that as a good lesson to be more cautious about wrapping things in `try/except` temporarily. Fixed in [984e97a](https://github.com/nationalarchives/ds-wagtail/pull/422/commits/984e97a6cd2c0785825a4a161ce3f03177a09ace)

TNA-Allan (approved):
Great work. Thanks @ababic

PR #426, opened 2022-07-22
Add IDs to all forms that didn't have one

gtvj:
This has been requested by the metrics team. Whereas I've ensured all
IDs are prefixed with `analytics-` I have not changed any existing IDs
since these might be included in existing reports etc.


ababic (approved):
LGTM :)

PR #439, opened 2022-08-15
Breadcrumbs feature added 

mozaki1:
Related ticket(s):
- <- Paste ticket URL here ->

## About these changes

<- Add human-readable description here ->

## How to check these changes

<- Where possible, provide guidance to help your reviewer ->

- [ ] Checked things thoroughly myself before handing over to you.
- [ ] Included the ticket number in the PR title to help us keep track of changes
- [ ] Ensured that my PR does not include any irrelevant commits.

### DIFF
--- a/etna/core/tests/test_middleware.py
+++ b/etna/core/tests/test_middleware.py
@@ -68,7 +68,7 @@ class TestInterpretCookiesMiddleware(SimpleTestCase):
         self._apply_middleware()
         context_data = self.response.context_data
         self.assertIs(context_data["cookies_permitted"], False)
-        self.assertIs(context_data["show_cookie_notice"], False)
+        self.assertIs(context_data["show_cookie_notice"], True)
         self.assertIs(context_data["show_beta_banner"], True)
 
     def test_behaviour_when_context_data_is_none(self):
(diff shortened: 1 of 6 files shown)


sdwilkes (changes requested):
Hi Mo, 
the current markup doesn't meet accessibility guidelines. The breadcrumbs should be inside a <nav> element and marked with 'aria-label=Breadcrumb'. The last (current) item also needs to be marked up with a 'aria-current=page' element.

You can more info here - https://www.w3.org/WAI/ARIA/apg/example-index/breadcrumb/index.html

If you want to see a TNA example, caselaw also follows this pattern - https://caselaw.nationalarchives.gov.uk/ewhc/qb/2022/2163

sdwilkes (changes requested):
Hi Mo, just a tiny change to the CSS (which I didn't spot before) - can we switch the text-decoration style around so that the ul.breadcrumb li a state is underlined and the hover state is set to none? 

This is the convention we use for other links. Then I can approve it. Thanks.

sdwilkes (approved):
Thanks Mo. That all looks good. We may have to do further work to make breadcrumbs work on mobile devices, but that can be done in a separate ticket. Thanks.

PR #442, opened 2022-08-17
Un 248

mozaki1:


templates/insights/insights_index_page.html
```diff
@@ -55,11 +55,11 @@ <h2>Discover all stories</h2>
         <ul class="card-group--list-style-none">
 
             {% for insights_page in insights_pages %}
-                      {% if insights_page not in featuredcollections %}
+
```
TNA-Allan, line 58: _featuredcollections_ is not part of insights page for which the latest items appearing twice are removed. 
There some code associated with it in models.py->get_context 
Is _featuredcollections_ required to be removed?

TNA-Allan (changes requested):
Please can we update the info in the comments PR template - Related ticket, about changes, etc.

templates/insights/insights_index_page.html
mozaki1: Yes , We do have duplications also on the insight landing page, however, based on the conversation I had with Lauren, she is Ok with it for now, but it needs addressing in the future, this could be via cashed property or inserted in context like now. So I left it as I know soon will be added

mozaki1: I can always remove it and add it in the future :)

TNA-Allan: >  she is Ok with it for now, but it needs addressing in the future,
Could we then address this when its required i.e. restore the if condition for _featuredcollections_.

Alternatively, update the ticket UN-248 to indicate required changes for _Insights Index Page_ . And keep a #TODO comment in get_context for retained code  _featuredcollections_

mozaki1: This is done now, I may implement it differently in the future, So all removed.

TNA-Allan (approved):
All Good

PR #444, opened 2022-08-19
upgrade to django 4.0

mozaki1:
Related ticket(s):
- <- Paste ticket URL here ->

## About these changes

<- Add human-readable description here ->

## How to check these changes

<- Where possible, provide guidance to help your reviewer ->

- [ ] Checked things thoroughly myself before handing over to you.
- [ ] Included the ticket number in the PR title to help us keep track of changes
- [ ] Ensured that my PR does not include any irrelevant commits.


etna/records/views/choosers.py
```diff
@@ -1,6 +1,6 @@
-from django.conf.urls import url
 from django.core.paginator import Page
 from django.shortcuts import Http404
+from django.urls import re_path as url
```
TNA-Allan, line 3: Since the url() function is deprecated and removed in 4.0, could we use _re_path_ instead of _as url_ as alias here does not look necessary.

mozaki1: Yes, Doing it this way will save changing the code elsewhere, the code uses URLs everywhere.

mozaki1: one single change is better than changing many files.

TNA-Allan: Yes, as discussed yesterday, we don't have URL's everywhere and it a change in this file itself.

package-lock.json
```diff
@@ -1,7179 +1,8 @@
 {
   "name": "ds-wagtail",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
TNA-Allan, line 4: Does `package-lock.json` require to be updated for the Django upgrade? It also appears to be downgraded.

mozaki1: These changes are all part of the poetry command, I am not sure I can answer this question. :)

TNA-Allan: Please can you share the poetry command for django upgrade?

mozaki1: I am not very sure , but I think it was poetry add  "django==4.0"

PR #449, opened 2022-08-23
DF-333:Added level to use OR operator

TNA-Allan:
Related ticket(s):
- https://national-archives.atlassian.net/browse/DF-333

## About these changes

- Added level field to use OR operator when performing search for Level filter.
- Level filter is hidden for “Records at other UK archives”

## How to check these changes
Under catalogue search select more than one level - example Item+Piece

- [x] Checked things thoroughly myself before handing over to you.
- [x] Included the ticket number in the PR title to help us keep track of changes
- [x] Ensured that my PR does not include any irrelevant commits.

### DIFF
--- a/templates/search/blocks/search_filters.html
+++ b/templates/search/blocks/search_filters.html
@@ -54,7 +54,9 @@
 
             {% include "search/includes/filter.html" with field=form.topic %}
 
-            {% include "search/includes/filter.html" with field=form.level %}
+            {% if form.group.value != 'nonTna' %}
+                {% include "search/includes/filter.html" with field=form.level %}
+            {% endif %}
 
             {% include "search/includes/filter.html" with field=form.closure %}
 
--- a/templates/search/blocks/search_filters.html
+++ b/templates/search/blocks/search_filters.html
@@ -54,7 +54,7 @@
 
             {% include "search/includes/filter.html" with field=form.topic %}
 
-            {% if form.group.value != 'nonTna' %}
+            {% if form.group.value != bucketkeys.NONTNA.value %}
                 {% include "search/includes/filter.html" with field=form.level %}
             {% endif %}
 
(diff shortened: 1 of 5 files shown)


templates/search/blocks/search_filters.html
```diff
@@ -54,7 +54,9 @@ <h4 class="search-filters__accordion-section-heading">Date</h4>
 
             {% include "search/includes/filter.html" with field=form.topic %}
 
-            {% include "search/includes/filter.html" with field=form.level %}
+            {% if form.group.value != 'nonTna' %}
```
mozaki1, line 57: Making decisions on hardcode data is tricky, so what happened if the data changes from nonTna to nontna , this code will no longer work,

TNA-Allan: Yes, its tricky. 

This is key data, ` "key": "nonTna",` in response and ideally that must not change. Its like a specification.
https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/ciim/constants.py#L60

if `nontna`, then the bucket for `Records for Other Archives` would not work as expected.

As such, all areas in search where this key is used would fail.

There are many such keys.- buckets - `tna, digitised,etc` 

I'm happy to do a lower comparison if otherwise.

mozaki1: The issue here is that will fail silently, and since it is a constant I suggest making it a constant in the relevant search view class, at least the change will be in one place. Would you be able to write a negative test, to handle invalid value, and then log the error?

TNA-Allan: > The issue here is that will fail silently
Isn't this relevant then `We did not find any results for your search' ?

> Would you be able to write a negative test, to handle invalid value, and then log the error?
Can it be more specific? Is this related to the comparison in the template? 
I am not sure of error logging for templates? Can you give me any pointers?

mozaki1: Yes, sorry , so if you move the 'nonTna' to the class and make it a constant ,  then you would be able to validate the value against what may come from the API or any source before returning the response, this way will know why it had failed by looking into log. in general hard-coded values must be constant ,also the advantage of making a constant then you will be able to change that constant value in one place . let's say the specification has changed for any reason then here you only change the constant once to the new value.

TNA-Allan: Can we please have a discussion?

mozaki1 (approved):

PR #532, opened 2022-11-10
DF-424:Title for search results

TNA-Allan:
Related ticket(s):
https://national-archives.atlassian.net/browse/DF-424

## About these changes
Return the search title for each search tab.

## How to check these changes

Navigate to each search tab and view the search title. 

- [x] Checked things thoroughly myself before handing over to reviewer.
- [x] Included the ticket number in the PR title to help us keep track of changes.

## For Reviewer

Once PR is merged, check and arrange to deploy on platform-sh.

### DIFF
--- a/etna/search/templatetags/search_tags.py
+++ b/etna/search/templatetags/search_tags.py
@@ -203,14 +203,14 @@ def hidden_fields_for_date_filter(selected_filters, form) -> str:
 
 
 @register.filter
-def search_title(searchtab) -> str:
+def search_title(search_tab) -> str:
     """
     Returns title for search tab
     """
-    if searchtab == SearchTabs.ALL.value:
+    if search_tab == SearchTabs.ALL.value:
         label = "All search results"
-    elif searchtab == SearchTabs.CATALOGUE.value:
+    elif search_tab == SearchTabs.CATALOGUE.value:
         label = "Catalogue search results"
-    elif searchtab == SearchTabs.WEBSITE.value:
+    elif search_tab == SearchTabs.WEBSITE.value:
         label = "Website search results"
     return label
--- a/etna/search/templatetags/search_tags.py
+++ b/etna/search/templatetags/search_tags.py
@@ -9,6 +9,8 @@ from django.utils.safestring import mark_safe
 
 import bleach
 
+from ...ciim.constants import SearchTabs
+
 register = template.Library()
 logger = logging.getLogger(__name__)
 
@@ -198,3 +200,17 @@ def hidden_fields_for_date_filter(selected_filters, form) -> str:
             html += include_hidden_fields(visible_field_names_str, form)
 
     return mark_safe(html)
+
+
+@register.filter
+def search_title(searchtab) -> str:
+    """
+    Returns title for search tab
+    """
+    if searchtab == SearchTabs.ALL.value:
+        label = "All search results"
+    elif searchtab == SearchTabs.CATALOGUE.value:
+        label = "Catalogue search results"
+    elif searchtab == SearchTabs.WEBSITE.value:
+        label = "Website search results"
+    return label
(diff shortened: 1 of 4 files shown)


TNA-Allan:
@mozaki1 / @jamesbiggs Please can you review this PR. Thanks.

jamesbiggs (approved):
Looks good Allan, cheers!

etna/search/templatetags/search_tags.py
```diff
@@ -198,3 +200,17 @@ def hidden_fields_for_date_filter(selected_filters, form) -> str:
...
+def search_title(search_tab) -> str:
+    """
+    Returns title for search tab
+    """
+    if search_tab == SearchTabs.ALL.value:
+        label = "All search results"
+    elif search_tab == SearchTabs.CATALOGUE.value:
+        label = "Catalogue search results"
+    elif search_tab == SearchTabs.WEBSITE.value:
+        label = "Website search results"
```
mozaki1, line 215: Sorry I am slightly late on this, if none of the conditions are evaluated then the return is Non, don't we need an else here? also why we are not returned immediately rather than wait till the end, for example, if search_tab == SearchTabs.ALL.value:  return label, also no checking on the search_tab before usage

etna/search/views.py
```diff
@@ -583,7 +583,7 @@ class CatalogueSearchView(BucketsMixin, BaseFilteredSearchView):
     default_group = "tna"
     form_class = CatalogueSearchForm
     template_name = "search/catalogue_search.html"
-    title_base = SearchTabs.CATALOGUE.value
```
mozaki1, line 586: I can see the class attribute name connected to the class name , before both had base on them

etna/search/views.py
```diff
@@ -294,10 +294,10 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
         data = super().get_datalayer_data(request)
         if self.form.cleaned_data.get("group"):
             custom_dimension8 = (
-                self.title_base + ": " + self.form.cleaned_data.get("group")
+                self.search_tab + ": " + self.form.cleaned_data.get("group")
             )
         else:
-            custom_dimension8 = self.title_base + ": " + "none"
+            custom_dimension8 = self.search_tab + ": " + "none"
```
mozaki1, line 300: Why  custom_dimension8 is a tuple here and the else is not?

etna/search/templatetags/search_tags.py
TNA-Allan: **Yup, its late - Its merged already.**

> if none of the conditions are evaluated then the return is Non, don't we need an else here?

None/else is not evaluated in this case as input param come from  enum SearchTabs itself.

>  also why we are not returned immediately rather than wait till the end, 

1. Once the else condition is evaluated. the next step is return - its a small piece of code.
2. I thought it would be cleaner to have one return than multiple returns in the if-else conditions

etna/search/views.py
TNA-Allan: > Why custom_dimension8 is a tuple here and the else is not?

custom_dimension8  is not a tuple, its a string enclosed in brackets. **Probably from some changes earlier. Not with this code base.**

etna/search/views.py
TNA-Allan: Update to use meaningful name. Its the search tab eventually.

etna/search/templatetags/search_tags.py
mozaki1: The only thing here is the full elifs will be evaluated before any return, but if one of them is true then just return immediately faster and less buggy :)

TNA-Allan: > None/else is not evaluated in this case as input param come from enum SearchTabs itself.

mozaki1: Not true, I can call the function passing None like :search_title(None) :), anyway the point I would like to make is we may need to be slightly more protective over our code.

TNA-Allan: Noted. Thank you.

PR #536, opened 2022-11-18
DF-334: datalayer filter count update

TNA-Allan:
Related ticket(s):
https://national-archives.atlassian.net/browse/DF-334

## About these changes

- Updates customMetric2 to reflect selected filters count across sum of all selected filter values

- provisions the same count as a context var to be used for Mobile UI

## How to check these changes

Verify count in customMetric2 of datalayer for selected filter count in desktop view.

- [x] Checked things thoroughly myself before handing over to reviewer.
- [x] Included the ticket number in the PR title to help us keep track of changes.

## For Reviewer

Once PR is merged, check and arrange to deploy on platform-sh.

### DIFF
--- a/etna/search/views.py
+++ b/etna/search/views.py
@@ -298,15 +298,12 @@ class BaseSearchView(SearchDataLayerMixin, KongAPIMixin, FormView):
             )
         else:
             custom_dimension8 = self.search_tab + ": " + "none"
+
         custom_dimension9 = self.form.cleaned_data.get("q") or "*"
-        try:
-            custom_metric2 = len(self.get_api_filter_aggregations(self.form)) - 1
-        except AttributeError:  # This is needed as it will throw an error as some pages do not have filters (e.g. Featured Search, or Record Creator bucket)
-            custom_metric2 = 0
+
         data.update(
             customDimension8=custom_dimension8,
             customDimension9=custom_dimension9,
-            customMetric2=custom_metric2,
         )
         return data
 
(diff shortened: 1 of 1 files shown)


TNA-Allan:
@mozaki1 / @jamesbiggs , please can any of you review this PR. Thanks.

jamesbiggs:
Hi Allan, record count (customMetric1) isn't working on "record creators" on catalogue results. I checked on develop and it's on there too, so not an issue with your code or anything.

customMetric2 working great though 😄

etna/search/views.py
```diff
@@ -474,9 +474,23 @@ def process_api_result(self, form: Form, api_result: Dict[str, Any]):
...
+        )
+        return data
+
     def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
         context = super().get_context_data(**kwargs)
         context["selected_filters"] = self.get_selected_filters(self.form)
+
+        # sum of all values accross all selected filters
+        self._custom_metric2 = context["selected_filters_count"] = sum(
+            map(len, context["selected_filters"].values())
```
mozaki1, line 491: Hi Allan,
 I am not sure what is it doing: sum(map(len, context["selected_filters"].values()), can you please explain?

TNA-Allan:
> Hi Allan, record count (customMetric1) isn't working on "record creators" on catalogue results. I checked on develop and it's on there too, so not an issue with your code or anything.

Nice spot , lets address this separately through another ticket.

etna/search/views.py
TNA-Allan: Hi Mo,
The explanation is given just above in the comment in simple terms. Furthermore, [map](https://docs.python.org/3.9/library/functions.html#map) calculates the length of each context value (list), then sum adds them up.  Python offers to create neat one liner code like this.

mozaki1: So map will apply the length to each "context["selected_filters"].values()" , the resulet will be another iterable transformation after applying the len function, for example [20,40,60,70] , then the sum, will sum the array to 190, which means you will have the selected_filters_count=190 , is that correct?

mozaki1 (approved):

PR #544, opened 2022-12-01
UN-445: BUG Remove repeating feedback mechanism

jamesbiggs:
Related ticket(s):
https://national-archives.atlassian.net/browse/UN-445

## About these changes

Hidden CTA survey banner

## How to check these changes

Make sure there is only one feedback mechanism

## Before assigning to reviewer, please make sure you have

- [ ] Checked things thoroughly myself before handing over to reviewer.
- [ ] Included the ticket number in the PR title to help us keep track of changes.
- [ ] Ensured that PR includes only commits relevant to the ticket.
- [ ] Waited for all CI jobs to pass before requesting a review.
- [ ] Added/updated tests and documentation where relevant.

## For Reviewer

Once PR is merged, check and arrange to deploy on platform-sh.

PR #539, opened 2022-11-23
Df 388 mobile filter value

JohnHeeryTNA:
Related ticket(s):
https://national-archives.atlassian.net/browse/DF-388

## About these changes

adding in the filter value for mobile search ui

## How to check these changes

Please test using a variety of terms and applying filters

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly myself before handing over to reviewer.
- [x] Included the ticket number in the PR title to help us keep track of changes.
- [x] Ensured that PR includes only commits relevant to the ticket.
- [x] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

## For Reviewer

Once PR is merged, check and arrange to deploy on platform-sh.

PR #545, opened 2022-12-01
Release 3.1.1 to main

jamesbiggs:
**Brings in:**
c5ac66f UN-445: Add display: none to cta-banner-survey (#544)
3b38ed5 Df 388 mobile filter value (#539)
d186497 DF-334: datalayer filter count update (#536)


scripts/src/modules/search/mobile-filter-expander.js
```diff
@@ -8,15 +9,11 @@ export default function() {
     if(!$searchFilterContainer || !$searchGrid || !$main) {
         return;
     }
+    //need case where validation warning is invoked
 
     //need case where validation warning is invoked
```
TNA-Allan, line 14: The comments are duplicated. Should be one line as in https://github.com/nationalarchives/ds-wagtail/blob/39cb00d0798be18967e7f5269e7312ead95b8229/scripts/src/modules/search/mobile-filter-expander.js#L12

PR #599, opened 2023-01-25
CHORE: Fix Reference to `PageRevision` in data migration

ababic:
We've noticed on other PRs that tests can fail due to app migrations running in an unpredictable order.

Previously, the data migration in the `insights` app would always run BEFORE the migration to rename the `PageRevision` model in wagtail - So, it was fine to carry on using that model name. However, when migration changes cause Django to run the wagtail migration BEFORE reaching ours, the model name is no longer valid, and we see the errors in CI again.

Because Django migrations do not have a way to specify "Run this migration BEFORE these other ones", but do have a way to specify "Run this migration AFTER these other ones", the safer/more consistant fix is to make sure our data migration runs AFTER the rename ones, and use the updates `Revision` model name so that the error goes away.


TNA-Allan (approved):
Thank you Andy this is superb! :)

PR #600, opened 2023-01-25
UN-324: Edit admin page to show the difference between environments

jamesbiggs:
Related ticket(s):
https://national-archives.atlassian.net/browse/UN-324

## About these changes

Changed colour Wagtail CMS colour scheme to differentiate between the environments

## How to check these changes

To test on your local you can overwrite `FEATURE_PLATFORM_ENVIRONMENT_TYPE` in `base.py` and set it to "development" or "production" to get different colour schemes.
After merging, it can be deployed to develop (and ultimately main/prod) to show that the different styles and text have been applied.

## For Reviewer

- Before merging PR

    `main` branch :

    `develop` branch:

        - select `Squash and merge`

## For Reviewer/Developer

- After merging PR:


jamesbiggs:
Please don't merge until the issue with develop is sorted 🙂

etna/core/templates/wagtailadmin/home.html
```diff
@@ -0,0 +1,4 @@
+{% extends "wagtailadmin/home.html" %}
+
+
+{% block branding_welcome %}Welcome to the etna {% if FEATURE_PLATFORM_ENVIRONMENT_TYPE == "staging" or FEATURE_PLATFORM_ENVIRONMENT_TYPE == "development" %}DEVELOP {% endif %}Wagtail CMS{% endblock %}
```
William-Blackie, line 4: Suggestion: Instead of checking that it is either `staging` or `development`. Why not just check it's not `production`?  
```suggestion
{% block branding_welcome %}Welcome to the etna {% if FEATURE_PLATFORM_ENVIRONMENT_TYPE != "production" %}DEVELOP {% endif %}Wagtail CMS{% endblock %}
```

Note: I'm not 100% sure of the prod' naming convention so just double check.

etna/core/wagtail_hooks.py
```diff
@@ -9,3 +10,10 @@ def editor_css():
     return format_html(
         '<link rel="stylesheet" href="{}">', static("css/dist/etna-wagtail-editor.css")
     )
+
+
+@hooks.register("insert_global_admin_css")
+def global_admin_css():
+    if settings.FEATURE_PLATFORM_ENVIRONMENT_TYPE in ["development", "staging"]:
```
William-Blackie, line 17: Note: A similar style of feedback applies to this line, we could just check if it's production, otherwise show the colours.

These changes are non-blocking. From our original conversation, I thought there was going to be multiple different styling (dev + staging).

William-Blackie (commented):
Hey @jamesbiggs,

nice work here, I've offered some suggestions on improvements but these are non-blocking. You'll notice in code reviews I write them a little different than you've seen before, I follow this style: https://conventionalcomments.org/ take a read!

Thanks - Will.

etna/core/templates/wagtailadmin/home.html
jamesbiggs: Hi Will, cheers for the suggestions - I agree that this is a better way to do it syntactically, although, it may cause some issues. Where I set up the .env variable in `base.py`, I made the default be `""`. I did this because I didn't want to manually assign it to a "default" environment.
This means that, if the `os.getenv` fails, it will be given `""` which will then default to the standard Wagtail styling. If I made the suggestions here, I would then have to make it default to `"production"`, which isn't a bad thing.

What are your thoughts on changing the default value to `"production"` instead of `""`?

William-Blackie: Personally, the default should be a production environment, a developer should have to configure via env' vars for others. How that's done doesn't particularly matter so your approach does work above. However, you can alter that conditional statement to allow for the behaviour just checking if there's a value, otherwise show the production version. 

I'll leave that choice to you. Feel free to merge.

William-Blackie (approved):

PR #606, opened 2023-01-28
Fix: RecordChooserField incompatibility with Wagtail's revision framework

ababic:
## About these changes

While our custom `RecordChooserField` works fine for the `Highlight` snippet model; when used on a model that uses revisions (e.g. `RecordArticlePage.record`), the value gets misinterpreted during the serialisation process, and a full string representation of the record ends up as the stored value (e.g. "Record title (IAID)"), which the field isn't designed to handle.

The key change here is the addition of the `value_to_string()`, which is responsible for returning a value suitable for serialisation. Since it's an `iaid` string that is usually stored in the database, it makes sense that the same value also be used as the serialised value.

There are a couple of extra changes here too:

1. I renamed the model field to `RecordField` because the interface shouldn't be a concern for model fields. While they can and do specify default form fields and widgets, those things are supposed to be interchangeable.
2. The 'value translation' logic has been split into two private methods to improve consistency:
    - `RecordField._convert_to_record_instance()`
    - `RecordField._extract_record_iaid()`

## How to check these changes

Add a "Record Revealed article" page below the "Collection Explorer" page (Explorer index page). You should be able to choose a record for the `record` field, publish the page, then reopen the edit view and see the same record selected.

## For Reviewer

- Before merging PR

    `main` branch :

    `develop` branch:

        - select `Squash and merge`

## For Reviewer/Developer

- After merging PR:

### DIFF
--- a/etna/articles/migrations/0048_add_recordarticlepage_and_pagegalleryimage.py
+++ b/etna/articles/migrations/0048_add_recordarticlepage_and_pagegalleryimage.py
@@ -54,7 +54,7 @@ class Migration(migrations.Migration):
                 ),
                 (
                     "record",
-                    etna.records.fields.RecordChooserField(
+                    etna.records.fields.RecordField(
                         db_index=True, verbose_name="record"
                     ),
                 ),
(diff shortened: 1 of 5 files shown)


jamesbiggs (approved):
Looks good Andy, cheers!

PR #626, opened 2023-02-02
UN-451: Tweak contentgroup1 datalayer label for articles models

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/UN-451

## About these changes

The ask was to change the current "Stories" value that is displayed for page types in the `articles` app, which I did - but, I figured it was also high time to ditch the "Work out the label from the app name" mechanism, as it's simply too clever for it's own good, and doesn't lend itself well to doing things differently in future (we already need to use a different value for record articles)

## How to check these changes

## Merging a release branch?

## Deployment guidance

### DIFF
--- a/etna/analytics/mixins.py
+++ b/etna/analytics/mixins.py
@@ -10,7 +10,7 @@ class DataLayerMixin:
     datalayer.
     """
 
-    gtm_content_group: str = "Page"
+    gtm_content_group: str = None
 
     def get_gtm_content_group(self) -> str:
         """
@@ -21,6 +21,10 @@ class DataLayerMixin:
         sufficient, but this method can be overridden to support cases where
         different per-instance values are required.
         """
+        if self.get_gtm_content_group is None:
+            raise NotImplementedError(
+                f"You must set the 'gtm_content_group' attribute on {self.__class__} to a string. Alternatively, you can override the 'get_gtm_content_group()' method to return a different value(s)."
+            )
         return self.gtm_content_group
 
     def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
--- a/etna/analytics/mixins.py
+++ b/etna/analytics/mixins.py
@@ -10,8 +10,18 @@ class DataLayerMixin:
     datalayer.
     """
 
+    gtm_content_group: str = "Page"
+
     def get_gtm_content_group(self) -> str:
-        raise NotImplementedError
+        """
+        Return a string to use as the 'contentGroup1' value in DataLayer
+        for this object.
+
+        For most cases, setting `gtm_content_group` on the class should be
+        sufficient, but this method can be overridden to support cases where
+        different per-instance values are required.
+        """
+        return self.gtm_content_group
 
     def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
         """
@@ -62,8 +72,7 @@ class SearchDataLayerMixin(DataLayerMixin):
     datalayer.
     """
 
-    def get_gtm_content_group(self) -> str:
-        return "Search"
+    gtm_content_group: str = "Search"
 
     def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
         data = super().get_datalayer_data(request)
--- a/etna/articles/models.py
+++ b/etna/articles/models.py
@@ -47,6 +47,9 @@ class ArticleIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
 
     new_label_end_date = datetime.now() - timedelta(days=21)
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Stories from the archives"
+
     class Meta:
         verbose_name = _("article index page")
 
@@ -120,6 +123,9 @@ class ArticlePage(
     ]
 
     new_label_end_date = datetime.now() - timedelta(days=21)
+
+    # DataLayerMixin overrides
+    gtm_content_group = "Stories from the archives"
 
     template = "articles/article_page.html"
 
--- a/etna/articles/models.py
+++ b/etna/articles/models.py
@@ -48,7 +48,7 @@ class ArticleIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
     new_label_end_date = datetime.now() - timedelta(days=21)
 
     # DataLayerMixin overrides
-    gtm_content_group = "Stories from the archives"
+    gtm_content_group = "stories"
 
     class Meta:
         verbose_name = _("article index page")
@@ -125,7 +125,7 @@ class ArticlePage(
     new_label_end_date = datetime.now() - timedelta(days=21)
 
     # DataLayerMixin overrides
-    gtm_content_group = "Stories from the archives"
+    gtm_content_group = "stories"
 
     template = "articles/article_page.html"
 
--- a/etna/articles/models.py
+++ b/etna/articles/models.py
@@ -282,6 +282,9 @@ class RecordArticlePage(
         verbose_name=_("featured article"),
     )
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Record articles"
+
     class Meta:
         verbose_name = _("record article")
         verbose_name_plural = _("record articles")
--- a/etna/collections/models.py
+++ b/etna/collections/models.py
@@ -51,6 +51,9 @@ class ExplorerIndexPage(AlertMixin, TeaserImageMixin, MetadataPageMixin, BasePag
         "articles.RecordArticlePage",
     ]
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
+
 
 class TopicExplorerIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
     """Topic explorer BasePage.
@@ -69,6 +72,9 @@ class TopicExplorerIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
         FieldPanel("body"),
     ]
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
+
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
 
     @cached_property
     def featured_pages(self):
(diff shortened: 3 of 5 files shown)


etna/analytics/mixins.py
```diff
@@ -10,8 +10,18 @@ class DataLayerMixin:
     datalayer.
     """
 
+    gtm_content_group: str = "Page"
+
```
TNA-Allan, line 14: The `DataLayerMixin` mixin is not a `Page`. Why give it a value `Page` ?
How about `None` ?

etna/home/models.py
```diff
@@ -40,6 +40,9 @@ class HomePage(AlertMixin, TeaserImageMixin, MetadataPageMixin, BasePage):
     settings_panels = BasePage.settings_panels + AlertMixin.settings_panels
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Homepage"
+
```
TNA-Allan, line 45:
```suggestion
    # DataLayerMixin overrides
    gtm_content_group = "Stories from the collection"

```

etna/articles/models.py
```diff
@@ -46,6 +46,9 @@ class ArticleIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
 
     new_label_end_date = datetime.now() - timedelta(days=21)
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Stories from the archives"
+
```
TNA-Allan, line 51: How about `gtm_content_group = HomePage.gtm_content_group`?

etna/articles/models.py
```diff
@@ -120,6 +123,9 @@ class ArticlePage(
 
     new_label_end_date = datetime.now() - timedelta(days=21)
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Stories from the archives"
```
TNA-Allan, line 127: How about `gtm_content_group = HomePage.gtm_content_group` ?

etna/collections/models.py
```diff
@@ -70,6 +73,9 @@ class TopicExplorerIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage):
     ]
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
```
TNA-Allan, line 77: How about `gtm_content_group = ExplorerIndexPage.gtm_content_group` ?

etna/collections/models.py
```diff
@@ -126,6 +132,9 @@ class TopicExplorerPage(AlertMixin, TeaserImageMixin, MetadataPageMixin, BasePag
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
     settings_panels = BasePage.settings_panels + AlertMixin.settings_panels
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
```
TNA-Allan, line 136: How about `gtm_content_group = ExplorerIndexPage.gtm_content_group` ?

etna/collections/models.py
```diff
@@ -167,6 +176,9 @@ class TimePeriodExplorerIndexPage(TeaserImageMixin, MetadataPageMixin, BasePage)
     ]
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
```
TNA-Allan, line 180: How about `gtm_content_group = ExplorerIndexPage.gtm_content_group` ?

etna/collections/models.py
```diff
@@ -227,6 +239,9 @@ class TimePeriodExplorerPage(AlertMixin, TeaserImageMixin, MetadataPageMixin, Ba
     promote_panels = MetadataPageMixin.promote_panels + TeaserImageMixin.promote_panels
     settings_panels = BasePage.settings_panels + AlertMixin.settings_panels
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
```
TNA-Allan, line 243: How about `gtm_content_group = ExplorerIndexPage.gtm_content_group` ?

etna/collections/models.py
```diff
@@ -279,6 +294,9 @@ class ResultsPage(AlertMixin, TeaserImageMixin, MetadataPageMixin, BasePage):
     ]
     subpage_types = []
 
+    # DataLayerMixin overrides
+    gtm_content_group = "Explorer"
```
TNA-Allan, line 298: How about `gtm_content_group = ExplorerIndexPage.gtm_content_group` ?

etna/home/models.py
ababic: Hi @TNA-Allan. Where did this value come from? Take a look at the [homepage](https://develop-sr3snxi-rasrzs7pi6sd4.uk-1.platformsh.site/) and look for the following script tag in the head:
```
<script id="gtmDatalayer" type="application/json">{"contentGroup1": "Homepage", "customDimension1": "offsite", "customDimension2": "", "customDimension3": "home page", "customDimension4": "", "customDimension5": "", "customDimension6": "", "customDimension7": "", "customDimension8": "", "customDimension9": "", "customDimension10": "", "customDimension11": "", "customDimension12": "", "customDimension13": "", "customDimension14": "", "customDimension15": "", "customDimension16": "", "customDimension17": ""}</script>
```

We need to preserve the values that display currently (they were [set here](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/core/models/basepage.py#LL19-L23C2) previously)

etna/articles/models.py
ababic: Again, this is the value that was previously determined by [this code here]([set here](https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/core/models/basepage.py#LL19-L23C2)). Why would we want this value to be different?

etna/collections/models.py
ababic: Because it's just a string. There's no reason why any of the strings on any of the models shouldn't be able to change at any point in time, and that shouldn't impact any other page type.

ababic: In other words... its purely coincidence that the value is the same for all page types in this module. There's absolutely no need for that coincidence to influence the implementation.

etna/home/models.py
TNA-Allan: The requirement says **Stories from the collection** for all story pages (including the homepage)

ababic: I think there's some misunderstanding there. I think @hoake meant that the 'index page' for the articles, not the 'home page'.

etna/collections/models.py
ababic: Yes, but what about when the value changes for one model, but not another?

TNA-Allan: > Yes, but what about when the value changes for one model, but not another?

Since the `gtm_content_group`  can override  DataLayerMixin, it can be updated to the required values when required, until then it adopts the common value.

ababic: Sorry Allan, but the whole idea behind this refactor is to simplify this as much as possible. I know it's very natural to want to reduce duplication in the codebase, but I really don't think it's worthwhile in this particular case. It should be clear from a ticket's requirements what needs to change and what doesn't, and a mass search/replace is just as easy to action as single one.

TNA-Allan: >... but the whole idea behind this refactor is to simplify this as much as possible. I know it's very natural to want to reduce duplication in the codebase, but I really don't think it's worthwhile in this particular case. 

The existing implementation for `get_gtm_content_group`  avoided duplication of code as it done now. 


> _Yes, but what about when the value changes for one model, but not another?_

_Since the gtm_content_group can override DataLayerMixin, it can be updated to the required values when required, until then it adopts the common value._

What is the problem with accessing another class attribute? 

> It should be clear from a ticket's requirements what needs to change and what doesn't, 

Do you want to clarify the requirements and proceed from there?

> a mass search/replace is just as easy to action as single one.

This does not look good. That is why we have `etna\ciim\constants.py`

ababic: I'm sorry Allan, but this is not what constants are for. They are for variables that have wider meaning throughout the codebase. These are just 'labels' without any wider meaning. You look at the class, you know what the value is just by looking at the code, you change it if it needs to change, and it doesn't affect anything else... Pure and simple.

TNA-Allan: > You look at the class, you know what the value is just by looking at the code, you change it if it needs to change, and it doesn't affect anything else... Pure and simple.

This seems contradicting to what you mentioned. -
>  and a mass search/replace is just as easy to action as single one.

Previous queries:
> What is the problem with accessing another class attribute?

Is there an issue with this approach?

> Do you want to clarify the requirements and proceed from there?

ababic: > This seems contradicting to what you mentioned

I'm honestly not seeing what you mean. Both my points prioritise simplicity and defend the current design. Using constants or some reference to a value defined elsewhere... It's all just unnecessary complexity. Like I said at the beginning: It is only coincidence that models share this value. There's no reason the value for any model shouldn't be changed in future.

TNA-Allan: > Using constants or some reference to a value defined elsewhere... It's all just unnecessary complexity.

I thought using a reference to a value defined elsewhere will eliminate the trouble of updating the same value everywhere.  That's why I said
> The same string name is used across different pages. Making changes in one place is convenient than same changes in various places.

> It's all just unnecessary complexity.

If your view is about making some reference to a value defined elsewhere in this case _complex_, I will go with your experienced view to keep the same name defined for `gtm_content_group` across different models.

etna/articles/models.py
TNA-Allan: As revised in the requirements,  `contentGroup1= "stories"` for all story pages i.e. Article index page, Article.

etna/articles/models.py
TNA-Allan: As revised in the requirements, `contentGroup1= "stories"` for all story pages i.e. Article index page, Article.

etna/home/models.py
TNA-Allan: As revised in the requirements, contentGroup1= "stories" for all story pages i.e. Article index page, Article. No change for contentGroup1 for Home page.

TNA-Allan (changes requested):

etna/analytics/mixins.py
ababic: I think a blank string makes sense, as that way we're always dealing with a consistent data type

TNA-Allan: Sure, go for it. Thanks.

ababic: Actually, I went with `None` as the default, as it allows us to do some detection and raise an error when the value hasn't been set (or the method overridden)

TNA-Allan (approved):
Looks good. Thanks :)

PR #640, opened 2023-02-10
Fix: Update fab command to clear the new renditions table instead of the old one

ababic:


TNA-Allan (approved):
Looks good. Thanks :)

PR #620, opened 2023-02-01
Un 387 frontend record revealed

William-Blackie:
Related ticket(s):
- <- https://national-archives.atlassian.net/browse/UN-204 ->

## About these changes
Here I've added a image gallery with tabbs for JS users and a CSS only version of non-js users. I've used Axe dev'tools to check and I'm not getting any warnings specifically for this feature but would appreciate @janehughes input when you have a moment.

## How to check these changes
You'll find these changes in on a CloserLookPage type, I made one under the CollectionExplorer, input the record revealed fields and this will populate the FE. 

Tab from the start of the doccument and you will be able to; open/close the gallery, select and change tabs, focus the content.

For non-js users (Disable this in your browser) you will see that the I've rendered the tabs one after another as that seemed best to fit the content, although no design is provided for this. Happy for a screenshare.

Thanks - Will.

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly before handing over to reviewer.
- [x] Checked PR title starts with ticket number as per project conventions to help us keep track of changes.
- [ ] Ensured that PR includes only commits relevant to the ticket.
- [ ] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

## For Reviewer

- Before merging PR

    `main` branch :

        - check PR merge tile begins with `release/<major.minor.patch>:<default/custom description>`
        - check PR merge tile begins with ticket number Ex `DF-XXX: Ticket name / short description`
        - select `Create a merge commit`

    `develop` branch:

        - check if merges into base branch are `not to be kept On Hold` and then proceed.
        - check PR merge tile begins with ticket number Ex `DF-XXX: Ticket name / short description`
        - select `Squash and merge`

## For Reviewer/Developer

- After merging PR:

  - check and arrange to deploy on platform-sh.
  - update status on JIRA ticket on deployment.
  - update slack on deployment progress of the JIRA ticket once deployed.

PR #641, opened 2023-02-10
Un 387 frontend record revealed

William-Blackie:
Related ticket(s):
- <- https://national-archives.atlassian.net/browse/UN-204 ->

## Note: Due to the merges the base branch is no-longer stable (or atleast I do not have the project knowledge to rebase it well enough to not destroy this branch). I've cherry-picked files across and fixed where I could. 

Some changes are still outsanding from the original MR: https://github.com/nationalarchives/ds-wagtail/pull/620 but should be stable enough to accessibility check.

## About these changes
Here I've added a image gallery with tabbs for JS users and a CSS only version of non-js users. I've used Axe dev'tools to check and I'm not getting any warnings specifically for this feature but would appreciate @janehughes input when you have a moment.

## How to check these changes
You'll find these changes in on a CloserLookPage type, I made one under the CollectionExplorer, input the record revealed fields and this will populate the FE. 

Tab from the start of the doccument and you will be able to; open/close the gallery, select and change tabs, focus the content.

For non-js users (Disable this in your browser) you will see that the I've rendered the tabs one after another as that seemed best to fit the content, although no design is provided for this. Happy for a screenshare.

Thanks - Will.

## For Reviewer

- Before merging PR

    `main` branch :

    `develop` branch:

        - select `Squash and merge`

## For Reviewer/Developer

- After merging PR:


sass/includes/_transcription.scss
```diff
@@ -0,0 +1,147 @@
...
+      border-bottom: $color__white solid 3px;
+    }
+
+  }
+
+  &__tabpanel {
+    margin-top: 2.5rem;
+    color: $color__white;
+
+    >p {
```
William-Blackie, line 142: I'd have to nest this class > rich-text > p to get this without this selector. Feels cleaner, what do you think @janehughes ?

templates/home/home_page.html
```diff
@@ -36,44 +36,7 @@
...
-                </h2>
-
-                <p id ="article-desc" class="aria-desc">This is a link to a new story about {{ page.featured_article.title }}</p>
-
-                <p>{{ page.featured_article.sub_heading }}</p>
-                <a class="tna-button--dark" href="{% pageurl page.featured_article %}" data-component-name="Featured Story: {{ page.featured_article.title }}" data-link-type="Button" {% if page.featured_article.first_published_at.date > page.new_label_end_date.date %}data-label = "New"{%endif%}>Read</a>
-            </div>
-        </div>
-    </div>
+        {% include "includes/article-spotlight.html" with page=page.featured_article %}
```
William-Blackie: Refactoring here to use the same template

templates/includes/article-spotlight.html
```diff
@@ -0,0 +1,36 @@
...
+        <div class="featured-article__description">
+            <h2 class="featured-article__heading">
+                <small>{% trans "Spotlight on" %}</small>
+                {{ page.title }}
+                {% if page.first_published_at.date > page.new_label_end_date.date %}
+                    <span aria-describedby="article-desc" role="link">NEW</span>
+                {% endif %}
+            </h2>
+
+            <p id="article-desc" class="aria-desc">This is a link to a new story about {{ page.title }}</p>
```
William-Blackie, line 30: Note: I've not seen this before, is this something we should be doing?

templates/includes/more-records.html
```diff
@@ -0,0 +1,23 @@
+{% load static %}
+{% load wagtailimages_tags %}
+{% load wagtailcore_tags %}
+
+<div class="more-records">
```
William-Blackie, line 5: These related pages do not exist yet, I can leave the template but cannot check it as the BE it was made for no longer exists.

William-Blackie (commented):
Hey all,

I've picked up the changes from the previous as well as made adjustments to the other files that was abscured by the MR being based off of WIP work, this no longer exists or was merged so no longer exists.

The related pages in the design AKA "Browser more records" and "Blogs and podcasts" no longer exist in the BE although there is some artifacts from the orignal MR. I've left them for now as it should be straight forward to tweak them but also could remove. This work will need to be split out to keep momentum. 

@janehughes, as John is on anual leave do you have capacity for a more general review and the accessibility?

templates/includes/article-spotlight.html
janehughes: If we provide more context to the button label (e.g. Read about the Caravan Club) then I don't think this is necessary.

templates/includes/more-records.html
janehughes: 👍

templates/includes/record-matters.html
```diff
@@ -0,0 +1,61 @@
...
+    href="{% record_url record is_editorial=True %}">{{ record.reference_number }}</a>
+                            </p>
+                        {% endif %}
+                    </div>
+                {% endif %}
+                {% if record.description %}
+                    <div id="readLessContent" class="hidden">
+                        {{ record.description|richtext|truncatewords_html:50 }}
+                        <button id="readMoreButton" class="record-matters__plain-button" href="">
+                            Read more
```
janehughes, line 26: I think this could be more descriptive, e.g. 'Expand to read more about {{ name }}' or similar - if that's too long visually we could add it as sr-only text or an aria-label

janehughes (commented):
Thanks Will! Have reviewed what's there and added a few questions/comments. Some are more general thoughts on how we're approaching styles and could be added to a tech debt to-do list as opposed to fixing it all here.

janehughes:
Here are my more specific notes following accessibility & UI review:

### Gallery
- I feel the image gallery itself would benefit from a section with an aria-label e.g. (<section aria-label=“Record Image gallery”></section>) to distinguish it from the rest of the page. Gives users a quick overview to identify its purpose without having to tab through all the content.
- ‘View images and transcripts’ needs cursor: pointer to indicate it’s a button
- After opening the gallery, user’s focus should move to the first bit of content inside the gallery ‘image 1 of 3’ - currently this gets missed either a) when pressing tab and jumping straight to ‘Transcript’ or b) using the arrow key, which seems to jump back to the ‘Record revealed’ heading. Need to make sure the tab order remains logical and content isn’t accidentally missed.
- Tablist has an aria-labelledby attribute ‘aria-labelledby="tablist-1-1”’  which doesn’t have a matching element
- Tabs need cursor: pointer to indicate they are buttons
- ‘Close’ button should look like a button rather than a link or static text (e.g. add a border/background colour) to make it stand out and help users identify it more easily
- ‘Close’ button could benefit from more descriptive aria-label (e.g. Close image gallery) - ‘close’ on its own isn’t sufficiently descriptive

### General page 
(I understand the page is still somewhat WIP so please ignore if these are WIP issues)
- Increase spacing between the ‘Record revealed’ page intro and the gallery
- The story of this record ‘Read more’ button doesn’t seem to be working correctly. Also needs aria-expanded attributes applied
- Colours need updating for CTA button in ‘Spotlight on’ block, currently doesn’t pass colour contrast

### UI/UX comments/suggestions

Having reviewed the static designs, it feels like there’s a lot of content at the bottom of the page which dilutes the potential impact of each section. I suggest we keep this bottom part of the page section very intentional and make sure we’re signposting users to the most helpful / relevant places to avoid overloading them with choices. 

- ‘Related records’ makes sense to keep as the most prominent section, signposting similar content the user might be interested in and help to keep them browsing
- ‘Related blogs and podcasts’ for a similar reason would be helpful supportive content- these should have less prominence than related records - consider removing the images and only keep titles/summary for each.
- Consider either removing one or both of Stories/More to explore sections. If we’re keen to keep both, I suggest reducing these to simple buttons, to avoid distracting the users attention from the more relevant sections.

sass/includes/_transcription.scss
William-Blackie: You're right. It's been turned off in the latest version so it now works. Yep good spot, added that.

templates/includes/article-spotlight.html
William-Blackie: That was one in one design but not the one I was given / working against. I'll leave as is.

templates/includes/record-matters.html
William-Blackie: Good point, added a label.

William-Blackie (commented):
Thanks for the review @janehughes

sass/includes/_more-records.scss
```diff
@@ -0,0 +1,52 @@
+.more-records {
+    background:#fbf8e9;
```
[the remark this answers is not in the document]
William-Blackie: I've pulled this out as it's legacy.

JohnHeeryTNA (approved):
Approving for merge

PR #645, opened 2023-02-14
DF-454: API to use fetch metadataId(s) instead of iaid(s)

TNA-Allan:
Ticket URL: https://national-archives.atlassian.net/browse/DF-454

## About these changes

Removes `iaid(s)` which will be deprecated and use `metadataId(s)` for `fetch/fetchAll` API's

## How to check these changes

As mentioned in the ticket.

## Merging a release branch?

## Deployment guidance


etna/ciim/client.py
```diff
@@ -405,14 +405,15 @@ def fetch_all(
...
+        replicaId (rid).
 
         Keyword arguments:
 
         ids:
             Generic identifiers. Matches on references_number or iaid
-        iaids:
-            Return matches on Information Asset Identifier
+        metadata_ids:
+            Return match on @admin.id for various Information Asset Identifier (IAID) formats.
```
ababic, line 416: Again, the `ids` option should really be updated to support `metadata_id` values, and we should do away with this option really.

etna/ciim/utils.py
```diff
@@ -205,7 +205,9 @@ def resolve_links(markup):
     def link_from_span(span):
         if link := pq(span).attr("link"):
             if iaid := re.match(r"\$link\((?P<iaid>[C0-9]*)\)", link).group("iaid"):
-                url = reverse("details-page-machine-readable", kwargs={"iaid": iaid})
+                url = reverse(
+                    "details-page-machine-readable", kwargs={"metadata_id": iaid}
+                )
```
ababic, line 210: There's still a weird mix of terminology here, which I think could cause confusion

etna/collections/models.py
```diff
@@ -323,7 +323,7 @@ def record(self):
         skip this record on the results BasePage.
         """
         try:
-            return Record.api.fetch(iaid=self.record_iaid)
+            return Record.api.fetch(metadata_id=self.record_iaid)
```
ababic, line 326: Still mixing terminology here, too

etna/records/blocks.py
```diff
@@ -61,7 +61,7 @@ def bulk_to_python(self, values):
         if not values:
             return []
         records_by_id = {
-            r.iaid: r for r in self.target_model.api.fetch_all(iaids=values)[1]
+            r.iaid: r for r in self.target_model.api.fetch_all(metadata_ids=values)[1]
```
ababic, line 64: Looks like we're still taking the `iaid` value of the records here as the canonical id.

ababic (changes requested):
Hi Allan. Not a bad start at all... I just think there are a few places that leave us with a mix of terminology. If `metadata_id` is becoming the canonical identifier for all records, I feel there would be great value in abstracting that away, and just calling it `id` project-wide.

TNA-Allan:
> Hi Allan. Not a bad start at all... I just think there are a few places that leave us with a mix of terminology. If `metadata_id` is becoming the canonical identifier for all records, I feel there would be great value in abstracting that away, and just calling it `id` project-wide.

Thank you Andy.  There is a sperate api param `id` which serves a different purpose than `metadataId`. I think since the term `metadata_id` is used in the API call and its purpose to extract various `iaids` from @admin.id, we should go with the same name as the calling parameter to avoid any confusion as to which param is used in the call. Using `metadataId` matches the URL and API call.

etna/ciim/client.py
```diff
@@ -171,8 +171,8 @@ def fetch(
 
         Keyword arguments:
 
-        iaid:
-            Return match on Information Asset Identifier
+        metadata_id:
+            Return match on @admin.id for various Information Asset Identifier (IAID) formats.
         id:
             Generic identifier. Matches on references_number or iaid
```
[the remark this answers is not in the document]
TNA-Allan: Noted this - perhaps a discussion will be required

TNA-Allan:
> @TNA-Allan It'd be worth considering whether we still need the `iaid` converter here, or whether that needs loosening to account for new ID formats.

@ababic The `iaid` converter already takes care of various `iaid's` and those will appear inside of @admin.id via the fetch

ababic:
> The iaid converter already takes care of various iaid's and those will appear inside of @admin.id via the fetch

I'm sure the `iaid` converter has some worth. I'm thinking about this more from a terminology and consistency perspective. How would you feel coming to this project anew without knowing the history here? `iaid` looks like a completely separate thing from the outside.

etna/ciim/client.py
ababic: Happy to approve the API client changes if that's out our control for now.

etna/ciim/client.py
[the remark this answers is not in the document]
TNA-Allan: @ababic , it requires availability of those concerned on these points. I have raised them and can get back - but it can be over next week if you want to hold it  at all.

TNA-Allan:
@ababic I've taken the some of the queries and conversation to the ticket to have a discussion if at all.

config/urls.py
```diff
@@ -48,7 +48,7 @@ def trigger_error(request):
 # Public URLs that are meant to be cached.
 public_urls = [
     path(
-        r"catalogue/id/<iaid:iaid>/",
+        r"catalogue/id/<iaid:metadata_id>/",
```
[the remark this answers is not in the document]
TNA-Allan: The `iaid` here is the name of the registered converter. I initially thought to change it as `metadata_id` to match the API and URL param but reverted since the converter  serves the purpose of `iaid` 
The notion for `id` is updated in the ticket. Would that be relevant now?

TNA-Allan:
@ababic 
> @ababic I've taken the some of the queries and conversation to the ticket to have a discussion if at all.
Going with our discussion, currently implemented changes to converter name. Any thing else as of now, let us do or handle it at the time its needed.

etna/records/blocks.py
TNA-Allan: Going with discussions.

etna/collections/models.py
TNA-Allan: Going with discussions.

etna/ciim/utils.py
TNA-Allan: Going with discussions.

config/urls.py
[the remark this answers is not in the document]
TNA-Allan: Going with discussions.

etna/ciim/utils.py
```diff
@@ -205,7 +205,9 @@ def resolve_links(markup):
     def link_from_span(span):
         if link := pq(span).attr("link"):
             if iaid := re.match(r"\$link\((?P<iaid>[C0-9]*)\)", link).group("iaid"):
```
ababic, line 207: Maybe one to look at in a separate PR, but the pattern for recognising the `iaid` here is notably different to patterns used elsewhere. This feels much too verbose.

etna/records/blocks.py
ababic: On this @TNA-Allan. The solution here will currently only work for records with an `.iaid` value. If the point of these changes it to accommodate records without an `iaid`, then shouldn't we use a different attribute here?

etna/records/fields.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, iaid: str):
         self.__dict__["iaid"] = iaid
```
ababic, line 23: If a `Record` can represent any type of record now (ie. with or without an `iaid` value), I don't think the custom field or the chooser should be using/storing the `iaid` value specifically, as that will no longer work in all circumstances. This is another reason why I suggest using a `id` throughout... we're not saying there cannot be other identifiers for different types of record - simply that there is a canonical property that does work in any circumstance.

etna/records/templatetags/records_tags.py
```diff
@@ -62,7 +62,9 @@ def record_url(
         ):
             return f"https://discovery.nationalarchives.gov.uk/details/r/{iaid}"
         try:
-            return reverse("details-page-machine-readable", kwargs={"iaid": iaid})
+            return reverse(
+                "details-page-machine-readable", kwargs={"metadata_id": iaid}
+            )
         except NoReverseMatch:
             pass
     if ref:
```
ababic, line 70: There's a similar issue in this module: Again, we are taking the `.iaid` property value from Records (or `["iaid"]` key from search results). This value won't be available for the new records we hope to support. We need to be using something other than `iaid` from the data here

etna/ciim/utils.py
TNA-Allan: Good spot @ababic . There are separate issues here.
1. The api response requires to match the pattern having link in it. A sperate ticket is raised to resolve these links. [DF-51](https://national-archives.atlassian.net/browse/DF-51) It may require further updates.
2. The Iaid patters are different from the Converter.  Nee to know if that pattern using C0-9 were intentional or otherwise.

etna/records/blocks.py
TNA-Allan: The current api response and attribute have the same name i.e. `iaid` until now, no other source is formalised.

etna/records/fields.py
TNA-Allan: Same as mentioned previously: The current api response and attribute have the same name i.e. `iaid` until now, no other source is formalised.

etna/records/templatetags/records_tags.py
TNA-Allan: Same as mentioned previously: The current api response and attribute have the same name i.e. iaid until now, no other source is formalised.

Further, I mentioned about `primaryIdentifier` we need to know further about this if at all this this the generic one. 

@ababic , perhaps to treat this as a separate issue and lets go with the current status of using metadataId and Iaid. Thoughts?

etna/records/fields.py
ababic: I'm really really confused :(

etna/records/templatetags/records_tags.py
ababic: If we are in limbo with terminology, then we should make the **fewest number of changes we possibly can** for now, then bring everything in-line once we have more certainty. Specifically, I would recommend changing nothing but the API client logic, so that we're using `metadata_id` as the parameter name instead of `iaid`.

etna/records/fields.py
TNA-Allan: Records without `iaid`, this needs further understanding and discussions. I believe we have tickets for them. Also I've mentioned about `primaryIdentifier` [here](https://github.com/nationalarchives/ds-wagtail/pull/645#discussion_r1115864254) . The circumstance of `iaid` has not changed until now. Therefore until then the custom field or chooser field should remain `iaid`. Hope this clarifies.

etna/records/templatetags/records_tags.py
TNA-Allan: >  I would recommend changing nothing but the API client logic, so that we're using `metadata_id` as the parameter name instead of `iaid`. - Everything else remains the same. 

In this case, the `Record.api.fetch(iaid=iaid)` where the keyword input param `iaid` is confusing  whereas the API actually uses `metadataId` internally. 

I had initially thought about removing the keyword input param when calling `api.fetch` so it will become `Record.api.fetch(iaid)`, but then I wanted it to be explicit `Record.api.fetch(metadata_id=iaid)` as we know what `metadata_id` is about as mentioned in the parameter comments.  I think something like a object for multiple formats.

If or when we know that a record does not hold `iaid` we can address it then. 

This are my thoughts @ababic

TNA-Allan: Thanks @ababic updated with minimal api changes.

ababic (approved):

ababic:
Good work @TNA-Allan :) Approved!

PR #673, opened 2023-02-24
UN-491: New label updates

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-491

## How to check these changes

Enter the admin panel, and mark the `Display a new label on the card of this page` under the promote tab of a articles/stories page. Then publish this, and make sure that the "NEW" label is showing on the articles/stories index page, and on the cards at the bottom of articles/stories pages.

All existing articles/stories should be `False` and have no date, all new articles/stories should be marked as `True` by default.

It should also set to `False` on publishing, if `True`. 

Can also be viewed here:
https://feature-un-491-new-qhrif3y-rasrzs7pi6sd4.uk-1.platformsh.site/stories/

## Merging a release branch?

## Deployment guidance


etna/articles/models.py
```diff
@@ -84,12 +86,21 @@ class TaggedArticle(ItemBase):
...
 
     The ArticlePage model.
     """
 
+    def with_content_json(self, content):
+        obj = super().with_content_json(content)
+        if self.live and obj.mark_new_on_next_publish:
+            obj.mark_new_on_next_publish = False
+            obj.newly_published_at = datetime.now().date()
+        return obj
```
ababic, line 102: I think we can simplify this slightly and always set `obj.newly_published_at` when the `obj.mark_new_on_next_publish` is `True`.

With regards to setting `obj.mark_new_on_next_publish` to `False `, if that is done here, the value wouldn't be preserved accross multiple draft version. Instead, we should do this elsewhere, when we know the 'publish' action is happening. Can you think of a good place for that?

ababic: Also, this behaviour should be implemented on the mixin, so that we only need to maintain a single version

etna/articles/models.py
```diff
@@ -43,7 +43,9 @@ class ArticleIndexPage(BasePageWithIntro):
         use_json_field=True,
     )
 
-    new_label_end_date = datetime.now() - timedelta(days=21)
+    @cached_property
+    def new_label_end_date(self):
+        return datetime.now().date() - timedelta(days=21)
```
ababic, line 48: Can you see any reason not to make this a simple `is_new` property that returns a boolean? That might make things a little easier to make use of in templates.

We should also move this behaviour to the mixin class so that we're adhering to DRY principles.

etna/articles/models.py
```diff
@@ -114,7 +125,9 @@ class ArticlePage(HeroImageMixin, ContentWarningMixin, BasePageWithIntro):
         index.SearchField("article_tag_names"),
     ]
 
-    new_label_end_date = datetime.now() - timedelta(days=21)
+    @cached_property
+    def new_label_end_date(self):
+        return datetime.now().date() - timedelta(days=21)
```
ababic, line 130: On the ticket, I suggested making the number of days an attribute on the mixin class, so that it can be overridden for individual types where needed. What do you think about that?

etna/articles/migrations/0061_alter_articlepage_mark_new_on_next_publish.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="articlepage",
+            name="mark_new_on_next_publish",
+            field=models.BooleanField(
+                default=True,
+                help_text="Mark this as true before publishing, if you want to display the 'new label' for 3 weeks",
+                verbose_name="display a new label on the card of this page",
```
ababic, line 19: Maybe a simple "Mark this page as 'new' when published" would make for a clearer label? 

The help text feels a bit overly wordy too. Perhaps we should just confirm where the indicator appears there, and leave the number of days up to whatever has been configured?

etna/articles/models.py
jamesbiggs: > With regards to setting `obj.mark_new_on_next_publish` to `False `, if that is done here, the value wouldn't be preserved accross multiple draft version. Instead, we should do this elsewhere, when we know the 'publish' action is happening. Can you think of a good place for that?

Hi Andy, I've made the other changes that you requested but can't think where to put this? I've had a look at https://github.com/wagtail/wagtail/blob/main/wagtail/models/__init__.py#L539 but the only thing I can think of is checking if `has_unpublished_changes`, to not set it back to False - but I tried this and it works for the draft - but then it doesn't work when trying to publish.

ababic: @jamesbiggs Moving everything to a `save()` method override would probably be the simplest solution. You should be able to look at `self.id` and `self.live` to figure out what you need.

etna/core/models/mixins.py
```diff
@@ -25,3 +29,47 @@ class ContentWarningMixin(models.Model):
...
+        help_text="Mark this as true before publishing, to display the 'new label'",
+    )
+
+    newly_published_at = models.DateField(
+        editable=False,
+        default=None,
+        null=True,
+    )
+
+    new_label_end_date = 21
```
ababic, line 49: Maybe something like `new_label_display_for_days` would be a better fit here?

etna/core/models/mixins.py
```diff
@@ -25,3 +29,47 @@ class ContentWarningMixin(models.Model):
...
         abstract = True
+
+
+class NewLabelMixin(models.Model):
+    """Mixin to allow editors to toggle 'new' label to be applied on-publish"""
+
+    mark_new_on_next_publish = models.BooleanField(
+        verbose_name="mark this page as 'new' when published",
+        default=True,
+        help_text="Mark this as true before publishing, to display the 'new label'",
```
ababic, line 40: This help text feels a bit redundant now, and we're using "Mark" to mean 'check this box' and 'mark the page', which could be confusing. Maybe just ditch the help text and add something later if editors don't think it's clear enough?

etna/core/models/mixins.py
```diff
@@ -25,3 +29,47 @@ class ContentWarningMixin(models.Model):
...
+
+    def with_content_json(self, content):
+        obj = super().with_content_json(content)
+        if obj.mark_new_on_next_publish:
+            obj.mark_new_on_next_publish = False
+            obj.newly_published_at = datetime.now().date()
+        return obj
+
+    @cached_property
+    def is_new(self):
```
ababic, line 59: I know I suggested this in the first place, but maybe `is_newly_published` would be a better fit? `is_new` sounds a little bit too generic (could potentially clash with something internal to Django/Wagtail somewhere), and isn't as descriptive as it could be.

etna/core/models/mixins.py
```diff
@@ -25,3 +29,47 @@ class ContentWarningMixin(models.Model):
...
+        expiry_date = datetime.now().date() - timedelta(days=self.new_label_end_date)
```
ababic, line 60: We should really import `django.utils.timezone` here in this module and update this to:
```suggestion
        expiry_date = timezone.now().date() - timedelta(days=self.new_label_display_for_days)
```

etna/articles/models.py
jamesbiggs: Hi @ababic I believe my latest changes should cover all the conversations :)

etna/core/models/mixins.py
```diff
@@ -25,3 +30,47 @@ class ContentWarningMixin(models.Model):
...
+            if self.newly_published_at > expiry_date:
+                return True
+        return False
+
+    promote_panels = [
+        MultiFieldPanel(
+            [
+                FieldPanel("mark_new_on_next_publish"),
+            ]
+        ),
```
ababic, line 72: @jamesbiggs Is the `MultiFieldPanel` necessary here? I would normally only expect to see them used to group multiple fields, and a `heading="Some heading"` value to group them?

etna/core/models/mixins.py
```diff
@@ -25,3 +30,47 @@ class ContentWarningMixin(models.Model):
...
+        null=True,
+    )
+
+    new_label_display_for_days = 21
+
+    def save(self, *args, **kwargs):
+        if self.live and self.mark_new_on_next_publish:
+            self.mark_new_on_next_publish = False
+            self.newly_published_at = timezone.now().date()
+        return super().save(*args, **kwargs)
```
ababic, line 55: @jamesbiggs I think all we're missing now is the bit that sets `mark_new_on_next_publish` to `False` in the `latest_revision.content` after saving. Without this, that checkbox will remain checked the next time the page is edited (Wagtail always loads the latest revision content into the page editor). We want that checkbox to be unchecked after the 'marking as new' has happened - then, for editors to have to deliberately find that box and check it again if they need to reset the timer.

jamesbiggs: Hi Andy, sorry it took me a while to get back to you - I had to sort out some issues when pulling the latest develop in with the migrations etc.

Is this not what we wanted? Currently, the editor can save the page as draft with the box ticked - this will then remain ticked, and the date won't be set.

When the page is published, the date is set, and the box goes back to the un-ticked state and stays that way.

From what I'm understanding, the `latest_revision.content` suggestion would make it so that any time an editor _opens_ the edit page, it'll set the box to be un-ticked?

ababic: @jamesbiggs The journey you describe is correct, and we're catering for that. However, you're just missing one crucial point, and that is: When you click to edit a page in Wagtail, it's not the saved state of the page you see loaded into the editor.... it's the `latest_revision` content. So, if `mark_new_on_next_publish` value remains True there, the checkbox will also be checked when you next come to edit the page.

William-Blackie (approved):
Hey James, just a couple of little tweaks but I'll approve this. 

thanks - Will.

templates/articles/article_index_page.html
```diff
@@ -40,15 +40,15 @@ <h2 class="sr-only">{% trans "Each collection insight takes an in-depth look int
...
+                    {% if page.featured_article.is_newly_published %}
                         <span aria-describedby="article-desc" role="link">NEW</span>
                     {% endif %}
                 </h2>
 
                 <p id ="article-desc" class="aria-desc">This is a link to a new story about {{ page.featured_article.title }}</p>
 
                 <p>{{ page.featured_article.teaser_text }}</p>
-                <a class="tna-button--dark" href="{% pageurl page.featured_article %}" data-component-name="Featured Article: {{ page.featured_article.title }}" data-link-type="Button" data-link="Read" {% if page.featured_article.first_published_at.date > page.new_label_end_date.date %}data-label = "New"{%endif%}>Read</a>
+                <a class="tna-button--dark" href="{% pageurl page.featured_article %}" data-component-name="Featured Article: {{ page.featured_article.title }}" data-link-type="Button" data-link="Read" {% if page.featured_article.is_newly_published %}data-label = "New"{%endif%}>Read</a>
```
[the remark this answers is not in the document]
jamesbiggs: Hi @William-Blackie do you mean what the purpose of the `data-label` is? Or the "New" label in general?

I think the data-label is for Google Analytics - The "New" label itself is to have a little green icon on the articles cards that says "new" on it

PR #693, opened 2023-03-02
UN-481: Upgrade to Django 4.1.7

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-481

## About these changes

Upgraded Django to 4.1.7 in Poetry.

## How to check these changes

Rebuild containers etc and check over the site to ensure that upgrade is working. Run the tests, test the core features of the site.

## Merging a release branch?

## Deployment guidance


ababic:
As outlined here in the release notes: https://docs.djangoproject.com/en/4.1/releases/4.1/#templates

Django will now use the [cached template loader](https://docs.djangoproject.com/en/4.1/ref/templates/api/#django.template.loaders.cached.Loader) by default when `DEBUG = True` (which is always the case for local development). It would be good to explore how this impacts development, e.g. Are custom template changes still reflected when you restart `python manage.py runserver`, or is more drastic action required?

ababic:
@jamesbiggs The yellow deprecation warnings that appear when the Django app is started are also version related, so it would make sense to also address those here as part of this PR. That could be a combination of:
1. Upgrading the relevant third-party package and seeing if it resolves itself (preferred)
2. Updating Django settings to ignore warnings that we are sure we no longer need to worry about.

jamesbiggs:
Hi @ababic, thanks for this - I had a look through the documentation and couldn't find anything major, so thank you for pointing this out. I'll investigate after I've resolved the issue on main :)

jamesbiggs:
> It would be good to explore how this impacts development, e.g. Are custom template changes still reflected when you restart `python manage.py runserver`, or is more drastic action required?

@ababic is this referring to just HTML templates, such as `article_page.html` or something deeper in Django? If HTML - it seems like it's working okay on my local? Might need some help with this though 🙂 

I had a look at the links you sent and tried adding the `loaders` thing here: https://docs.djangoproject.com/en/4.1/ref/templates/api/#django.template.loaders.cached.Loader

But this came back with this error: `django.core.exceptions.ImproperlyConfigured: app_dirs must not be set when loaders is defined.`

ababic:
@jamesbiggs In answer to your questions...

> is this referring to just HTML templates, such as article_page.html or something deeper in Django? If HTML - it seems like it's working okay on my local? Might need some help with this though 🙂

My worry is that the caching will make it difficult to make updates to template files during development, and seeing those changes reflected on the pages etc when you refresh the browser. Could you maybe try updating some include/block templates while running the Django server, and check that you can see template changes when you save them?

> I had a look at the links you sent and tried adding the loaders thing, but this came back with this error: django.core.exceptions.ImproperlyConfigured: app_dirs must not be set when loaders is defined.

This part of the Django config is a bit hairy... most people use the 'Django default', where the `app_dirs` option is used to automatically populate the `loaders` option. We only really need to change anything should the new default 'cached loader' make it difficult to develop locally and see changes. In that case loaders should be set to a tuple: `("django.template.loaders.filesystem.Loader", "django.template.loaders.app_directories.Loader")`, and the `app_dirs` option removed.

jamesbiggs:
> @jamesbiggs In answer to your questions...
> 
> > is this referring to just HTML templates, such as article_page.html or something deeper in Django? If HTML - it seems like it's working okay on my local? Might need some help with this though 🙂
> 
> My worry is that the caching will make it difficult to make updates to template files during development, and seeing those changes reflected on the pages etc when you refresh the browser. Could you maybe try updating some include/block templates while running the Django server, and check that you can see template changes when you save them?
> 
> > I had a look at the links you sent and tried adding the loaders thing, but this came back with this error: django.core.exceptions.ImproperlyConfigured: app_dirs must not be set when loaders is defined.
> 
> This part of the Django config is a bit hairy... most people use the 'Django default', where the `app_dirs` option is used to automatically populate the `loaders` option. We only really need to change anything should the new default 'cached loader' make it difficult to develop locally and see changes. In that case loaders should be set to a tuple: `("django.template.loaders.filesystem.Loader", "django.template.loaders.app_directories.Loader")`, and the `app_dirs` option removed.

I've updated base.py to not use `APP_DIRS` anymore, and added the loaders 😄 Tested on my local machine and templates seem to change when edited and refreshed etc.

I also don't get any yellow warnings when running commands - I think this is done?

ababic:
Hi @jamesbiggs

> I've updated base.py to not use APP_DIRS anymore, and added the loaders 😄 Tested on my local machine and templates seem to change when edited and refreshed etc.

I think you might have missed my point a little. If the new defaults are okay, there's no need to update the template settings. We should only deviate from the Django defaults IF they aren't suitable. If the you can save changes to include/block templates while running the Django server and see the changes reflected in the browser, the defaults are fine.

> I also don't get any yellow warnings when running commands - I think this is done?

Adding `DEFAULT_AUTO_FIELD = "django.db.models.AutoField"`` to Django settings should be a last resort really. Where those warnings correspond to models in third-party apps, the more appropriate fix is to update the app (because app maintainers usually have their own preference of what that default should be, which they add to `apps.py`)

jamesbiggs:
Cheers @ababic 🙂 

>We should only deviate from the Django defaults IF they aren't suitable. If you can save changes to include/block templates while running the Django server and see the changes reflected in the browser, the defaults are fine.

Okay, gotcha 👍 

> Adding `DEFAULT_AUTO_FIELD = "django.db.models.AutoField"` to Django settings should be a last resort really. Where those warnings correspond to models in third-party apps, the more appropriate fix is to update those app to a more recent version and see if that resolves it (app maintainers usually have their own preference of what that default should be, which they add to `apps.py` or specify on the models themselves)

I've updated `django-birdbath` and `django-allauth`, but the last two warnings that are coming up are related to something we've made internally (I believe) 

`core.MainMenuItem: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.`

and

`core.SiteSettings: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.`

Is it okay for me to set the `DEFAULT_AUTO_FIELD` here if I create an `apps.py` file? If so, should it be `AutoField` or `BigAutoField`?

ababic:
> Is it okay for me to set the DEFAULT_AUTO_FIELD here if I create an apps.py file? If so, should it be AutoField or BigAutoField?

Yep, I think adding an `apps.py` to the core app and setting `default_auto_field = "django.db.models.AutoField"` like we do elsewhere would be ideal.

PR #712, opened 2023-03-09
Fix: Incorrectly formatted revision data for ArticlePages

ababic:


jamesbiggs (approved):
All good, cheers Andy! 😎

PR #730, opened 2023-03-16
DF-596: Prevent stripping of HTML from record descriptions in detail view

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/DF-596

## About these changes

We rightly strip HTML from record descriptions in listings, where links and other HTML entities cause unnecessary complication for users. However, in certain circumstances, we want to display the full value as retrieved from the API, with any links etc left in-tact.

This PR changes the `description` property available for records, so that it no longer strips HTML. A new `listing_description` property is added to make the value available with the same treatment.

## How to check these changes

As per the example in the ticket, head to http://0.0.0.0:8000/catalogue/id/C18099/ locally and look out for the 'Description' value there. It should now include the link from the original value.

However, when you see the description in search results, it'll still be the safe/text-only version:
http://0.0.0.0:8000/search/catalogue/?q=www.firekills.gov.uk&group=tna

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/records/models.py
+++ b/etna/records/models.py
@@ -227,7 +227,7 @@ class Record(DataLayerMixin, APIModel):
         """
         Returns the records full description value with all HTML left intact.
         """
-        return mark_safe(self._get_raw_description(use_highlights=False))
+        return mark_safe(self._get_raw_description())
 
     @cached_property
     def listing_description(self) -> str:
@@ -236,7 +236,7 @@ class Record(DataLayerMixin, APIModel):
         anywhere. When highlight data is provided by the API, <mark> tags
         will be left in-tact, but and other HTML is stripped.
         """
-        if raw := self._get_raw_description(use_highlights=False):
+        if raw := self._get_raw_description(use_highlights=True):
             return mark_safe(strip_html(raw, preserve_marks=True))
         return ""
 
--- a/etna/records/models.py
+++ b/etna/records/models.py
@@ -224,15 +224,28 @@ class Record(DataLayerMixin, APIModel):
 
     @cached_property
     def description(self) -> str:
-        if raw := self._get_raw_description():
+        """
+        Returns the records full description value with all HTML left intact.
+        """
+        return mark_safe(self._get_raw_description(use_highlights=False))
+
+    @cached_property
+    def listing_description(self) -> str:
+        """
+        Returns a version of the record's description that is safe to use
+        anywhere. When highlight data is provided by the API, <mark> tags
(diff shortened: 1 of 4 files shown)


TNA-Allan (changes requested):

etna/records/models.py
```diff
@@ -224,15 +224,28 @@ def catalogue_source(self) -> str:
...
+        return mark_safe(self._get_raw_description(use_highlights=False))
+
+    @cached_property
+    def listing_description(self) -> str:
+        """
+        Returns a version of the record's description that is safe to use
+        anywhere. When highlight data is provided by the API, <mark> tags
+        will be left in-tact, but and other HTML is stripped.
+        """
+        if raw := self._get_raw_description(use_highlights=False):
```
TNA-Allan, line 239:
```suggestion
    @cached_property
    def highlight_description(self) -> str:
        """
        Returns a version of the record's description that is safe to use
        anywhere. When highlight data is provided by the API, <mark> tags
        will be left in-tact, but and other HTML is stripped.
        """
        if raw := self._get_raw_description(use_highlights=True):
```
Field name with `highlight_` in it may serve the purpose of using description with highlight. Thoughts?
The `use_highlights` param requires to be enabled to True in order to use it in `_get_raw_description`

etna/records/models.py
```diff
@@ -224,15 +224,28 @@ def catalogue_source(self) -> str:
...
             return mark_safe(strip_html(raw, preserve_marks=True))
         return ""
 
-    def _get_raw_description(self) -> str:
-        try:
-            return "... ".join(self.highlights["@template.details.description"])
-        except KeyError:
-            pass
+    def _get_raw_description(self, use_highlights: bool = True) -> str:
+        if use_highlights:
```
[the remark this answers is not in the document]
ababic: Happy for highlight use to be opt-in rather than opt-out... as it does have a significant impact on the return value. Fixed in [eef6f6f](https://github.com/nationalarchives/ds-wagtail/pull/730/commits/eef6f6f861fd14f458e1d5457c6fa123e450a4ac)

etna/records/models.py
ababic: The purpose of `listing_description` is a bit more nuanced than just the use of highlights (the stripping of non-highlight HTML, for example). It can also be used in listings where highlights aren't even present, so `highlight_description` in that context would be misleading.

TNA-Allan (approved):
Looks good. Thanks @ababic

PR #733, opened 2023-03-16
UN-391: Topic and Time Period page updates

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-391

## About these changes

Added new properties to Topic + Time period explorer pages to retrieve: related highlight galleries, related articles, and related record articles.

These are all retrieved from the first selected (most relevant) topic/time period.

## How to check these changes

There's no FE for this yet, but check that the correct pages are being returned by each function.

You will need to create Highlight Galleries/Articles/Record Articles and apply Topics/Time periods to them.

## Merging PR guidance

## Deployment guidance


etna/collections/models.py
```diff
@@ -145,12 +150,38 @@ def related_articles(self):
...
+        return (
+            RecordArticlePage.objects.live()
+            .public()
+            .filter(id__in=self.related_page_ids)
+            .order_by("-first_published_at")
+            .exclude(title=self.featured_record_article)
+        )
+
+    @cached_property
+    def related_highlight_galleries(self):
```
William-Blackie, line 173: Note: let's use the naming convention of `related_highlight_gallery_pages`, we're getting the highlight gallery pages, not specifically the gallery, it's semantics but I think it's clearer, what do you think?

etna/collections/models.py
```diff
@@ -145,12 +150,38 @@ def related_articles(self):
...
+        return (
+            HighlightGalleryPage.objects.live()
+            .public()
+            .filter(id__in=self.related_page_ids)
```
William-Blackie, line 177: Suggestion: I think we need to add the select related here for the `teaser_image`?

etna/collections/models.py
```diff
@@ -249,12 +284,38 @@ def related_articles(self):
...
+    @cached_property
+    def related_record_articles(self):
+        from etna.articles.models import RecordArticlePage
+
+        return (
+            RecordArticlePage.objects.live()
+            .public()
+            .filter(id__in=self.related_page_ids)
+            .order_by("-first_published_at")
+            .exclude(title=self.featured_record_article)
```
William-Blackie, line 303: Suggestion: In terms of queries, it's important the order these are made for efficiency. It's better to move the excludes further up the query.

In this query, you're filtering all availble pages by a certain id's, then ordering them, and then excluding. It would be more effecient to exclude first, then filter and etc.

etna/collections/models.py
```diff
@@ -145,12 +150,38 @@ def related_articles(self):
...
```
William-Blackie, line 169: Thought: Is this right? if the featured_record_article is `None` (as it can be left blank), then you're filtering a QS via a reference (PK). I think you'll find this produces an unintended results. You're better off with `exclude(pk=self.featured_record_article)`.

Have a look at this SO post here; https://stackoverflow.com/questions/2165865/django-queries-id-vs-pk it'll explain a little more about the difference between PK V.S. ID. I would suggest using PK going forward.

William-Blackie (commented):
Hey James, good work here. I've left some comments / improvements. They cross over most of the work so please apply the thoughts to the rest of this PR. :)

William-Blackie:
PS. If you could also add some FE boiler plate to iterate out these, that's super helpful to an FE developer (who's not nessecerily full stack). It also easier as you will have an understanding of the current BE representation which without having to dig through BE models any other developer will not have, it's not a must do but you'll keep the questions of "how does this work" to a minimum :smile:

William-Blackie:
Nice work @jamesbiggs, looks good to me! :rocket:

William-Blackie (approved):

PR #736, opened 2023-03-17
UN-494: Show character count for `teaser_text`

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/UN-494

## About these changes

Adds a Draftail-like character count to any plain text input with a `maxlength` attribute (as long as it's present when the form is initially rendered).

This means that behaviour will be a little inconsistent between inputs present in newly added blocks VS blocks that were present on load... But, I can't find a nice way to cater for those. Let's just see how we get on with this lightweight version and we can complicate things further later if necessary?

I have also snuck some style overrides to help the legibility of Draftail character counts when help text is long enough to clash.

## How to check these changes

Edit any page in the Wagtail admin and open the **Promote** tab. You should see a character count when editing the following fields:
- Slug
- Title tag
- Teaser text

NOTE: The position is a little different when compared with the Draftail version - but, I couldn't find a graceful way to solve that.

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/core/static/admin/js/inputLengthIndicators.js
+++ b/etna/core/static/admin/js/inputLengthIndicators.js
@@ -40,7 +40,8 @@ const initializeLengthIndicator = function (input) {
     // Add hidden HTML element to display the count
     const lengthIndicator = document.createElement("div");
     lengthIndicator.className = lengthIndicatorClassname;
-    lengthIndicator.style.display = "none";
+    lengthIndicator.classList.add("w-help-text");
+    lengthIndicator.style.visibility = "hidden";
     input.after(lengthIndicator);
 
     // Add initial value to indicator
@@ -48,11 +49,11 @@ const initializeLengthIndicator = function (input) {
     updateLengthIndicator(lengthIndicator, charCount, maxChars);
 
     input.onfocus = function () {
-        lengthIndicator.style.display = "block";
+        lengthIndicator.style.visibility = "visible";
     };
 
     input.onblur = function () {
-        lengthIndicator.style.display = "none";
+        lengthIndicator.style.visibility = "hidden";
     };
 
     input.oninput = function () {
--- a/etna/core/static/admin/js/inputLengthIndicators.js
+++ b/etna/core/static/admin/js/inputLengthIndicators.js
@@ -22,16 +22,11 @@ const countChars = function (text) {
 const updateLengthIndicator = function (
     lengthIndicator,
     length,
-    maxChars = null,
+    maxChars,
 ) {
-    if (maxChars === null) {
-        lengthIndicator.innerHTML = countHTMLBase + `${length}`;
-    }
-    else {
-        lengthIndicator.innerHTML = countHTMLBase + `${length}/${maxChars}`;
-    }
+    lengthIndicator.innerHTML = countHTMLBase + `${length}/${maxChars}`;
 
-    if (maxChars !== null && length > maxChars) {
+    if (length > maxChars) {
         lengthIndicator.classList.add(lengthIndicatorExceededClassname);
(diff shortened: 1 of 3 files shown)


etna/core/static/admin/js/inputLengthIndicators.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const updateLengthIndicator = function (
+    lengthIndicator,
+    length,
+    maxChars,
+) {
+    lengthIndicator.innerHTML = countHTMLBase + `${length}/${maxChars}`;
+
+    if (length > maxChars) {
```
William-Blackie, line 28: Thought: As we're using max-length as selector, will this max-limit ever be exceded? It's enforced client side from what I can see.

etna/core/static/admin/js/inputLengthIndicators.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+};
+
+const initializeLengthIndicator = function (input) {
+    // Look for relevant attributes on the input
+    const maxChars = input.getAttribute("maxlength");
+
+    // Add hidden HTML element to display the count
+    const lengthIndicator = document.createElement("div");
+    lengthIndicator.className = lengthIndicatorClassname;
```
William-Blackie, line 41: Suggestion: `classList.add` allows for multiple classes so you can shorten this to `classList.add(lengthIndicatorClassname, "w-help-text")`

etna/core/static/css/wagtail-overrides.css
```diff
@@ -2,13 +2,45 @@
...
 
+/* Add padding and background colour to Draftail character counts to
+aid legibility when appearing over help text */
+
+.inputlengthindicator {
+  display: block;
+  background-color: #fff;
+  position: absolute;
+  bottom: -1.6rem;
+  right: 0;
```
William-Blackie, line 13: I see your point about trying to make this ledgible, on mobile it's a displaying on-top of the help-text. Without some real reaching I don't think we can do too much about this. It's still ledgible but not an accessible, but then again an the original implementation I don't think this is solved either as it relies on not having help-text.

We could add a media query to the help-text to force the text to wrap at say 80% of the div size on mobile and 90% on desktop. I'll leave this to you to decide. 

TLDR; I think this is fine as is.

William-Blackie (approved):
Hey @ababic, nice work here. I've left some thoughts but nothing blocking IMO.

Thanks - Will.

etna/core/static/admin/js/inputLengthIndicators.js
ababic: You're right, that's unlikely. This is actually a remnant of when I was supporting a `max-words` option, which doesn't benefit from any client-side enforcement. Saying that, we have had some situations where we have changed the `max_length` value, and existing values might need to be trimmed down to bring them into line - in which case, this may still be of some benefit.

etna/core/static/admin/js/inputLengthIndicators.js
ababic: Something new learned :) Implemented in [be92ad4](https://github.com/nationalarchives/ds-wagtail/pull/736/commits/be92ad4b57dacab6b70a2fe62d537bbdca8069a3). Cheers Will!

etna/core/static/css/wagtail-overrides.css
ababic: Yes, It's a tricky one. I did initially use a `float: right` instead of absolute positioning, which helped somewhat, but felt a little more fragile. I'm probably just going to leave this as it is, as I want to avoid overriding default Wagtail stuff as much as possible.

PR #737, opened 2023-03-17
UN-498: Made adjustments to `models` and the original migration

jamesbiggs:
Quick change for some values on the previous UN-498 PR!


ababic (approved):
LGTM!

PR #744, opened 2023-03-22
UN-507: Content warning on Highlight Gallery and Record Article

jamesbiggs:
## **NOTE FOR REVIEWER: Please don't merge this, just review! It will have migration clashes with my other PR https://github.com/nationalarchives/ds-wagtail/pull/749 which I will resolve when one or the other is merged**

Ticket URL: https://national-archives.atlassian.net/browse/UN-507

## About these changes

Added Content Warning to Highlight Gallery (Record Article was already done)

## How to check these changes

Check on Highlight Gallery and Record Article pages to ensure that the Content Warning Mixin has been applied

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly before handing over to reviewer.
- [x] Checked PR title starts with ticket number as per project conventions to help us keep track of changes.
- [x] Ensured that PR includes only commits relevant to the ticket.
- [x] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

## Merging PR guidance

Follow [docs\developer-guide\contributing.md](https://nationalarchives.github.io/ds-wagtail/developer-guide/contributing/)

## Deployment guidance

Follow [docs\infra\environments.md](https://nationalarchives.github.io/ds-wagtail/infra/environments/)

PR #749, opened 2023-03-23
UN-513: Remove long description from Highlight Gallery

jamesbiggs:
## **NOTE FOR REVIEWER: Please don't merge this, just review! It will have migration clashes with my other PR https://github.com/nationalarchives/ds-wagtail/pull/744 which I will resolve when one or the other is merged**

Ticket URL: https://national-archives.atlassian.net/browse/UN-513

## About these changes

Removed `long description` from the Highlight model which is used in the Highlight Gallery page. Now using the `description` field which is on the `Image` model instead.

## How to check these changes

Check a Highlight Gallery page and ensure that it is using the `description` field and there is no `long description` field anymore.

## Merging PR guidance

## Deployment guidance


etna/collections/models.py
```diff
@@ -543,15 +542,9 @@ class Highlight(Orderable):
...
-        features=settings.RESTRICTED_RICH_TEXT_FEATURES,
-        max_length=400,
-    )
 
     panels = [
         FieldPanel("image"),
-        FieldPanel("long_description"),
     ]
 
     def clean(self) -> None:
```
William-Blackie, line 550: Hey @jamesbiggs,

this works locally for me. I would suggest that we do something similar here in terms of checking that a record has an image, but we check that the record has a description. The design is dependent on it.

William-Blackie (commented):
Nice work @jamesbiggs, Just one suggestion for a more consistent editor experience.

etna/collections/models.py
```diff
@@ -543,26 +542,26 @@ class Highlight(Orderable):
...
+            error_fields = []
+            if self.image.record is None:
+                error_fields.append("'record'")
+            if self.image.description == "":
+                error_fields.append("'description'")
+            if error_fields:
+                raise ValidationError(
+                    {
+                        "image": [
+                            f"Only images with a {' and '.join(error_fields)} specified can be used for highlights."
```
William-Blackie, line 561: Hey @jamesbiggs,

I can see what you're doing here and why you're doing it. It's smart, but it doesn't have to be (although this would be very applicable to other parts of the project going forward). As _both_ variables are requied to be used as a highlight it would be more user friendly to display the full conditions i.e. "Only images with a record and description can be used as a highlight".

William-Blackie: I would think something along the lines of 

```
if self.image:
    if not self.image.record or not self.image.description:
         # Raise an error 

```

Just a small prod, you can you use a python interpreter to quickly check your logic when doing this type of stuff; 

Python 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> string = ""
>>> bool(string)
False

tldr; "" is evaluated as `False` while any character is evaluated as `True`

William-Blackie (approved):
Hey @jamesbiggs, 

Good work here, I've left a couple of suggestions for how to make this simipler, I'll approve this as they're non-blocking but will make the logic clearer

PR #755, opened 2023-03-27
Un 512 related record article pages

William-Blackie:
Ticket URL: [here](https://national-archives.atlassian.net/jira/software/c/projects/UN/boards/75?modal=detail&selectedIssue=UN-512)

This adds the related record article pages:
- 
-
-

0 0 0 0_8000_admin_pages_37_edit_preview__mode=

0 0 0 0_8000_pattern-library_render-pattern_patterns_pages_donate_donate_form html
0 0 0 0_8000_explore-the-collection_explore-by-time-period_medieval_an-example-gallery-page_ (1)

## How to check these changes

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/collections/models.py
+++ b/etna/collections/models.py
@@ -542,15 +542,9 @@ class Highlight(Orderable):
         on_delete=models.SET_NULL,
         verbose_name=_("image"),
     )
-    long_description = RichTextField(
-        verbose_name=_("long description"),
-        features=settings.RESTRICTED_RICH_TEXT_FEATURES,
-        max_length=400,
-    )
 
     panels = [
         FieldPanel("image"),
-        FieldPanel("long_description"),
     ]
 
     def clean(self) -> None:
--- a/etna/collections/models.py
+++ b/etna/collections/models.py
@@ -260,6 +260,7 @@ class TimePeriodExplorerPage(AlertMixin, BasePageWithIntro):
         FieldPanel("body"),
         FieldPanel("start_year"),
         FieldPanel("end_year"),
+        InlinePanel("page_time_periods", label="Time Period Pages")
     ]
 
     settings_panels = BasePage.settings_panels + AlertMixin.settings_panels
@@ -288,11 +289,9 @@ class TimePeriodExplorerPage(AlertMixin, BasePageWithIntro):
 
         return (
             ArticlePage.objects.exclude(pk=self.featured_article)
-            .live()
-            .public()
-            .filter(pk__in=self.related_page_pks)
+            .exclude(teaser_image=None)
             .order_by("-first_published_at")
-            .select_related("teaser_image")
+            .select_related("teaser_image")[:4]
         )
 
     @cached_property
--- a/sass/includes/_card-grid.scss
+++ b/sass/includes/_card-grid.scss
@@ -1,12 +1,22 @@
 .card-grid {
     @extend .row;
     display: grid;
-    grid-gap: 1rem;
-    margin: 1rem 0;
+    grid-gap: 40px;
+    margin: 20px 0;
     
     &--duo {
         @media (min-width: map-get($grid-breakpoints, "md")) {
             grid-template-columns: repeat(2, 1fr);
+        }
+    }
+
+    &--quad {
+        @media (min-width: map-get($grid-breakpoints, "sm")) {
+            grid-template-columns: repeat(2, 1fr);
+        }
+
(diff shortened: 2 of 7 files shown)


etna/collections/models.py
```diff
@@ -260,6 +260,7 @@ class TimePeriodExplorerPage(AlertMixin, BasePageWithIntro):
         FieldPanel("body"),
         FieldPanel("start_year"),
         FieldPanel("end_year"),
+        InlinePanel("page_time_periods", label="Time Period Pages")
```
William-Blackie, line 263: Note: @ababic / @jamesbiggs - I'm not sure if this was intended to be left out?

etna/collections/models.py
```diff
@@ -543,15 +542,9 @@ class Highlight(Orderable):
...
     )
-    long_description = RichTextField(
-        verbose_name=_("long description"),
-        features=settings.RESTRICTED_RICH_TEXT_FEATURES,
-        max_length=400,
-    )
 
     panels = [
         FieldPanel("image"),
-        FieldPanel("long_description"),
```
William-Blackie, line 554: I'm not sure what the deal is with this being here, but I can't save the page model to get the relations without this being removed.

jamesbiggs: Did you review my branch where I removed it? Maybe some lingering migrations or something

etna/collections/models.py
```diff
@@ -288,11 +289,9 @@ def related_articles(self):
 
         return (
             ArticlePage.objects.exclude(pk=self.featured_article)
-            .live()
-            .public()
-            .filter(pk__in=self.related_page_pks)
+            .exclude(teaser_image=None)
             .order_by("-first_published_at")
-            .select_related("teaser_image")
+            .select_related("teaser_image")[:4]
```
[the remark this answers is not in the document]
jamesbiggs: I agree with the `.exclude(teaser_image = None)`, but why are we removing the `.filter()` for related pages and a check to make sure the page is live + public? Could there not also be more than 4 related article pages?

etna/collections/models.py
jamesbiggs: The `page_time_periods` InlinePanel is used for "tagging" the pages/relating them to a `TimePeriodExplorerPage` (e.g. Medieval, Post-Modern, etc). This page model is what the InlinePanel would be referring to, so I don't think we want it on there because we don't need to "tag" the page

William-Blackie: Right but how do these tags show up in related listing if we aren't building at least a few relationships between them?

jamesbiggs: The `page_time_periods` works by giving the `TimePeriodPage` a reference to the "parent" `TimePeriodExplorerPage`. Then we use the `.filter(pk__in=self.related_page_pks)` to get all the `TimePeriodPage`s that contain a reference to the `TimePeriodExplorerPage`. `page_time_periods` would only allow the `TimePeriodExplorerPage` to reference itself (or another `TimePeriodExplorerPage`) which isn't what we want here

etna/collections/models.py
William-Blackie: I did, but was this removed or just the field? The field does not exist in the DB, I've rebased twice but perhaps this is not in `develop` and I'm using data that's old / modified from that PR.

sass/includes/_card-grid.scss
```diff
@@ -1,15 +1,29 @@
 .card-grid {
     @extend .row;
     display: grid;
-    grid-gap: 1rem;
-    margin: 1rem 0;
```
William-Blackie, line 5: Note: I've generalised this, I assumed more consistency with card-grids but I've moved them to their sepecifc modfiers.

sass/includes/_highlight-cards.scss
```diff
@@ -0,0 +1,76 @@
...
+        padding: 3% 5%;
+
+        @media screen  and (min-width: map-get($grid-breakpoints, "sm")) {
+            padding: 3% 9%;
+            
+        }
+    }
+
+    &__link {
+        text-decoration: none !important;
```
William-Blackie, line 16: Note: I'm really not a fan of using !important, but the current targetting specificity of the `a` in typography means I need to use this or make multiple site wide changes.

JohnHeeryTNA: I agree, as a general rule we want to avoid the use of !important but in certain cases we can use it.

PR #785, opened 2023-04-06
DF-607: Fix JS error on search results page

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/DF-607

## About these changes

The JS is developed in such a way that assumes the list of selected filters is always rendered, so I've refactored slightly so that the JS only runs when everything is present.

There was also some weird/unnecessary wrapping of functions here, which I've tried to simplify. I've also tried to make the remaining function names a bit more obvious.

## How to check these changes

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/scripts/src/modules/analytics/search/search_filters_tracking.js
+++ b/scripts/src/modules/analytics/search/search_filters_tracking.js
@@ -14,7 +14,7 @@ const pushActiveFilterData = (filterList) => {
     // get current search bucket
     const searchBucket = filterList.getAttribute('data-search-bucket');
 
-    const filters = filterList.querySelector("[data-filter]");
+    const filters = filterList.querySelectorAll("[data-filter]");
 
     // create array to store currently active filters
     const activeFilters = [];
--- a/scripts/src/modules/analytics/search/search_filters_tracking.js
+++ b/scripts/src/modules/analytics/search/search_filters_tracking.js
@@ -1,21 +1,24 @@
 import push_to_data_layer from "./../push_to_data_layer";
 
-const getFilters = () => {
+const pushActiveFilterDataOnLoad = () => {
     // get filters after DOM has loaded and they have rendered on page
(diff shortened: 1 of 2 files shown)


scripts/src/modules/analytics/search/search_filters_tracking.js
```diff
@@ -1,21 +1,24 @@
...
 
-const loopFilters = (filters) => {
+const pushActiveFilterData = (filterList) => {
     // get current search bucket
-    const searchBucket = document.querySelector('[data-search-bucket]').getAttribute('data-search-bucket');
-    
+    const searchBucket = filterList.getAttribute('data-search-bucket');
+
+    const filters = filterList.querySelector("[data-filter]");
+
```
William-Blackie, line 18: Note / Suggestion: Depending on how defensive we want to be here. We could check that `searchBucket` / `filters` are not null values. Similar to how you've done above and log an error to the console. 

In theory, this should always have those values but I'll leave that choice to you. As you've pointed out this code is a little awkward anyway. It would be worth adding Sentry to the FE if we can to avoid this issue being overlooked in the future as I assume that's the case.

William-Blackie (approved):
Hey @ababic, good work here. I've left an optional suggestion for you to look at, the sentry suggestion is more of a future todo.

scripts/src/modules/analytics/search/search_filters_tracking.js
ababic: Ahah. I was originally replying to say `querySelector()` will give me an empty list when there are no matches, but I was totally wrong about that. `querySelectorAll()` is what I really want here (there are supposed to be multiple matching elements).

ababic: In regards to adding Sentry... I think that's definitely worthwhile, but probably one for a separate PR.

PR #789, opened 2023-04-06
Un 201 related stories topic and time period pages

William-Blackie:
Ticket URL: [here](https://national-archives.atlassian.net/jira/software/c/projects/UN/boards/75?modal=detail&selectedIssue=UN-511)

## About these changes

Add a human-readable description here

## How to check these changes

Where possible, provide guidance to help your reviewer

## Before assigning to reviewer, please make sure you have

- [ ] Checked things thoroughly before handing over to reviewer.
- [ ] Checked PR title starts with ticket number as per project conventions to help us keep track of changes.
- [ ] Ensured that PR includes only commits relevant to the ticket.
- [ ] Waited for all CI jobs to pass before requesting a review.
- [ ] Added/updated tests and documentation where relevant.

## Merging PR guidance

Follow [docs\developer-guide\contributing.md](https://nationalarchives.github.io/ds-wagtail/developer-guide/contributing/)

## Deployment guidance

Follow [docs\infra\environments.md](https://nationalarchives.github.io/ds-wagtail/infra/environments/)

PR #790, opened 2023-04-06
UN-378: Spacing between Featured Records and Image components

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-378

## About these changes

Added `margin-bottom` to the `featured-records` element in the SASS.

## How to check these changes

Go on a Stories/Article page, add a featured records block to the page.

I would suggest William Cuffey, an example here:
image

Then view the page and ensure there is a gap between the featured-records component, and the content below it

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/sass/includes/_featured-records.scss
+++ b/sass/includes/_featured-records.scss
@@ -1,4 +1,6 @@
 .featured-records {
+  margin-bottom: 4rem;
+
   &__heading {
     color: $color__dark;
     background-color: $color__yellow_2;
--- a/sass/includes/_featured-records.scss
+++ b/sass/includes/_featured-records.scss
@@ -1,5 +1,5 @@
 .featured-records {
-  margin-bottom: 4rem;
+  margin-bottom: 2rem;
 
   &__heading {
     color: $color__dark;


William-Blackie:
Hey @jamesbiggs,

I can see why you've chosen this approach but as this component is used in a lot of places this likely breaks the design else where. I have also indirectly solved this in this PR [here](https://github.com/nationalarchives/ds-wagtail/pull/789/files#diff-ea211768803649fd93065ba607aec82de53fb5684bc3647df2de2cbfb7faa550R59). It's better practice (IMO) to use wrappers here instead of adding padding to the component itself to allow greater flexibility in use of the component.

jamesbiggs:
Actually, sorry, @William-Blackie, this component isn't used anywhere else outside of the articles pages as far as I can see - my CSS/SASS understanding isn't too good but I'm not sure if your PR covers this component?

sass/includes/_featured-records.scss
```diff
@@ -1,4 +1,6 @@
 .featured-records {
+  margin-bottom: 4rem;
```
William-Blackie, line 2: Hey @jamesbiggs,

You're right, it was my mistake, this template is not used where I thought it was. The templates/ styling class names are too similar. I think 4rem here is little too large, could we use 2rem? That's more inkeeping with the rest of the spacing used within the streamfield elements.

William-Blackie (approved):
Sorry @jamesbiggs, you're right my mistake. This is a different component, the naming conventions are so similar!

PR #803, opened 2023-04-13
CHORE: Tweak venv dependency cache keys in GitHub workflows

ababic:
I'm hoping this will reduce the likelihood of bad cache hits, which appear to be causing workflows to fail sporadically.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,7 +70,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: venv-${{ runner.os }}-${{ env.PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: venv-${{ runner.os }}-${{ env.PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}
+          key: ds-wagtail-venv-${{ runner.os }}-${{ env.PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}
```
TNA-Allan, line 73: Hi @andy, this is new to me. What is `ds-wagtail-venv`?

ababic: This is the 'cache key' used to temporarily store the contents of the virtual environment. What I'm trying to do here is ensure the value is unique to this project, so that there's no chance of confusion with other projects.

TNA-Allan (approved):
Thanks @ababic

PR #807, opened 2023-04-14
DF-396: Fix disappearing selected filters when other applied filters make them redundant

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/DF-396

## How to check these changes

Check out the following URL in develop: https://develop-sr3snxi-rasrzs7pi6sd4.uk-1.platformsh.site/search/catalogue/?collection=DEFE&collection=HW&collection=RGO&level=Piece&closure=Open+Document%2C+Open+Description&q=test+search+term&group=tna

Notice how the **Collections** filter options for 'DEFE' and 'RGO' have disappeared, despite being applied?

Now check out the same URL locally: http://localhost:8000/search/catalogue/?collection=DEFE&collection=HW&collection=RGO&level=Piece&closure=Open+Document%2C+Open+Description&q=test+search+term&group=tna

The missing filter options should now appear with with a '(0)' count in brackets to indicate they have no matches when combined with the other active filters.

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly before handing over to reviewer.
- [x] Checked PR title starts with ticket number as per project conventions to help us keep track of changes.
- [x] Ensured that PR includes only commits relevant to the ticket.
- [x] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

## Merging PR guidance

Follow [docs\developer-guide\contributing.md](https://nationalarchives.github.io/ds-wagtail/developer-guide/contributing/)

## Deployment guidance

Follow [docs\infra\environments.md](https://nationalarchives.github.io/ds-wagtail/infra/environments/)

PR #808, opened 2023-04-14
DF-610:country filter record creators

TNA-Allan:
Ticket URL: https://national-archives.atlassian.net/browse/DF-610

## About these changes

Added country filter to Record creators bucket.

## How to check these changes

As mentioned in the ticket.

## Merging PR guidance

## Deployment guidance


etna/search/views.py
```diff
@@ -398,6 +399,30 @@ def get_api_kwargs(self, form: Form) -> Dict[str, Any]:
             sort_order=form.cleaned_data.get("sort_order"),
         )
 
+    def aggregations_params(self) -> Tuple[str]:
+        """
+        The aggregations params may be specific to a bucket and will be filtered upon.
+        There is a max (8) params that can be passed to the api aggregations search param.
+        Returns a list of aggregation params.
+        """
```
ababic, line 407: Instead of adding a new method here, we should update the `Bucket` class to accept this configuration as an optional argument, and pull it from there

ababic (commented):
We should probably update `DynamicMultipleChoiceField` to use these widget and class values by default. It would save us a lot of lines of code.

See: https://github.com/django/django/blob/main/django/forms/fields.py#L84
And: https://github.com/django/django/blob/main/django/forms/fields.py#L219-L225

UPDATE:

I have made this change over in https://github.com/nationalarchives/ds-wagtail/pull/807

TNA-Allan:
> We should probably update `DynamicMultipleChoiceField` to use these widget and class values by default. It would save us a lot of lines of code.
> 
> See: https://github.com/django/django/blob/main/django/forms/fields.py#L84 And: https://github.com/django/django/blob/main/django/forms/fields.py#L219-L225
> 
> UPDATE:
> 
> I have made this change over in #807

Thanks @ababic . The change is brought here.

etna/search/views.py
TNA-Allan: Sure this is better approach. Implemented the same. Thanks.
1) In order to make the bucket attribute available before the API call, it was required to [set_current_bucket](https://github.com/nationalarchives/ds-wagtail/blob/6fc416a81f17375204be3dabc9e27eb01de023dc/etna/search/views.py#L233) when group value was given.
2) There is a follow on ticket that addresses the required aggregations params for each bucket [DF-641](https://national-archives.atlassian.net/browse/DF-641)

etna/ciim/constants.py
```diff
@@ -35,6 +72,7 @@ class Bucket:
     result_count: int = None
     is_current: bool = False
     results: List[Any] = None
+    api_aggregations_params: List[str] = None
```
ababic, line 75: Maybe just `aggregations` would do as an attribute name?

etna/ciim/constants.py
```diff
@@ -27,6 +27,43 @@ class SearchTabs(StrEnum):
...
+    values = []
+    for aggregation in api_aggregations_params:
+        item_count = 10
+        if aggregation == Aggregation.GROUP:
+            # Fetch more 'groups' so that we receive a counts
+            # for any bucket/tab options we might be showing
+            # (not just the 10 most popular)
+            item_count = 30
+        values.append(f"{aggregation}:{item_count}")
+    return values
```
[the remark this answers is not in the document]
ababic: If we want the dataclass to be responsible for returning a normalized list, that can easily be done by adding a `get_normalised_aggregations()` method to `Bucket`

ababic (changes requested):
Hi @TNA-Allan. I have a few more suggestions here for you.

etna/search/views.py
```diff
@@ -403,30 +409,14 @@ def get_api_aggregations(self) -> List[str]:
...
-            values.append(f"{aggregation}:{item_count}")
-        return values
+        The aggregations params may be specific to a bucket and will be filtered upon.
+        Returns a list of aggregation params.
+        Ex: ["group:30", "catalogue:10",]
+        """
+        for bucket in self.bucket_list:
+            if bucket.is_current:
+                return bucket.api_aggregations_params
+        return []
```
ababic, line 419: I feel like knowing the 'current bucket' is common enough that we should add a `current_bucket` cached property to the view to pull this out. I have a feeling it could be useful do be able to do `view.current_bucket` in templates too.

TNA-Allan: Made use of `BucketList.current`

etna/ciim/constants.py
```diff
@@ -36,6 +53,31 @@ class Bucket:
     is_current: bool = False
     results: List[Any] = None
 
+    aggregations: List[str] = field(default_factory=list)
+
+    @property
```
ababic, line 58: @TNA-Allan I don't think there's any need for this to be a property. If you want it to be, The name should probably be changed to something less method-like (e.g. `aggregations_normalised`)

etna/ciim/constants.py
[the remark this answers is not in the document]
TNA-Allan: Thanks @ababic, Updated with `get_normalised_aggregations` felt better control over those numeric values from one place

etna/ciim/constants.py
```diff
@@ -36,6 +53,31 @@ class Bucket:
...
+    def get_normalised_aggregations(self) -> List[str]:
+        """
+        In the API response, the items with the highest number of matches are
+        included for each aggregation. Those values are used to indicate
+        counts for each 'bucket', and to update the form field choices, so that
+        the most relevant filter options are shown.
+        Ex: ["group:30", "catalogue:10",]
+        """
```
ababic, line 66:
```suggestion
        """
        Returns a list of strings to include as the 'aggregations' option value when assembling
        an API request for this bucket. Each string must be in the format `"filter-name:item-count"`.
        """
```

etna/ciim/constants.py
```diff
@@ -73,39 +115,149 @@ def __iter__(self):
...
             label="Online records at The National Archives",
             description="Results for records available to download and held at The National Archives that match your search term.",
+            aggregations=(
+                Aggregation.COLLECTION,
+                Aggregation.LEVEL,
+                Aggregation.TOPIC,
+                Aggregation.CLOSURE,
+                Aggregation.HELD_BY,
+                Aggregation.CATALOGUE_SOURCE,
+                Aggregation.GROUP,
```
ababic, line 140:
```suggestion
                Aggregation.GROUP + ":30",
```

etna/ciim/constants.py
```diff
@@ -36,6 +53,31 @@ class Bucket:
...
+        """
+        values = []
+        for aggregation in self.aggregations:
+            item_count = 10
+            if aggregation == Aggregation.GROUP:
+                # Fetch more 'groups' so that we receive a counts
+                # for any bucket/tab options we might be showing
+                # (not just the 10 most popular)
+                item_count = 30
+            values.append(f"{aggregation}:{item_count}")
```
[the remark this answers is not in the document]
TNA-Allan: @ababic , special handling of `item_count = 30` is suggested for removal, should ":30" (below comments) be a constant, as we are giving it special treatment which is common to all buckets?

etna/ciim/constants.py
```diff
@@ -73,39 +115,149 @@ def __iter__(self):
...
             label="Records at The National Archives",
             description="Results for records held at The National Archives that match your search term.",
+            aggregations=(
+                Aggregation.COLLECTION,
+                Aggregation.LEVEL,
+                Aggregation.TOPIC,
+                Aggregation.CLOSURE,
+                Aggregation.HELD_BY,
+                Aggregation.CATALOGUE_SOURCE,
+                Aggregation.GROUP,
```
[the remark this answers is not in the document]
TNA-Allan: @ababic should ":30" be a constant, as we are giving it special treatment which is common to all buckets?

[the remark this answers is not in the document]
ababic: See answer above @TNA-Allan

etna/ciim/constants.py
TNA-Allan: Done.

etna/ciim/constants.py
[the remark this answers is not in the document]
TNA-Allan: This is Done. Thanks @ababic

etna/ciim/constants.py
```diff
@@ -56,27 +56,22 @@ class Bucket:
...
+            bits = aggregation.split(":")
+            if len(bits) == 2:
+                values.append(bits[0] + ":" + bits[1])
+            else:
+                values.append(bits[0] + ":10")
         return values
 
     def __post_init__(self):
-        self.aggregations = self.get_normalised_aggregations
+        self.aggregations = self.aggregations_normalised
```
ababic: It's nice that you're thinking about the API here, but it isn't important that this value be normalised at this point in time. It's fine for the value to be lazily generated when it is actually required (you'll see this pattern a lot in Django)

etna/ciim/constants.py
```diff
@@ -137,7 +133,8 @@ def __iter__(self):
                 Aggregation.CLOSURE,
                 Aggregation.HELD_BY,
                 Aggregation.CATALOGUE_SOURCE,
-                Aggregation.GROUP,
+                Aggregation.GROUP
+                + ":30",  # Fetch more 'groups' so that we receive counts for any bucket/tab options we might be showing (not just the 10 most popular)
```
ababic, line 137: I don't think this needs an explanation (especially not every time)

TNA-Allan: 1.  What are the thoughts on why it does not need explanation?  It looks important as per the existing code base [here](https://github.com/nationalarchives/ds-wagtail/blob/4f04108367c81a0b41b801be3dfa3031bdd6e11e/etna/search/views.py#L422-L428) which is why when it was moved here, moved the comment as well.
2. When I put it, I thought its an important value (and not a variable) that must be passed, otherwise make it constant (as it was applied to all buckets in previous code) since it is repetitive. But then you mentioned "_I would say let's avoid adding new constants for this. It's only coincidence that we want 30 groups for each bucket. There is no reason why this shouldn't change in future_." What are the thoughts on not repeating repeating it every time?

etna/search/views.py
```diff
@@ -83,21 +92,11 @@ def get_buckets(
             for bucket in bucket_list:
                 bucket.result_count = doc_counts_by_key.get(bucket.key, 0)
 
-        if current_bucket_key:
-            # set 'is_current=True' for the relevant bucket
-            for bucket in bucket_list:
-                if bucket.key == current_bucket_key:
-                    bucket.is_current = True
-                    break
-
```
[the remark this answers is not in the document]
TNA-Allan: To get the current bucket, there is `BucketList.current` have used it over [here ](https://github.com/nationalarchives/ds-wagtail/blob/d71188a56a39682c24a816feaed1edfaab2dd2c4/etna/search/views.py#L416) in this PR 
The reason for removing this see my comment.
> In order to make the bucket attribute available before the API call, it was required to `set_current_bucket` when group value was given.

etna/search/views.py
```diff
@@ -49,26 +50,34 @@ class BucketsMixin:
...
+        """
+        The `current_bucket_key` is provided, any bucket with a `key` value matching
+        the provided value will have it's `is_current` value set to `True`,
+        other buckets are reset `False`.
+        """
+        for bucket in self.bucket_list:
+            bucket.is_current = False
+            if bucket.key == current_bucket_key:
+                bucket.is_current = True
+        return None
```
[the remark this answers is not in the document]
TNA-Allan: We can get the current bucket as implemented [here ](https://github.com/nationalarchives/ds-wagtail/blob/d71188a56a39682c24a816feaed1edfaab2dd2c4/etna/search/views.py#L416)

etna/search/views.py
```diff
@@ -222,11 +221,17 @@ def get(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
...
+        Sets the current bucket when a valid group is given.
         """
         form = self.form = self.get_form()
         is_valid = form.is_valid()
         self.api_result = None
         self.current_bucket_key = form.cleaned_data.get("group")
+
+        if self.current_bucket_key:
+            self.set_current_bucket(current_bucket_key=self.current_bucket_key)
+
```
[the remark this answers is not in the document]
TNA-Allan: To get the current bucket, there is BucketList.current have used it over [here ](https://github.com/nationalarchives/ds-wagtail/blob/d71188a56a39682c24a816feaed1edfaab2dd2c4/etna/search/views.py#L416)in this PR
see my comment

> In order to make the bucket attribute available before the API call, it was required to set_current_bucket when group value was given.

etna/ciim/constants.py
ababic: It's only as harmful as any other non-optimal code... Having this here means more code for you to test and maintain, and the module probably takes a few more microseconds to initialise. It's just a better design to put off 'work' until it needs doing.

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: @ababic  Earlier it seemed confusing to know that `get_buckets()` also sets the current bucket. We need to access `self.bucket_list.current.aggregations` before the earlier setting  `is_current` status in `get_buckets()` which is called via `get_context_data` . Therefore `set_current_bucket` does that before `get_context_data`.

etna/ciim/constants.py
```diff
@@ -53,12 +53,13 @@ class Bucket:
     is_current: bool = False
     results: List[Any] = None
 
+    # Fetch more 'groups' so that we receive counts for any bucket/tab options we might be showing Ex: ["group:30"]
```
ababic, line 56: Sorry @TNA-Allan - this comment does not make sense here. When I suggested updating the dataclass, I meant adding something like: 

# By default, 10 items of each aggregation are requested from the API. This can be overridden by using a string in the format '{name}:{number_of_items}'

etna/ciim/constants.py
TNA-Allan: I have gone with your suggestions. Can you please clarify what you'd like to do over here, or maybe can you take this as separate ticket?

etna/search/views.py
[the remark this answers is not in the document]
ababic: What's the best way forward here, @TNA-Allan?

etna/ciim/constants.py
TNA-Allan: Why don't you want to say about `group:30`, as it seems quite out of the blue using that number without any reason. Thoughts?

ababic: I can tell you're getting frustrated. Can I suggest taking your lunch and looking at this with fresh eyes when you come back. This value-specific comment simply makes no sense on a dataclass.

TNA-Allan: He He .. I'm good @ababic, Its confusing when at one instant we use `:30` and say why, _then when putting the same comment when its called - ( explanation is not needed there)_ we loosing such an important comment which was previously there to know why its `:30` and not any other number. It's a random number - yes, but this number gets the desired buckets counts we are displaying. Is there a way not to loose this comment?

ababic: From my perspective, we aren't losing anything important. The dataclass explains what is expected of instances, and the instances obey the restrictions dataclasses API. If you wanted to understand why the ":30" was there previously, there was nothing to make that clear. But now, you would just look at the dataclass, see the comment, and understand.

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: >  Now we have two methods for preparing that data that need to be called in a specific order. Surely we can find a way to make this work that doesn't add this unnecessary complication?

Yes 
1) set_current_bucket() - set the currents bucket after the form is validated which is available before calling the API and required so that we use `self.bucket_list.current.aggregations` before the API call.
2) get_buckets() - uses data after the API is called which serves other purposes  Ex: update result_count which is available only after API call.

Its following a general approach to way/pattern Bucket attributes are initialised. I don't think its complicated.

etna/ciim/constants.py
TNA-Allan: My thoughts are, while refactoring my be fine, removing an existing commented code which does not belong to the code change for this PR does not feel right, besides someone would know why that number is :30  and not :10 or :15.

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: @ababic I will try using BucketList.get_bucket()

ababic:
Hi @TNA-Allan. I have implemented the `get_current_bucket()` changes in a separate PR, as I think that's where most of the confusion has come from: https://github.com/nationalarchives/ds-wagtail/pull/811

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: > Hi @TNA-Allan. I have implemented the get_current_bucket() changes in a separate PR, as I think that's where most of the confusion has come from: https://github.com/nationalarchives/ds-wagtail/pull/811

Thanks for this, will apply changes from PR once we're ready.

etna/ciim/constants.py
```diff
@@ -36,6 +53,27 @@ class Bucket:
...
+            bits = aggregation.split(":")
+            if len(bits) == 2:
+                values.append(bits[0] + ":" + bits[1])
+            else:
+                values.append(bits[0] + ":10")
+        return values
+
+    def __post_init__(self):
+        self.aggregations = self.aggregations_normalised
+
```
ababic, line 76: Hmm. My feedback on this seems to have been removed. Could we remove this please, and just use 'bucket. aggregations_normalised' in the view?

etna/ciim/constants.py
TNA-Allan: Updated for default aggregations and comments [e174551c](https://github.com/nationalarchives/ds-wagtail/pull/808/commits/e174551cb7db57d32fa5e682625de2911f931675)

etna/ciim/constants.py
TNA-Allan: > Hmm. My feedback on this seems to have been gone missing.
Sorry I did not receive email notification either. However there is one above for `self.aggregations_normalised`

TNA-Allan: I've now updated this method in the commit 
> Updated for default aggregations and comments [e174551c](https://github.com/nationalarchives/ds-wagtail/pull/808/commits/e174551cb7db57d32fa5e682625de2911f931675)

etna/ciim/constants.py
```diff
@@ -69,7 +68,7 @@ class Bucket:
     results: List[Any] = None
 
     # By default, 10 items of each aggregation are requested from the API. This can be overridden by using a string in the format '{name}:{number_of_items}'
-    aggregations: List[str] = field(default_factory=lambda:DEFAULT_AGGREGATIONS)
+    aggregations: List[str] = field(default_factory=lambda: DEFAULT_AGGREGATIONS)
```
ababic, line 71:
```suggestion
    aggregations: List[str] = DEFAULT_AGGREGATIONS
```

etna/ciim/constants.py
```diff
@@ -53,7 +66,7 @@ class Bucket:
     is_current: bool = False
     results: List[Any] = None
 
-    # Fetch more 'groups' so that we receive counts for any bucket/tab options we might be showing Ex: ["group:30"]
+    # By default, 10 items of each aggregation are requested from the API. This can be overridden by using a string in the format '{name}:{number_of_items}'
     aggregations: List[str] = field(default_factory=list)
```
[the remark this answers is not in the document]
TNA-Allan: Yes,  [bf1f8f ](https://github.com/nationalarchives/ds-wagtail/pull/808/commits/bf1f8f9734f1912a429ca9af3cddfb1e4ba6c745)

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: PR [811]( https://github.com/nationalarchives/ds-wagtail/pull/811)  applied

ababic:
Hi @TNA-Allan. Something has gone wrong with this PR... it's reporting changes from other branches. Would you be able to look at rebasing so that the PR only contains commits relevant to this work?

etna/search/tests/test_views.py
```diff
@@ -784,20 +794,50 @@ def test_current_bucket(self):
...
+                    aggregations=(
+                        Aggregation.COLLECTION,
+                        Aggregation.LEVEL,
+                        Aggregation.TOPIC,
+                        Aggregation.CLOSURE,
+                        Aggregation.HELD_BY,
+                        Aggregation.CATALOGUE_SOURCE,
+                        Aggregation.GROUP + ":30",
+                        Aggregation.TYPE,
+                    ),
```
ababic, line 806: If this is the same as `DEFAULT_BUCKETS`, there's no need to specify this here any longer (same for all tests)

etna/search/views.py
```diff
@@ -404,30 +405,11 @@ def get_api_aggregations(self) -> List[str]:
...
-                # for any bucket/tab options we might be showing
-                # (not just the 10 most popular)
-                item_count = 30
-            values.append(f"{aggregation}:{item_count}")
-        return values
+        The aggregations params may be specific to a bucket and will be filtered upon.
+        Returns a list of aggregation params for the current bucket.
+        Ex: ["group:30", "catalogue:10",]
+        """
+        return self.current_bucket.aggregations_normalised
```
ababic, line 412: @TNA-Allan What happens for views that don't have a 'current bucket' (like `FeaturedSearchView`)

TNA-Allan: Those view will need specific Aggregations passed to it.  I thought to include this change along with [DF-641](https://national-archives.atlassian.net/browse/DF-641)  Aggregations passed to the search API should be restricted to filters in "Refine Results" for the selected bucket. (Update the ticket to reflect this as well)

ababic (approved):
Thanks @TNA-Allan. This is good to merge now. I do think we need to look at docstrings for test cases to explain what they are testing and why, but that can be picked up as a separate bit of work.

TNA-Allan:
> Thanks @TNA-Allan. This is good to merge now. I do think we need to look at docstrings for test cases to explain what they are testing and why, but that can be picked up as a separate bit of work.

@ababic , I'm not sure what this is about. We can discuss.

PR #810, opened 2023-04-18
UN-470: [BE] Record article promo component

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/UN-470

## About these changes

Adds the `FeaturedRecordArticleBlock` block option, and makes it available as a content option on the `ArticlePage`

## How to check these changes

Edit an "Article Page" (e.g. [The Mangrove Nine](http://localhost:8000/admin/pages/87/edit/))

Within a content section, add a new block type and confirm you see the new "Featured record article" option. It should look a little bit like this:

Screenshot 2023-04-18 at 5 35 41 pm

The block just allows you to select a "Record Article", and that is all.

FE to be reviewed and refined in: https://national-archives.atlassian.net/browse/UN-169

## Merging PR guidance

## Deployment guidance


templates/blocks/featured_record_article.html
```diff
@@ -0,0 +1,37 @@
+{% load static wagtailimages_tags wagtailcore_tags %}
+
+{% if value.page.live or request.is_preview %}
```
William-Blackie, line 3: Thought: It's strange to handle this in the FE and could lead to a broken experience for editors that would be hard to figure out. I'm not suggestion that this is not the correct way but feel's fragile, is there anything we can do in the BE to enforce this?

William-Blackie (approved):

templates/blocks/featured_record_article.html
ababic: Problem there is... it's perfectly valid for in-progress content to link to other in-progress content (maybe they are working on a 'section' of related content). Pages also change states in a way that is difficult to enforce, and probably shouldn't be enforced. If you linked to a live page, then that page is later unpublished temporarily while it undergoes an update, does that mean the page with the link is now invalid?

ababic: I can tell you from experience that 'validation' isn't the answer :)

ababic: There might be some way to do this sort of thing with custom chooser and structblocks in Wagtail, but that would definitely need to be a separate piece of work.

PR #811, opened 2023-04-19
CHORE: Set `self.current_bucket` in search views

ababic:


etna/search/views.py
```diff
@@ -49,11 +49,20 @@ class BucketsMixin:
...
     # The source data for get_buckets()
     bucket_list: BucketList = None
-    # Can be updated by the view for get_current_bucket_key() to pick up
+
+    # To be updated by the view for get_current_bucket() to pick up
     current_bucket_key: str = None
 
-    def get_current_bucket_key(self):
-        return self.current_bucket_key
+    def get_current_bucket(self) -> Union[None, Bucket]:
```
TNA-Allan, line 56:
```suggestion
    def get_bucket(self) -> Union[None, Bucket]:
```
Since we have an attribute `Bucket.is_current` and setting of current is determined in `get_buckets()` Would this name be fine?

TNA-Allan (changes requested):

etna/search/views.py
TNA-Allan: I think defer the wording for this `_current_` until its set in `get_buckets()` ?

ababic: Removed in [ff66ac9](https://github.com/nationalarchives/ds-wagtail/pull/811/commits/ff66ac9d4611f984dcd749217c286ca07afeec0b)

etna/search/views.py
```diff
@@ -75,10 +74,10 @@ def get_buckets(
 
         bucket_list = copy.deepcopy(self.bucket_list)
 
-        if group_buckets:
+        if aggregations_data:
```
TNA-Allan, line 77:
```suggestion
        if bucket_counts:
```

etna/search/views.py
```diff
@@ -75,10 +74,10 @@ def get_buckets(
...
             # set `result_count` for each bucket
             doc_counts_by_key = {
-                group["key"]: group["doc_count"] for group in group_buckets
+                group["key"]: group["doc_count"] for group in aggregations_data
```
TNA-Allan, line 80:
```suggestion
                group["key"]: group["doc_count"] for group in bucket_counts
```

TNA-Allan (changes requested):

etna/search/views.py
```diff
@@ -226,7 +224,11 @@ def get(self, request: HttpRequest, **kwargs: Any) -> HttpResponse:
         form = self.form = self.get_form()
         is_valid = form.is_valid()
         self.api_result = None
+
         self.current_bucket_key = form.cleaned_data.get("group")
+        if self.current_bucket_key and getattr(self, "bucket_list", None):
+            self.current_bucket = self.bucket_list.get_bucket(self.current_bucket_key)
```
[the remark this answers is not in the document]
ababic: 1. Because good code does what it needs to do when it needs to. To get the aggregations for the current bucket, you'll just access via  `self.current_bucket.aggregatioms`. Simple, no? 
2. Two different things... Two different purposes, same source data. `self.current_bucket` is to allow the view to access relevant config for that bucket. The other is data is for the context/template to facilitate rendering. Didn't I already explain this on Slack?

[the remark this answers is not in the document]
TNA-Allan: > 1) Why wouldn't we update Bucket.is_current attribute can be updated at this point, therefore making the Bucket.aggregations attribute to pass on to the API?
> 2)  The line of code seems complicated to know that we have two different ways for a current_bucket

Thoughts: The thought in that [PR](https://github.com/nationalarchives/ds-wagtail/pull/808) was use _only one method to get the current bucket_ that is established by `is_current` , than two separate ones. However as per the design of using constants "Bucket as a constant". it did change its value.  Moving the Bucket class to _another file would allow using one method to get the current bucket,_ but perhaps it may be complicated for other reasons.

It appears to me that the `Bucket` constant has two sets of attributes:
1) attributes holds fixed values (`key, label`, etc)
2) attributes "place holders" hold variable values that are meant to be updated (`is_current`, etc). They are updated after taking a copy of BucketList constant for update. 
In this way the bucket remains a constant and the copied/class attribute is passed to the context. And this results in :
> Two separate things, two different purposes, same source data.
Where as the thought was about using the one method to get the current bucket where `is_current` is set to True.

The other bit, question was not how to use it, I am aware of how its going to be used with the changes you made - this will change "self.bucket_list.current.<aggregations>" 

Regarding the line for explanation - to elaborate more it was for the purpose making sure the "constant" is not updated as done in the PR. Is the Bucket pattern common, or perhaps if its required add a doc string over there.

I really do appreciate the patience and explanation. Thanks @ababic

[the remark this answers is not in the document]
ababic: Just generally speaking, I think it's important that we zoom out and look at the bigger picture here:
1. It's very rare to come across "perfect" code that everybody understands in an instant. It's even rarer in a project very much in active development.
2. The aim of ANY pull request should be to achieve what it needs to with minimal disruption to other code. Sprawling, multi-factored pull requests are difficult to review and encourage rabbit holes of discussion like this that tend to resort in miscommunication and waste far more time than is necessary.
3. As the primary maintainer of this code going forward, you should definitely be keeping a mental list about the things that cause misunderstanding in the codebase, things that are broken or noticeably inefficient. It's also your responsibility to understand what changes are worthwhile and which are not. Broadly speaking, I would say this codebase is far too young and hasn't had enough exposure to other devs for us to know what is clear, what isn't, and what will cause real problems in the long run. Anecdotal misunderstanding of other peoples code is common.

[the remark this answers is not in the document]
TNA-Allan: Thanks @ababic

TNA-Allan (approved):
Nice work. Thanks @ababic

PR #877, opened 2023-05-10
CHORE: Upgrade poetry to 1.4.2

ababic:
Resolves build errors for me locally. Hopefully, the same will be true of platform.sh :)


TNA-Allan (approved):
Thanks @ababic

PR #879, opened 2023-05-11
DF-658: Back to search results

TNA-Allan:
Ticket URL: https://national-archives.atlassian.net/browse/DF-658

## About these changes

These changes adopt storing the search url in a sesssion so that when navigating to a details pages from search will enable the user to get back to the same search url before navigation to the details page.
This functionality is applies to 
- Featured search
- Catalogue search
- Record Details page (which has the back to search results button)

## How to check these changes

From search
1. Navigate to Catalogue search/Featured search and perform a search query.
2. Navigate to a record details from the search results.
3. Navigate to other links within the search (optional)
4. Click on Back to search results - will get back to 1) search query

From other locations which have access to record details page.
1. Navigate to an Article Page
2. Navigate to a record details.
3. Navigate to other links within the search (optional)
4. Click on Start a new search - will get to featured search

## Merging PR guidance

## Deployment guidance


etna/core/middleware.py
```diff
@@ -122,3 +125,53 @@ def process_template_response(
...
+        return response
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        # create or retain session var
+        if hasattr(view_func, "view_class"):
+            # set session when navigating from search view containing links to record details page
+            if view_func.view_class in (CatalogueSearchView, FeaturedSearchView):
+                request.session.set_expiry(settings.SESSION_EXPIRY_VALUE)
+                request.session["back_to_search_url"] = request.get_full_path()
+                return None
```
ababic, line 150: Why do we need a Middleware for this? Why not do this in the views themselves?

etna/records/views/records.py
```diff
@@ -76,10 +77,20 @@ def record_detail_view(request, iaid):
     # if page.is_digitised:
     #     image = Image.search.filter(rid=page.media_reference_id).first()
 
+    if request.session.get("back_to_search_url", ""):
+        back_to_search_url = request.session.get("back_to_search_url")
+        back_to_search_label = "Back to search results"
+    else:
+        url = reverse("search-featured")
+        back_to_search_url = f"{url}"
+        back_to_search_label = "Start a new search"
```
ababic, line 86:
```suggestion
        featured_search_url = reverse("search-featured")
        back_to_search_url = request.session.get("back_to_search_url") or featured_search_url
        back_to_search_label = "Start a new search" if back_to_search_url == featured_search_url else "Back to search results"
```

config/settings/base.py
```diff
@@ -397,3 +398,7 @@
 FEATURE_FEEDBACK_MECHANISM_ENABLED = strtobool(
     os.getenv("FEATURE_FEEDBACK_MECHANISM_ENABLED", "False")
 )
+
+SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
```
ababic, line 402: As outlined in [the docs](https://docs.djangoproject.com/en/4.2/topics/http/sessions/), this changes the SESSION_ENGINE for all of Django's session usage. Is that what you want?

etna/core/middleware.py
```diff
@@ -122,3 +125,53 @@ def process_template_response(
...
+        response = self.get_response(request)
+
+        return response
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        # create or retain session var
+        if hasattr(view_func, "view_class"):
+            # set session when navigating from search view containing links to record details page
+            if view_func.view_class in (CatalogueSearchView, FeaturedSearchView):
+                request.session.set_expiry(settings.SESSION_EXPIRY_VALUE)
```
ababic, line 148: Is the aim here to limit the length of time the value is respected? Why do we care?

ababic: Setting the session to expire here affects more than just the life of this data (users will be logged out earlier too). We should think of alternative solutions really.

etna/records/views/records.py
TNA-Allan: Nice thanks

etna/core/middleware.py
TNA-Allan: I thought to have it in one place - create+delete the sessions. Its not needed here otherwise. I've moved it to the views.

etna/core/middleware.py
TNA-Allan: > Is the aim here to limit the length of time the value is respected? Why do we care?
You are right. I checked without the settings its the same, not required.

> Setting the session to expire here affects more than just the life of this data (users will be logged out earlier too). We should think of alternative solutions really.
I don't know another solution. Please advise.

ababic: As far as I can tell, you have two options:
1. Don't worry about it: The session will expire when it expires, and the data will no longer be available.
2. Add a timestamp to the session at the same time, and use that to decide how long to respect the session data for.

etna/core/middleware.py
```diff
@@ -122,3 +124,27 @@ def process_template_response(
...
+        return response
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        if view_func.__name__ == record_detail_view.__name__:
+            # retain session var regardless of is exitence
+            return None
+
+        # delete session var a non-search view or function
+        if request.session.get("back_to_search_url", ""):
+            del request.session["back_to_search_url"]
```
ababic, line 149: Do we need to do this? If the user searches again, the value will be updated. If they end up on a record detail page some other way, is remembering their search a big problem?

TNA-Allan: Good catch @ababic Fixed to retain the session for all views.

etna/core/middleware.py
TNA-Allan: >2. Add a timestamp to the session at the same time, and use that to decide how long to respect the url value for.

I'm a bit confused with point 2) with 
> Setting the session to expire here affects more than just the life of this data (users will be logged out earlier too).

Wouldn't the user get logged out on the timestamp?

ababic: I'm not saying to set an expiry on the session... just to add a timestamp to the data, like we are doing for the URL value.

etna/records/tests/test_views.py
```diff
@@ -722,3 +723,72 @@ def test_invalid_response_from_kong_raises_404(self):
...
+        record_detail_url = reverse(
+            "details-page-machine-readable", kwargs={"iaid": "C13359805"}
+        )
+        session = self.client.session
+        session["back_to_search_url"] = expected_url
+        session.save()
+        response = self.client.get(record_detail_url)
+        expected_label = "Back to search results"
+        self.assertEqual(response.context.get("back_to_search_label"), expected_label)
+        self.assertEqual(response.context.get("back_to_search_url"), expected_url)
```
ababic, line 751: **Suggestion**: You could swap these two assertions for a an [assertContains](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.SimpleTestCase.assertContains) to gain additional assurance that the template is using the correct variables and rendering the button successfully.

etna/records/tests/test_views.py
```diff
@@ -722,3 +723,72 @@ def test_invalid_response_from_kong_raises_404(self):
...
+            "https://kong.test/data/fetch",
+            json=create_response(
+                records=[
+                    create_record(iaid="C13359805"),
+                ]
+            ),
+        )
+        record_detail_url = reverse(
+            "details-page-machine-readable", kwargs={"iaid": "C13359805"}
+        )
```
ababic, line 744: We should move as much duplicate code from out of individual test cases and into `setUp()` as we can

etna/records/tests/test_views.py
TNA-Allan: - Have updated the tests for assertContains, however  `parsed_url = parse.quote(browser_search_url, safe="")` does not feel right for a scenario with query params - `def test_back_to_search_render_with_catalogue_search` .This is regardless of using urllib.parse.quote/unquote to set or decode the saved search.
- Unsure about another way to run this test without encoding the saved url in the unit test.

etna/records/tests/test_views.py
```diff
@@ -739,56 +736,51 @@ def test_back_to_search_context_with_catalogue_search(self):
...
         session.save()
-        response = self.client.get(record_detail_url)
-        expected_label = "Start a new search"
-        self.assertEqual(response.context.get("back_to_search_label"), expected_label)
-        self.assertEqual(response.context.get("back_to_search_url"), expected_url)
+
+        response = self.client.get(self.record_detail_url)
+
+        self.assertContains(response, back_to_search_label)
+        self.assertContains(response, f'href="{browser_search_url}"')
```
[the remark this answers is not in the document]
TNA-Allan: The link with label contains an include see https://github.com/nationalarchives/ds-wagtail/blob/6592dbd641c6a6f788d28bf5a76c80121bcf88b8/templates/includes/records/record-cta-panel.html#L5-L8 which I thought would unnecessary add a level of attributes not belonging to this piece of work to be tested.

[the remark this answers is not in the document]
ababic: Sure, but you can copy and paste the generated HTML for the button then check for the presence of that. That'll make for a clearer and more efficient test.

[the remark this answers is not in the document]
ababic: If the classnames etc are revised in future, the breakage should be easy to fix.

etna/records/tests/test_views.py
ababic: Why not just recreate in local dev and copy the generated HTML into a variable?

etna/core/middleware.py
TNA-Allan: Thanks Andy, now updated for option 2)

etna/ciim/constants.py
```diff
@@ -753,3 +754,5 @@ class Display(StrEnum):
         "long_display_name": "Paper catalogues available to view at The National Archives",
     },
 ]
+
+SEARCH_URL_RETAIN_DELTA = timezone.timedelta(hours=48)
```
ababic, line 758: @TNA-Allan I don't feel like this app is the right place for this value. I would suggest just adding it to `etna/records/views/records.py` as a global

etna/records/tests/test_views.py
```diff
@@ -722,3 +725,100 @@ def test_invalid_response_from_kong_raises_404(self):
...
+        """navigation to record details from previous search (session is set since its coming from search catalogue)"""
+
+        # raw search value: "/search/catalogue/?sort_by=title&q=london&filter_keyword=paper&level=Item&collection=ADM&collection=BT&closure=Open+Document%2C+Open+Description&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=1900&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=2020&per_page=20&sort_order=asc&display=list&page=2&group=tna"
+        search_url_gen_html_resp = "%2Fsearch%2Fcatalogue%2F%3Fsort_by%3Dtitle%26q%3Dlondon%26filter_keyword%3Dpaper%26level%3DItem%26collection%3DADM%26collection%3DBT%26closure%3DOpen%2BDocument%252C%2BOpen%2BDescription%26opening_start_date_0%3D%26opening_start_date_1%3D%26opening_start_date_2%3D1900%26opening_end_date_0%3D%26opening_end_date_1%3D%26opening_end_date_2%3D2020%26per_page%3D20%26sort_order%3Dasc%26display%3Dlist%26page%3D2%26group%3Dtna"
+
+        session = self.client.session
+        session[
+            "back_to_search_url_timestamp"
+        ] = self.back_to_search_url_timestamp.strftime("%Y%m%d-%H%M%S")
```
ababic, line 759: We should keep things simple/standard and use `datetime.isoformat()` to convert datetimes to strings

etna/records/views/records.py
```diff
@@ -76,10 +83,28 @@ def record_detail_view(request, iaid):
...
     #     image = Image.search.filter(rid=page.media_reference_id).first()
 
+    # Back to search - default url
+    back_to_search_url = reverse("search-featured")
+
+    # Back to search button - update url when timestamp is not expired
+    if back_to_search_url_timestamp := request.session.get(
+        "back_to_search_url_timestamp"
+    ):
+        back_to_search_url_timestamp = datetime.datetime.strptime(
```
ababic, line 93: We should keep things simple/standard and use `datetime.fromisoformat()` to convert strings back to datetimes

etna/records/views/records.py
```diff
@@ -76,10 +83,28 @@ def record_detail_view(request, iaid):
...
+            back_to_search_url_timestamp, "%Y%m%d-%H%M%S"
+        )
+        back_to_search_url_timestamp = back_to_search_url_timestamp.astimezone(pytz.utc)
```
ababic, line 96: If we use `datetime.isoformat()` and `datetime.fromisoformat()`, we shouldn't have to worry about the timezone here

etna/search/tests/test_views.py
```diff
@@ -1717,3 +1717,47 @@ def test_record_creators_country_long_filter(self):
                 "&size=20"
             ),
         )
+
+
+class CreateSessionInfoForBackToSearchTest(SearchViewTestCase):
+    @responses.activate
+    def test_create_session_info_for_catalogue_search(self):
+        search_url = reverse("search-catalogue")
+        expected_url = "/search/catalogue/?sort_by=title&q=london&filter_keyword=paper&level=Item&collection=ADM&collection=BT&closure=Open+Document%2C+Open+Description&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=1900&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=2020&per_page=20&sort_order=asc&display=list&page=2&group=tna"
```
ababic, line 1726: These are pretty heavy-weight tests for just checking that a session value is set. Could we see instead about extending one of the existing tests to include the check in a single request/response cycle?

etna/search/views.py
```diff
@@ -313,6 +314,18 @@ def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
...
 
+    def set_session_info(self) -> None:
+        """
+        Usually called where there are links to the record details page in the search results.
+        Sets session in order for it to be used in record details page when navigating from search results.
+        """
+        back_to_search_url_timestamp = timezone.now()
+        self.request.session[
+            "back_to_search_url_timestamp"
+        ] = back_to_search_url_timestamp.strftime("%Y%m%d-%H%M%S")
```
ababic, line 325:
```suggestion
        ] = back_to_search_url_timestamp.isoformat()
```

etna/records/views/records.py
```diff
@@ -76,10 +83,28 @@ def record_detail_view(request, iaid):
...
 
+    # Back to search - default url
+    back_to_search_url = reverse("search-featured")
+
+    # Back to search button - update url when timestamp is not expired
+    if back_to_search_url_timestamp := request.session.get(
+        "back_to_search_url_timestamp"
+    ):
+        back_to_search_url_timestamp = datetime.datetime.strptime(
+            back_to_search_url_timestamp, "%Y%m%d-%H%M%S"
```
ababic, line 94: Do we need to worry about the value not matching the format here?

etna/search/views.py
```diff
@@ -313,6 +314,18 @@ def get_context_data(self, **kwargs: Any) -> Dict[str, Any]:
...
+        self.request.session["back_to_search_url"] = self.request.get_full_path()
```
ababic, line 326: Tiny tiny thing: I would set the URL first, as the timestamp means nothing with it.

etna/records/tests/test_views.py
```diff
@@ -722,3 +725,100 @@ def test_invalid_response_from_kong_raises_404(self):
...
+                    create_record(iaid="C13359805"),
+                ]
+            ),
+        )
+
+        self.record_detail_url = reverse(
+            "details-page-machine-readable", kwargs={"iaid": "C13359805"}
+        )
+
+        self.expected_button_link_gen_value_fmt = '<a class="cta-primary-panel__link" href="{back_to_search_url}" data-link-type="Link" data-link="Back to search results" data-component-name="Navigation">\n\n\n\n<svg\n    class="icon icon--chevron-left cta-primary-panel__link-icon"\n    aria-hidden="true"\n    focusable="false"\n>\n    <use xlink:href="#chevron-left" href="#chevron-left"></use>\n</svg>\nBack to search results</a>'
```
ababic, line 746: What happens if we just add the opening tag here? 
```suggestion
        self.expected_button_link_gen_value_fmt = '<a class="cta-primary-panel__link" href="{back_to_search_url}" data-link-type="Link" data-link="Back to search results" data-component-name="Navigation">'
```

etna/records/tests/test_views.py
```diff
@@ -722,3 +725,100 @@ def test_invalid_response_from_kong_raises_404(self):
...
+        self.back_to_search_url_timestamp = timezone.now()
+
+    @responses.activate
+    def test_back_to_search_render_with_catalogue_search_within_expiry(self):
+        """navigation to record details from previous search (session is set since its coming from search catalogue)"""
+
+        # raw search value: "/search/catalogue/?sort_by=title&q=london&filter_keyword=paper&level=Item&collection=ADM&collection=BT&closure=Open+Document%2C+Open+Description&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=1900&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=2020&per_page=20&sort_order=asc&display=list&page=2&group=tna"
+        search_url_gen_html_resp = "%2Fsearch%2Fcatalogue%2F%3Fsort_by%3Dtitle%26q%3Dlondon%26filter_keyword%3Dpaper%26level%3DItem%26collection%3DADM%26collection%3DBT%26closure%3DOpen%2BDocument%252C%2BOpen%2BDescription%26opening_start_date_0%3D%26opening_start_date_1%3D%26opening_start_date_2%3D1900%26opening_end_date_0%3D%26opening_end_date_1%3D%26opening_end_date_2%3D2020%26per_page%3D20%26sort_order%3Dasc%26display%3Dlist%26page%3D2%26group%3Dtna"
```
ababic, line 754: I'm not sure about the variable name here... feels like the original one is a better fit

ababic (changes requested):
Thanks @TNA-Allan. I think we're close!

etna/records/tests/test_views.py
TNA-Allan: Thanks for the ISO format - when you mentioned simple I was wondering what what it about.

etna/ciim/constants.py
TNA-Allan: This is moved to records.py

etna/records/views/records.py
TNA-Allan: Done.

etna/records/views/records.py
TNA-Allan: Done.

etna/search/views.py
TNA-Allan: Done done :)

etna/records/tests/test_views.py
TNA-Allan: Missed this bit as the earlier implementation had a label in the button which was changing. Fixed.

etna/records/tests/test_views.py
TNA-Allan: I kept this `search_url_gen_html_resp`  separate so that we focus only on the url here, the other bit - `expected_button_link_gen_value_fmt` may change independently - ex. the class name or so.

TNA-Allan: A small change updated - removed raw search value comment.

etna/records/tests/test_views.py
```diff
@@ -722,3 +725,97 @@ def test_invalid_response_from_kong_raises_404(self):
...
+        response = self.client.get(self.record_detail_url)
+
+        expected_button_link_gen_value = self.expected_button_link_gen_value_fmt.format(
+            back_to_search_url=new_search_url
+        )
+        self.assertContains(response, expected_button_link_gen_value)
+
+    @responses.activate
+    def test_new_search_render_with_session(self):
+        """navigation to record details from previous search (session is set since its coming from search featuerd, but without query)"""
```
ababic, line 805: Not a blocking thing, but these docstrings should really do a better job of outlining the expectations of the test - not just the scenario they are testing.

etna/records/views/records.py
```diff
@@ -76,10 +83,28 @@ def record_detail_view(request, iaid):
...
+    if back_to_search_url_timestamp := request.session.get(
+        "back_to_search_url_timestamp"
+    ):
+        back_to_search_url_timestamp = datetime.datetime.fromisoformat(
+            back_to_search_url_timestamp
+        )
+        back_to_search_url_timestamp_delta = (
+            back_to_search_url_timestamp + SEARCH_URL_RETAIN_DELTA
+        )
+        if timezone.now() <= back_to_search_url_timestamp_delta:
```
ababic, line 100: Non-blocking: There's no need to define a variable for everything, especially when variable names are so lengthy. You could just calculate and test in a single line here:

```suggestion
        if timezone.now() <= (back_to_search_url_timestamp + SEARCH_URL_RETAIN_DELTA):
```

ababic (approved):

etna/records/views/records.py
TNA-Allan: I did this somewhere else in an assignment and missed it here - `session["back_to_search_url_timestamp"] = (self.back_to_search_url_timestamp - SEARCH_URL_RETAIN_DELTA).isoformat()` - Fixed

etna/records/tests/test_views.py
TNA-Allan: I came up with this "Test covers navigation to record details without a previous search (session is not set since its not coming from search)". Happy to change it otherwise.

PR #887, opened 2023-05-15
CHORE: Update dark mode develop theme

jamesbiggs:
## About these changes

I have made the side-bar on develop a dark green to fit in with the dark mode - this is so that the editors can tell the difference between develop and main.

## How to check these changes

Set `FEATURE_PLATFORM_ENVIRONMENT_TYPE = "development"` in your `local.py` and it will show you the styles that are meant for develop only.

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/core/wagtail_hooks.py
+++ b/etna/core/wagtail_hooks.py
@@ -22,5 +22,5 @@ def editor_js():
 @hooks.register("insert_global_admin_css")
 def global_admin_css():
     if settings.FEATURE_PLATFORM_ENVIRONMENT_TYPE != "production":
-        return "<style> @media (prefers-color-scheme: light) { :root {--w-color-primary: #00623B; --w-color-primary-200: #003c1e;} } @media (prefers-color-scheme: dark) { :root {--w-color-surface-menus: #002510; --w-color-surface-menu-item-active: #002010;} }</style>"
+        return "<style> @media (prefers-color-scheme: light) { :root {--w-color-primary: #00623B; --w-color-primary-200: #003c1e;} } @media (prefers-color-scheme: dark) { :root {--w-color-surface-menus: #002510; --w-color-surface-menu-item-active: #001810;} }</style>"
     return ""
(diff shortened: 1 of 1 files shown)


TNA-Allan:
@james

Re: 

> How to check these changes
Delete the conditional statement in the wagtail_hooks.py, and it will load the CSS that is on develop only

Wanted to avoid any physical changes for testing. Would the below help instead ?

1. An alternative is keep the var in local setup to run locally.
2. Below change - still better, I think, as it would severe the purpose of all environments production, develop and even local - and the intention may be so?

https://github.com/nationalarchives/ds-wagtail/blob/014a64a943aac6193c2c8c4865e796b6c4e5206f/config/settings/base.py#L396

_Change to_ 
`FEATURE_PLATFORM_ENVIRONMENT_TYPE = os.getenv("PLATFORM_ENVIRONMENT_TYPE", "any-other-environment")`

jamesbiggs:
Yes you could do that on your local :) However you can display the difference in "environments" is fine. This is already on develop, but only has styles for light-mode so this change was just for dark mode - the difference between develop/main (develop/prod env) is already tested + working :)

jamesbiggs:
@TNA-Allan 1. okay :)
2. I'm not entirely sure what you meant by this, but I think having it default to `production` is okay :)

TNA-Allan:
> @TNA-Allan 1. okay :) 2. I'm not entirely sure what you meant by this, but I think having it default to `production` is okay :)

Re: 
> Below change - still better, I think, as it would severe the purpose of all environments production, develop and even local - and the intention may be so


- If we don't have a local var `FEATURE_PLATFORM_ENVIRONMENT_TYPE`, the settings will go with "production". This update will take care of that. Also when deployed to platform sh, it will still respect the `PLATFORM_ENVIRONMENT_TYPE`
- Currently we don't have a documentation for setting this on local, this would make a production like setting. A step in developer setup will be helpful. if we're retaining the default "production".

Thanks for updating 1) - I'm going with this for this PR.

TNA-Allan (approved):
Looks good. Thanks @jamesbiggs

PR #894, opened 2023-05-17
DF-664:Fix show record opening date error for invalid date in Catalogue search

TNA-Allan:
Ticket URL: https://national-archives.atlassian.net/browse/DF-664

## About these changes

Changes to show record opening date error for invalid date in Catalogue search

## How to check these changes

- Navigate to Catalogue Results - Records at the National Archives
- Enter invalid date range in record opening date filter, From Date after To Date ex: From 2000, To 1999

Observe 
- the error message displayed
- No results message

## Merging PR guidance

## Deployment guidance


etna/search/views.py
```diff
@@ -379,6 +378,12 @@ def form_invalid(self, form):
             if field_name in form.errors:
                 return HttpResponseBadRequest(str(form.errors[field_name]))
 
+        if not self.api_result:
```
ababic, line 381: We can drop this conditional and always do this... the API shouldn't have been queried if the form is invalid.

etna/search/views.py
```diff
@@ -90,17 +90,16 @@ def get_buckets_for_display(
...
-            for bucket in buckets:
-                if bucket.result_count:
-                    buckets_contain_results = True
-                    break
+        # Set this to True if any buckets have results
+        buckets_contain_results = False
+        for bucket in buckets:
+            if bucket.result_count:
+                buckets_contain_results = True
+                break
```
ababic, line 102: Can we revert these `get_context_data()` changes if they aren't actually needed?

ababic (changes requested):
Hi @TNA-Allan. In addition to the feedback points already shared, I'm struggling to understand the reasoning behind a lot of the template changes. If the page is otherwise functioning as expected, we should be changing as little as possible.

TNA-Allan:
> Hi @TNA-Allan. In addition to the feedback points already shared, I'm struggling to understand the reasoning behind a lot of the template changes. If the page is otherwise functioning as expected, we should be changing as little as possible there.

Hi @ababic 

- This condition [here ](https://github.com/nationalarchives/ds-wagtail/blob/99f0561faa27ae37274a78737ecdac5d281a3114/templates/search/catalogue_search.html#L32-L35) blocks the the form errors from showing - which is the main issue.
- Its based on the the code [before](https://github.com/nationalarchives/ds-wagtail/blob/39bf313563744d381acd4c3c8002e20ab42df750/templates/search/catalogue_search.html) DF-594 with some minor tweaks, and making absolutely sure the `EndToEndSearchTestCase` from DF-594 is not broken.
- While there are tests around date errors Eg. `test_opening_start_date_before_opening_end_date_is_invalid`,  the new test case `test_render_invalid_date_range_message` in this PR will ensure date form errors it wont be broken.

etna/search/views.py
TNA-Allan: Nice. Fixed.

etna/search/views.py
TNA-Allan: The condition [if self.bucket_list](https://github.com/nationalarchives/ds-wagtail/blob/99f0561faa27ae37274a78737ecdac5d281a3114/etna/search/views.py#L96-L111): is dropped as 
- The [return ](get_context_data)assigns uninitialized  values in context when if condition is used.
- the `bucket_list` is assigned for search views [ex.](https://github.com/nationalarchives/ds-wagtail/blob/99f0561faa27ae37274a78737ecdac5d281a3114/etna/search/views.py#L586)

ababic: Okay. But as a rule, we should be aiming to keep changes as minimal as possible. Not only does it reduce unnecessary risk and make change history much easier to track... It makes review less open to misinterpretation.

templates/search/blocks/search_results.html
```diff
@@ -24,6 +24,7 @@
         {% include './search_results__selected-filters.html' %}
 
         {% if form.errors or not paginator.count %}
+            {# No results were found due to criteria not staisfying the refine results filters #}
```
ababic, line 27: I think we should remove this, as it might just be that there were no results.

TNA-Allan: Removed.

etna/search/tests/test_views.py
```diff
@@ -480,6 +480,58 @@ def test_selected_filter_options_remain_visible(self):
...
+        self.assertIn(
+            "<li>Try removing any filters that you may have applied</li>", content
+        )
+        self.assertIn(
+            '<h2 class="search-filters__heading">Refine results</h2>', content
+        )
+        self.assertIn(
+            '<input type="text" name="filter_keyword" value="kew" class="search-filters__search" id="id_filter_keyword">',
+            content,
+        )
```
ababic, line 527: For the sake of consistency, are there no existing assertions that can be used here?

TNA-Allan: Nice. Fixed.

TNA-Allan: > For the sake of consistency, are there no existing assertions that can be used here?

I've updated and fixed to reflect the changes for error messages specifically.

TNA-Allan:
@ababic there may be some issues not spotted earlier with the template or so. But that will require investigation and fix on another ticket. But for now, this ticket is to resolve the issue for date errors.

ababic (approved):

PR #900, opened 2023-05-19
UN-596: Featured record link

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-596

## About these changes

Created FE following the designs listed on the ticket

## How to check these changes

Create a "record links" block on an Article page, and check that the designs meet the desktop and mobile view.

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/sass/includes/_record-links.scss
+++ b/sass/includes/_record-links.scss
@@ -73,7 +73,11 @@
         
         &__title a:hover,
         &__title a:focus{
-            outline: 0.312rem solid $color__dark;
+            color: #004C7E;
+        }
+
+        &__title a:focus{
+            outline: 0.312rem solid #004C7E;
         }
 
         &__dates {


sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
+.record-links {
+    &__wrapper {
+        display: flex;
+        align-items: center;
+        flex-direction: column;
+        padding-left:10%;
+        padding-right:10%;
+
+        @media (max-width: $screen__md) {
```
janehughes, line 9: Tip: it's best practice (and v helpful) to build from a mobile-first approach. This way, the first styles you add target mobile and then any media queries are for anything above mobile. So you'd do `@media min-width` rather than `@media (max-width ...)`. It's a slightly different way of thinking about it but it makes the styling process much more efficient, mobile browsers only render the relevant styles and it means you don't need to override or re-create the same properties multiple times.

sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
+.record-links {
+    &__wrapper {
+        display: flex;
+        align-items: center;
+        flex-direction: column;
+        padding-left:10%;
```
janehughes, line 6: And lets put this left & right 10% padding in a larger breakpoint so it's only targeting tablet/desktop

sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
...
+            padding-right:5%;
+        }
+    }
+
+    &__list {
+        list-style: none;
+        margin-bottom: 0px;
+        width: 100%;
+        padding-left: 0;
+        border-top: 0.3rem solid #ffcc00;
```
janehughes, line 20:
```suggestion
        border-top: 0.3rem solid $color__yellow_2;
```

sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
...
+
+    &__item {
+        display: flex;
+        align-items: center;
+        border: 1px solid #D9D9D6;
+        padding-top: 2rem;
+        padding-bottom: 2rem;
+        width: 100%;
+
+        @media (max-width: $screen__lg) {
```
janehughes, line 41: As I mentioned above - lets swap out media queries for min-widths and move/swap the relevant styles as needed

sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
...
+    
+    &__list a:hover &__item,
+    &__list a:focus &__item{
+        outline: 0.312rem solid $color__dark;
+    }
+
+    &__item {
+        display: flex;
+        align-items: center;
+        border: 1px solid #D9D9D6;
```
janehughes, line 36: I added a new variable for this colour in a recent PR - looks like it's `$color__grey-9` :)

janehughes:
```suggestion
        border: 1px solid #D9D9D6;
        border-top: 0;
```

We can get rid of the border-top - this will eliminate the 'double border' effect when two items sit together. The parent container's top yellow border acts as the top border for the first item in the list, so we can remove the top border for the individual items altogether :)

sass/includes/_record-links.scss
```diff
@@ -0,0 +1,92 @@
...
+            padding-right: 2rem;
+        }
+
+        &__title {
+            margin: 5px 0;
+            color: #000000;
+            text-decoration-line: underline;
+            font-family: $font__roboto-mono;
+            font-style: normal;
+            font-weight: 700;
```
janehughes, line 74: We're importing the roboto font from google, but it looks like we're missing the bold version as the '700' weight isn't actually having any effect on the title. Let me know if you're not sure how to update this

templates/records/blocks/record_links_block.html
```diff
@@ -1,23 +1,24 @@
...
         {% for item in value.items %}
-            {% with item.record.url as url %}
-                <li class="recordlinks__item">
-                    {% if item.thumbnail_image %}
-                        <div class="recordlinks__item__thumbnail">
-                            <a href="{{ url }}">
-                                {% image item.thumbnail_image fill-200x200 alt="" class="recordlinks__thumbnail-image" %}
-                            </a>
+                <li>
+                    <a href="{{ item.record.url }}">
```
janehughes, line 6: After a discussion with Will & Nick last week we've agreed to limit the links on cards to the title (or title and image) only. I think for this element keeping it to the title alone is sufficient as the image is quite small). So we'll need to move the `<a>` tag to inside the title please

janehughes (changes requested):
Thank you James! This looks really good - I've made a few change requests primarily around the structuring of the CSS and a few minor changes to neaten up the CSS a bit. Please let me know if you have any questions or need clarifications. Thanks!

templates/records/blocks/record_links_block.html
jamesbiggs: Hi @janehughes okay, thank you!
What link interaction are we having when hovering over link text? Every link seems to be different at the minute as it's all changing so I'm not sure what to copy :)

janehughes: Nick's very helpfully included all the link interaction styles in the design system - you can find this component here https://www.figma.com/file/uo8YY98lLws8E1HEfpdn5X/TNA-Etna---Design-system?type=design&node-id=2846-40842&t=dA9vGMTUz1iAojNv-0 :) If you skim through the other components though you can see general link styles follow the same pattern. Essentially remove underline on hover and use the blue colour. Focus should stay as the default (thick blue border)

janehughes (approved):
Good to go, thanks James :)

PR #904, opened 2023-05-19
Un 379 artice page image streamfield fe

William-Blackie:
Ticket URL: [ here
](https://national-archives.atlassian.net/jira/software/c/projects/UN/boards/75?modal=detail&selectedIssue=UN-226)
## About these changes

Add a human-readable description here

I've updated the referenced images in this ticket (Which is the content image block) as well as the other image blocks to use the new design styling, layout and images sizes.

I've also added a little BE to support checking images if they are landscape or portrait to change the styling for portait images, please note these changes are only on the block itself.

## How to check these changes

Add a ContentImageBlock to the ArticlePage, both a portrait image and lanscape. 
Here's some screencasts of these changes;
[Screencast from 19-05-23 16:16:58.webm](https://github.com/nationalarchives/ds-wagtail/assets/22497837/b7a20b10-157c-4f90-9090-d11bd6392b97)
[Screencast from 19-05-23 16:17:46.webm](https://github.com/nationalarchives/ds-wagtail/assets/22497837/83907adf-3c23-482b-84a0-89c92b2eba3b)

## Merging PR guidance

## Deployment guidance


templates/includes/hero-img.html
```diff
@@ -1,34 +1,39 @@
 {% load wagtailimages_tags %}
 {% load wagtailcore_tags %}
-{% if page.hero_image %}
+{% if hero_image %}
```
ababic, line 3: I can think we can safely assume that hero images are only used in the context of specific pages, right? In which case, `page.hero_image` is fine? Am I missing something?

templates/articles/article_page.html
```diff
@@ -10,7 +10,7 @@
 {% block breadcrumb %}{% endblock %}
 {% block content %}
     {% include 'includes/generic-intro.html' with breadcrumb=True title=page.title intro=page.intro %}
-    {% include 'includes/hero-img.html' %}
+    {% include 'includes/hero-img.html' with hero_image=page.hero_image caption=page.hero_image_caption %}
```
ababic, line 13: Personally, I would rather this just be `{% include "includes/page-hero-img.html" %}` and the template grab the necessary values from `page`. Is there a reason for this that I'm not getting?

ababic (commented):
Letting someone with better FE knowledge review the styles.. just a couple of questions around include context.

templates/includes/hero-img.html
William-Blackie: You're right, but the logic behind this it de-couples the BE variable naming convention to the FE logic. 

For example we have two pages, one page that has a specific hero image and one that does not so we use the listing image instead. We can use the same template by just passing in the correct image the variable `hero_image`. 

Expanding the scope a little bit, if you think about the streamfield elements that are also used in pages, it allows us to use one template for both as we can map to the variable names instead of the BE properties.

templates/articles/article_page.html
ababic: Happy with the explanation. Let's keep this as is

janehughes (approved):

PR #991, opened 2023-06-07
DF-698: Add Creation date filter to search filters

TNA-Allan:
Ticket URL: 
https://national-archives.atlassian.net/browse/DF-698
https://national-archives.atlassian.net/browse/DF-702

## About these changes

Add creation date filter to search

## How to check these changes

- Navigate to Catalogue search
- View and interact with the Date filter ranges in tna, nonTna, digitised buckets
   Ex: http://127.0.0.1:8000/search/catalogue/?filter_keyword=kew&created_start_date_0=01&created_start_date_1=01&created_start_date_2=2000&created_end_date_0=12&created_end_date_1=12&created_end_date_2=2012&opening_start_date_0=&opening_start_date_1=&opening_start_date_2=&opening_end_date_0=&opening_end_date_1=&opening_end_date_2=&per_page=20&sort_order=asc&display=list&group=tna
   

## Merging PR guidance

## Deployment guidance


etna/core/fields.py
```diff
@@ -80,6 +80,8 @@ def __init__(
 
         super().__init__(fields=fields, **kwargs)
 
+        self.label_suffix = ""  # Removes : as label suffix
+
```
ababic, line 84: Where did this requirement come from, @TNA-Allan?

etna/search/forms.py
```diff
@@ -189,7 +191,19 @@ class BaseCollectionSearchForm(forms.Form):
...
     )
     opening_end_date = DateInputField(
-        label="To",
+        label="to",
+        required=False,
+        default_day=END_OF_MONTH,
+        default_month=12,
+    )
+    created_start_date = DateInputField(
+        label="From",
```
ababic, line 200: Any reason why we're mixing case for labels here? In my experience, form field labels usually start with an upper-case letter.

etna/search/forms.py
```diff
@@ -234,8 +248,26 @@ def clean(self):
                 "opening_end_date"
             ):
                 self.add_error(
-                    "opening_start_date",
-                    CUSTOM_ERROR_MESSAGES.get("invalid_date_range"),
+                    None,
+                    ValidationError(
+                        _(CUSTOM_ERROR_MESSAGES.get("invalid_date_range")),
+                        code="invalid_date_range_for_opening_dates",
```
ababic, line 254: Using `invalid_date_range` as the code for both sets of dates should be fine. Think of it more as a 'category' for the error. You already know the 'for_opening_dates' part because of the field to which the error is attached.

etna/search/templatetags/search_tags.py
```diff
@@ -57,14 +62,28 @@ def query_string_exclude(
...
+        visible_field_names = "filter_keyword collection topic level closure opening_start_date opening_end_date created_start_date created_end_date catalogue_source held_by type country location"
+        if form.errors:
+            # visible non dynamic input fields when in error, includes hidden dynamic input fields
+            visible_field_names = "created_start_date created_end_date opening_start_date opening_end_date"
+    elif form_id_or_visible_field_names == "catalogue-search-form":
+        visible_field_names = "q"
+    elif form_id_or_visible_field_names == "catalogue-sort-form-desktop":
+        visible_field_names = "sort_by"
+    else:
+        visible_field_names = form_id_or_visible_field_names
```
ababic, line 85: We don't want this configuration living here in this stray function that nobody ever looks at. It should be down to the calling code to specify the fields that are shown each time. Can you think of another way for this to be provided?

etna/search/templatetags/search_tags.py
```diff
@@ -137,3 +131,19 @@ def extended_in_operator(lhs_operand, *rhs_operand_list) -> bool:
...
+    """
+    for item in form.non_field_errors().data:
+        if item.code == non_field_error_code:
+            error_text = item.error_list[0].message
+            html = '<ul class="errorlist nonfield"><li>{error_text}</li></ul>'.format(
+                error_text=error_text
+            )
+            return mark_safe(html)
+
+    return ""
```
ababic, line 149: @TNA-Allan I'm afraid I'm not sure where/how this relates to the other changes... could we perhaps save this for a separate PR?

etna/search/tests/test_forms.py
```diff
@@ -55,14 +73,20 @@ def test_various_opening_date_inputs_is_valid(self):
                 form = CatalogueSearchForm(form_data)
                 self.assertTrue(form.is_valid(), label)
 
-    def test_opening_start_date_before_opening_end_date_is_invalid(self):
+    def test_start_date_after_end_date_is_invalid(self):
```
ababic, line 76: Similar thing here. A separate test for each range would be clearer

etna/search/tests/test_views.py
```diff
@@ -505,7 +505,17 @@ def test_render_invalid_date_range_message(self):
...
+                "created_start_date_1": "",
+                "created_start_date_2": "2010",
+                "created_end_date_0": "",
+                "created_end_date_1": "",
+                "created_end_date_2": "2005",
+                "opening_start_date_0": "",
+                "opening_start_date_1": "",
                 "opening_start_date_2": "2000",
+                "opening_end_date_0": "",
+                "opening_end_date_1": "",
```
ababic, line 518: Same issue here, I would say. This dict was purposefully minimal before, and only contained the minimal data needed for the test. Now we've inflated it, it's difficult to understand what is being tested. Let's go with a separate test.

etna/search/views.py
```diff
@@ -526,45 +526,20 @@ def get_selected_filters(self, form: Form) -> Dict[str, List[Tuple[str, str]]]:
...
-                CUSTOM_ERROR_MESSAGES.get("invalid_date_range")
-                not in form_error_messages
-            ):
+        # prepare filter labels for date fields
+        for date_item in [
+            ("opening_start_date", "Record Opening from: %d-%m-%Y"),
+            ("opening_end_date", "Record Opening to:  %d-%m-%Y"),
+            ("created_start_date", "Date from: %d-%m-%Y"),
+            ("created_end_date", "Date to: %d-%m-%Y"),
+        ]:
```
ababic, line 535: This list should be a tuple, since it doesn't need to be mutable. We should unpack the item tuples here to make the following code more readable too, e.g.:
```suggestion
        for field_name, label in [
            ("opening_start_date", "Record Opening from: %d-%m-%Y"),
            ("opening_end_date", "Record Opening to:  %d-%m-%Y"),
            ("created_start_date", "Date from: %d-%m-%Y"),
            ("created_end_date", "Date to: %d-%m-%Y"),
        ]:
```

etna/search/views.py
```diff
@@ -526,45 +526,20 @@ def get_selected_filters(self, form: Form) -> Dict[str, List[Tuple[str, str]]]:
...
-            if (
-                CUSTOM_ERROR_MESSAGES.get("invalid_date_range")
-                not in form_error_messages
-            ):
-                return_value.update(
-                    {
-                        "opening_start_date": [
-                            (
-                                opening_start_date,
-                                opening_start_date.strftime(
```
ababic, line 546: @TNA-Allan Where does this validation happen now? It should really belong to the form, but I can't see it there at a glance.

templates/search/blocks/search_filters.html
```diff
@@ -8,7 +8,7 @@ <h2 class="search-sort-view__heading">
         {{ form.sort_by.errors }}
         {{ form.sort_by }}
 
-        {{ 'sort_by'|include_hidden_fields:form }}
+        {{ 'catalogue-sort-form-mobile'|include_hidden_fields:form }}
```
ababic, line 11: This feels a bit wrong. How do we know what magic value to use here? It's at least semi-clear when the field name is used.

templates/search/blocks/search_filters.html
```diff
@@ -77,6 +101,7 @@ <h4 class="search-filters__accordion-section-heading" id="record_opening_date">R
                             <!--
                             <p id="search_filters__opening-date-help-text">Helper text here (needed for when YYYY MM DD are eventually made separate fields)</p>
                             -->
+                            {{ "invalid_date_range_for_opening_dates"|non_field_error_by_code:form }}
```
ababic, line 104: I'm struggling to make sense of this. Why `|non_field_error_by_code`? Surely these errors belong to a field? We shouldn't be using validation error codenames in the template here either, really - it all feels way too specific.

templates/search/blocks/search_filters.html
```diff
@@ -91,11 +116,8 @@ <h4 class="search-filters__accordion-section-heading" id="record_opening_date">R
 
                             <input type="submit" value="Update" class="search-filters__submit">
 
-                            {# include hidden fields when date is errored or updated #}
-                            {{ selected_filters.items|hidden_fields_for_date_filter:form }}
```
ababic, line 95: I am SO LOST :( What was this doing before, and how come we're removing it now?

templates/search/blocks/search_filters.html
```diff
@@ -113,7 +135,7 @@ <h4 class="search-filters__accordion-section-heading" id="record_opening_date">R
             {% endif %}
         </div>
 
-        {{ 'filter_keyword collection topic level closure opening_start_date opening_end_date catalogue_source held_by type country location'|include_hidden_fields:form }}
+        {{ 'catalogue-filters-form'|include_hidden_fields:form }}
```
ababic, line 138: @TNA-Allan A custom filter just isn't the right device here. This should really be [simple tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#simple-tags), and we need to find a way for these values to be specified on the form instance, so that they aren't hardcoded in the template.

templates/search/blocks/search_results__selected-filters.html
```diff
@@ -7,7 +7,7 @@ <h2 class="sr-only">Selected filters</h2>
         {% for field_name, selected_options in selected_filters.items %}
```
ababic, line 7: Can `selected_filters.items` not give us values that are fit for direct display? The template shouldn't need to have all of these special cases for individual fields.

templates/search/blocks/search_results_hero.html
```diff
@@ -8,7 +8,7 @@ <h1 class="search-results-hero__heading">{{search_tab|search_title}}</h1>
         {{ form.q }}
         
         {% if search_tab == searchtabs.CATALOGUE.value or search_tab == searchtabs.WEBSITE.value %}
-            {{ 'q'|include_hidden_fields:form }}
+            {{ 'catalogue-search-form'|include_hidden_fields:form }}
```
ababic, line 11: I've mentioned this API already. Just highlighting here so that it isn't missed.

templates/search/blocks/search_sort_and_view_options.html
```diff
@@ -11,7 +11,7 @@ <h2 class="search-sort-view__heading">
         {{ form.sort_by.errors }}
         {{ form.sort_by }}
 
-        {{ 'sort_by'|include_hidden_fields:form }}
+        {{ 'catalogue-sort-form-desktop'|include_hidden_fields:form }}
```
ababic, line 14: I've mentioned this API already. Just highlighting here so that it isn't missed.

templates/search/catalogue_search.html
```diff
@@ -23,7 +23,8 @@
 {% endblock %}
 
 {% block content %}
-    {{ form.non_field_errors }}
+    {# commented for future requirements #}
+    {% comment %} {{ form.non_field_errors }} {% endcomment %}
```
ababic, line 27: We shouldn't be doing anything with non-field errors here. Why is this change needed?

etna/core/fields.py
TNA-Allan: Have updated the PR ticket. This is come from DF-702 - https://national-archives.atlassian.net/browse/DF-702?focusedCommentId=49550.

ababic (changes requested):
Hi @TNA-Allan, I've done what I can with review here. The general gist is: This additional complexity has highlighted just how fragile and inflexible the current form rendering is, and we need to take a step back and think about adjusting things properly to make things more dynamic, and have 'configuration' in the right place.

etna/search/forms.py
TNA-Allan: Done. Thanks.

TNA-Allan: Further to comment below -
> @ababic , I've update the comments in the ticket for the non field error. The error is not specific to a particular field, but a set of fields. I've changed to reflect the same code as `invalid_date_range` but I know think it should be made separate if at some point we want to shift errors to another location where it can be distinguished for with set of fields its related to.

Should I revert this back to individual codes here?

templates/search/catalogue_search.html
TNA-Allan: Same comment as above on non -field errors and mention in the ticket.

templates/search/blocks/search_filters.html
TNA-Allan: The doc strings in the method should be helpful.

templates/search/blocks/search_filters.html
TNA-Allan: > A custom filter just isn't the right device here. This should really be [simple tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#simple-tags), and we need to find a way for these values to be specified on the form instance, so that they aren't hardcoded in the template

Perhaps would it be bound field methods, but I'm not sure how to go about it with the complexities for this specific page with multiple forms and their behaviours (and upcoming tickets to reflect that).

etna/search/templatetags/search_tags.py
TNA-Allan: > Hi @TNA-Allan, I've done what I can with review here. The general gist is: This additional complexity has highlighted just how fragile and inflexible the current form rendering is, and we need to take a step back and think about adjusting things properly to make things more dynamic, and have 'configuration' in the right place.

Agree.

> A custom filter just isn't the right device here. This should really be [simple tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#simple-tags), and we need to find a way for these values to be specified on the form instance, so that they aren't hardcoded in the template

Perhaps would it be bound field methods, but I'm not sure how to go about it with the complexities for this specific page with multiple forms and their behaviours (and upcoming tickets to reflect that).

TNA-Allan:
> Hi @TNA-Allan, I've done what I can with review here. The general gist is: This additional complexity has highlighted just how fragile and inflexible the current form rendering is, and we need to take a step back and think about adjusting things properly to make things more dynamic, and have 'configuration' in the right place.

Agree.

> A custom filter just isn't the right device here. This should really be [simple tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#simple-tags), and we need to find a way for these values to be specified on the form instance, so that they aren't hardcoded in the template

Perhaps would it be bound field methods, but I'm not sure how to go about it with the complexities for this specific page with multiple forms and their behaviours (and upcoming tickets to reflect that). 

Please then can you help.

etna/search/templatetags/search_tags.py
TNA-Allan: Some how, I've lost this comment and add it again - may have linked to some other comment below while I tried to delete it.

templates/search/blocks/search_results_hero.html
TNA-Allan: Same comment in relation to `hidden_fields_for_date_filter` and `bound field methods`

templates/search/blocks/search_sort_and_view_options.html
TNA-Allan: Same comment in relation to `hidden_fields_for_date_filter` and `bound field methods`

etna/search/views.py
TNA-Allan: @ababic  Its not a validation. Its where the filter labels are applied. 
see 
`# prepare filter labels for date fields
        for date_item in [
            ("opening_start_date", "Record Opening from: %d-%m-%Y"),
            ("opening_end_date", "Record Opening to:  %d-%m-%Y"),
            ("created_start_date", "Date from: %d-%m-%Y"),
            ("created_end_date", "Date to: %d-%m-%Y"),
        ]:`

etna/search/templatetags/search_tags.py
TNA-Allan: Similar to comment -

> A custom filter just isn't the right device here. This should really be [simple tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#simple-tags), and we need to find a way for these values to be specified on the form instance, so that they aren't hardcoded in the template

Perhaps would it be bound field methods, but I'm not sure how to go about it with the complexities for this specific page with multiple forms and their behaviours (and upcoming tickets to reflect that).

templates/search/blocks/search_filters.html
ababic: @TNA-Allan What we really need here is a new `DateTimeRangeField` field class that has the open and start dates as 'subfields' and built-in validation for the combined values.

etna/search/tests/test_views.py
TNA-Allan: same reason to comment on separate test to each range.

templates/search/blocks/search_results__selected-filters.html
TNA-Allan: Nope, both servers its own purpose.

etna/search/tests/test_forms.py
ababic: Same answer on separate comment

etna/search/tests/test_views.py
ababic: This IS an end-to-end test (handling a request), so happy enough for this to remain as is.

etna/search/tests/test_views.py
```diff
@@ -519,9 +529,10 @@ def test_render_invalid_date_range_message(self):
             "<li>Try removing any filters that you may have applied</li>", content
         )
         self.assertSearchWithinOptionRendered(content)
-        self.assertIn(
-            "<li>There is a problem. Start date cannot be after end date.</li>", content
+        content_error_message_count = content.count(
+            f'<ul class="errorlist nonfield"><li>{CUSTOM_ERROR_MESSAGES.get("invalid_date_range")}</li></ul>'
```
ababic, line 533: Could we keep this simple and just test for the list item with hardcoded text? I know hardcoding the test is slightly more fragile, but the value is hardly likely to change (and if it did, the error would be expected)

etna/search/forms.py
ababic: Why even ask them? You certainly know how to make like difficult for yourself :)

templates/search/blocks/search_filters.html
ababic: Something like `{% render_form_fields_as_hidden form=form exclude='sort_by' %}` would be far better!

templates/search/blocks/search_filters.html
ababic: `{% render_form_fields_as_hidden form include='sort_by field_1 field_2' %}` would be a nicer interface here.

ababic:
@TNA-Allan Can you please make a start on the `{% render_form_fields_as_hidden %}` tag? It should have `exclude` and `include` options to allow you to specify field names one way or the other.

etna/search/templatetags/search_tags.py
ababic: I've hinted at this in other comments. If this is a simple tag that accepts either an `exclude` or `include` argument, you can figure out the real field list by looking at `form.fields` and subtracting from that as necessary.

etna/search/views.py
ababic: Oh, I see. You've removed this and decided to do it in the template instead then? That feels very backwards to me. You should revert the change and have the labels be prepared in the right format for the template to pick up.

templates/search/blocks/search_filters.html
ababic: How are you getting on with the custom DateTimeRangeField @TNA-Allan?

templates/search/blocks/search_results_hero.html
ababic: `{% render_form_fields_as_hidden form=form exclude='q' %}` would be much better

templates/search/catalogue_search.html
ababic: Like I've said elsewhere, errors relating to field input are field errors. We shouldn't be changing what non field errors are doing.

etna/search/tests/test_forms.py
TNA-Allan: Done. [here](https://github.com/nationalarchives/ds-wagtail/pull/991/commits/f17e917ada03a73c8a6062a5abaadf9156cf579d).

templates/search/blocks/search_filters.html
TNA-Allan: > @TNA-Allan What we really need here is a new DateTimeRangeField field class that has the start and end dates as 'subfields' and built-in validation for the combined values.

Is this usable with out customised the date form component?

etna/search/tests/test_views.py
TNA-Allan: Done. [here](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/tests/test_views.py#L629-L632). Also updated the ul with id for each set of date as they are non field errors.

TNA-Allan:
> @TNA-Allan Can you please make a start on the `{% render_form_fields_as_hidden %}` tag? It should have `exclude` and `include` options to allow you to specify field names one way or the other.

@ababic how will this cover [here-include_hidden_fields](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/templatetags/search_tags.py#L93-L95)  and [here-test_render_invalid_date_range_all_dates](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/tests/test_views.py#L641-L643)

templates/search/blocks/search_filters.html
TNA-Allan: see comment below for `render_form_fields_as_hidden`

etna/search/templatetags/search_tags.py
TNA-Allan: > I've hinted at this in other comments. If this is a simple tag that accepts either an exclude or include argument, you can figure out the real field list by looking at form.fields and subtracting from that as necessary.

I am aware of the form field as its implemented [here ](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/templatetags/search_tags.py#L88-L106) . raised a query below 

> @ababic how will this cover [here-include_hidden_fields](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/templatetags/search_tags.py#L93-L95) and [here-test_render_invalid_date_range_all_dates](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/tests/test_views.py#L641-L643)

templates/search/blocks/search_filters.html
TNA-Allan: same as 

> @ababic how will this cover [here-include_hidden_fields](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/templatetags/search_tags.py#L93-L95) and [here-test_render_invalid_date_range_all_dates](https://github.com/nationalarchives/ds-wagtail/blob/f17e917ada03a73c8a6062a5abaadf9156cf579d/etna/search/tests/test_views.py#L641-L643)

templates/search/catalogue_search.html
TNA-Allan: @ababic isn't a combination of fields a non field error?

> Same comment as above on non -field errors and mention in the ticket.

> Currently Date validation error - "There is a problem. Start date cannot be after end date." is tagged to a field opening_start_date, whereas this is a non field error. Moved it to non field error. The error is not specific to a particular field, but a set of fields.

> We have non field errors displayed at the top of the page (commented out for now), whereas we want the user to to views those errors at the Filter location. 

Would you like to take this discussion in the ticket?

etna/search/templatetags/search_tags.py
```diff
@@ -61,32 +61,29 @@ def query_string_exclude(
     return query_dict.urlencode()
 
 
-@register.filter
-def include_hidden_fields(form_id_or_visible_field_names, form) -> str:
+@register.simple_tag
+def render_form_fields_as_hidden(form, exclude="", exclude_on_error="") -> str:
```
ababic, line 65: What is `exclude_on_error` for? This tag is simply rendering hidden inputs. Any validation errors will be surfaced next to the editable versions of those fields. Remember why we're using this tag in the first place: we have multiple form elements  with different selections of visible/editable fields, yet when either form is posted, ALL values need to be preserved in the querystring, or they will be lost on the next request.

templates/search/blocks/search_filters.html
```diff
@@ -135,7 +135,7 @@ <h4 class="search-filters__accordion-section-heading" id="record_opening_date">R
             {% endif %}
         </div>
 
-        {{ 'catalogue-filters-form'|include_hidden_fields:form }}
+        {% render_form_fields_as_hidden form "filter_keyword collection topic level closure opening_start_date opening_end_date created_start_date created_end_date catalogue_source held_by type country location" "created_start_date created_end_date opening_start_date opening_end_date" %}
```
ababic, line 138: Like I explained earlier, the tag should have an `include` option for cases like this, allowing you supply `"q sort_by"` here, and not have to update this list here to account for every field on every form.

etna/search/views.py
ababic: @TNA-Allan just please go and look how this method is used and what those values are used for.  I realise validation is not happening here, but you're still removing code that should be here and using unnecessary conditionals in the template to render the 'applied filters' bit. If the purpose of this method is to prepare values for the front end, then that's what is should be doing.

templates/search/blocks/search_filters.html
ababic: I would swap the values round at the view level when putting together the API request, and leave the form/fields alone.

etna/search/forms.py
TNA-Allan: @ababic 
1) The want is to **keep the existing functionality** i.e. these date errors - specifically when start and end date are entered together must appear above the respective set date fields - Created dates and Record opening dates - see ticket comment. Unfortunately the existing functionality induces a **non-field error** into a field - ex `record_opening_start_date` which I sought to update as we now have 2 sets of dates.
2) How would we then now which non field error is for which set of the date fields? I have removed the identification generated from the tag.
3) About error code in `add_error` method - I got some details for here when I did it.
https://docs.djangoproject.com/en/4.2/ref/forms/api/#django.forms.Form.add_error
> The error argument can be a string, or preferably an instance of ValidationError

3) Do you want to discuss another structure, please bring it to the ticket and concerned?

etna/search/views.py
TNA-Allan: @ababic 
I've checked the conditions for date error message prevents showing the selected date label on error. This is restored the code for this. With this the refactored condition for date labels is reverted, now the new date[ labels for created dates](https://github.com/nationalarchives/ds-wagtail/blob/ad810061e2d07b47510432f6e43272e8e8187515/etna/search/views.py#L539-L572) are added the same way as record opening dates.

templates/search/blocks/search_filters.html
TNA-Allan: @ababic 
the original code is `{{ 'filter_keyword collection topic level closure opening_start_date opening_end_date catalogue_source held_by type country location'|include_hidden_fields:form }}` which uses `@register.filter`
As you've suggested I've updated it to `@register.simple_tag` with the new name `render_form_fields_as_hidden`

The original bit handles the form fields that need to be included by excluding the ones from the list. This avoids some manual update when a new form field is introduced.

Why do we want to re-write to have the `include` param?

etna/search/templatetags/search_tags.py
TNA-Allan: @ababic I don't understand the idea of a subfield

TNA-Allan:
https://github.com/nationalarchives/ds-wagtail/pull/1038
https://github.com/nationalarchives/ds-wagtail/pull/1041

PR #1031, opened 2023-06-13
Change Docker process in preparation for AWS

ahosgood:
## About these changes

Added in a staged build in the Dockerfile to compile CSS and JS assets and then copy into the Python image. Also updated the Node version in `.nvmrc` to match that already in `.platform.app.yaml` (the one we use for deployment).

Updated the order of Docker commands to better utilise cached images during build which results in a faster repeat builds.

*EDIT 2023-06-20:* Refactored Dockerfile ready for deployment on AWS. Introduced run scripts for production (behind gunicorn) and local development (builds CSS and JS, watches for changes without the need to run separately). Migrations run on container launch.

## How to check these changes

Delete the Docker containers and images and build them again. Where the CSS and JS didn't used to be ready (you had to do them separately) the assets should be already compiled and the site should look styled.

- [ ] ~~Checked PR title starts with ticket number as per project conventions to help us keep track of changes.~~


Dockerfile
```diff
@@ -1,3 +1,17 @@
+# Generate static assets (CSS and JavaScript)
+FROM node:18.16 AS staticassets
+WORKDIR /home
+COPY package.json package-lock.json webpack.config.js ./
+RUN npm install
+COPY scripts ./scripts
+COPY sass ./sass
+RUN npx webpack --config webpack.config.js
```
ababic, line 8: Hi @ahosgood Was excited to give this a go, but things are hanging here for me when rebuilding locally (gave up after 5 mins). Any ideas?

ahosgood: No idea - is this the step it's hanging on (`npx webpack`)?

ababic: Yep!

ahosgood: That's really strange. This was meant to be the first part of getting a Docker image that could be deployed on AWS so building the assets needed to happen at this stage...

The command is run inside a container so it shouldn't be a difference in OS or environment which makes me feel like it could be more of a network issue?

Would you be able to try it by replacing the command with:

```sh
RUN npx webpack --verbose --config webpack.config.js
```

...and then share where it gets to on your machine?

ababic: Latest version is still hanging for me, unfortunately. Maybe something to do with me being on an M1 Macbook Pro?

ahosgood: Hmm... quite possibly. I'm on an Apple M2 Pro chip - I wonder if there's a difference.

Will investigate!

ahosgood: @ababic - I've now switched this to use the npm scripts which are no precompiled like npx scripts. Hopefully that means this step will rely on the `sass` library that gets installed from the `package.json`.

Really sorry to ask, but could you please give this another whirl?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,47 @@
+name: Docker
+
+on:
+  schedule:
+    - cron: '44 6 * * *'
+  push:
+    branches:
+      - main
+      - develop
+      - chore/docker-static-assets # TEMP
```
ahosgood, line 10: TODO: Remove before merge

ababic:
@ahosgood Sorry for the delay in getting back to this. I needed to be in a place where I could delete all of the existing containers and images locally, to rule out any mixups.

Things are getting a lot further this time, though it seems to be freezing at the last step now:
`RUN npm run compile`

config/settings/production.py
```diff
@@ -1,10 +1,19 @@
 from .base import *  # noqa: F401
+import os
 
-DEBUG = False
+# TODO: Temporary until the static files can be served via S3 or a CDN
+DEBUG = True
```
ahosgood, line 5: TODO

etna/images/migrations/0002_migrate_old_image_data_to_new_model.py
```diff
@@ -80,8 +80,9 @@ class Migration(migrations.Migration):
...
         # When running forwards, increment autoid to reflect additions
-        migrations.RunSQL(
-            'SELECT setval(pg_get_serial_sequence(\'"images_customimage"\',\'id\'), coalesce(max("id"), 1), max("id") IS NOT null) FROM "images_customimage";',
-            migrations.RunSQL.noop,
-        ),
+        # TODO: Figure out why this breaks...
+        # migrations.RunSQL(
+        #     'SELECT setval(pg_get_serial_sequence(\'"images_customimage"\',\'id\'), coalesce(max("id"), 1), max("id") IS NOT null) FROM "images_customimage";',
+        #     migrations.RunSQL.noop,
+        # ),
```
ahosgood, line 87: TODO

config/settings/production.py
```diff
@@ -1,10 +1,19 @@
...
 
 # Ensure that certificates are always checked by Kong client in production
 KONG_CLIENT_VERIFY_CERTIFICATES = True
 
+# TODO: Generate SECRET_KEY
+SECRET_KEY = os.getenv("SECRET_KEY", "")
+# Need to get the IP of the load balancer or reverse proxy
+ALLOWED_HOSTS = ["*"]
```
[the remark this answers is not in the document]
ahosgood: @milh0use - this needs to go in the secrets. We can talk about this in a bit.

etna/images/migrations/0002_migrate_old_image_data_to_new_model.py
ahosgood: @ababic - I think you committed this migration. Do you know why it might break when I run `poetry run python manage.py migrate` on a completely fresh Postgres instance?

ababic: @ahosgood I couldn't say for sure, but maybe it has something to do with there being no rows in the table. Now these migrations have been run everywhere that previously had data to migrate, it should be safe to lave this commented out

PR #1038, opened 2023-06-14
DF-698 (A): Remove opening date range custom validation

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/DF-698

## About these changes

The first in a series of PRs to satisfy DF-698.

## How to check these changes

## Merging PR guidance

## Deployment guidance


etna/search/forms.py
```diff
@@ -226,20 +228,25 @@ class BaseCollectionSearchForm(forms.Form):
...
+        opening_start = cleaned_data.get("opening_start_date")
+        opening_end = cleaned_data.get("opening_end_date")
+
+        if opening_start and opening_end and opening_start > opening_end:
+            # if both dates have valid values but invalid when together
+            self.add_error(
+                "opening_start_date",
+                ValidationError(
+                    "This date must be earlier than or equal to the 'to' date.",
+                    code="range_invalid",
```
TNA-Allan, line 245:
```suggestion
                    code="date_range_invalid",
```
Could the name be specific to dates? If not why?

templates/search/includes/filter.html
```diff
@@ -16,4 +16,8 @@
             <input type="submit" value="Update" class="search-filters__submit">
         </fieldset>
     </div>
+{% elif field.data and not field.errors %}
+    {% for value in field.data %}
+        <input type="hidden" name="{{ field.name }}" value="{{ value }}">
+    {% endfor %}
```
TNA-Allan, line 22: Question: Why is this here and what scenario it covers?

TNA-Allan (commented):
> I also added the following fixes/simplifications for things that cropped up in https://github.com/nationalarchives/ds-wagtail/pull/991

I believe the things cropped up also includes decisions over errors placements, error message.

> Stopped trying to be clever about when (and when not to) show the date filters and just show them always. I quickly ran into some weird cases with them hiding when they shouldn't (e.g. when using them, they hide if there are no results, making it impossible to tweak the value)

Can I know which cases please?

etna/search/views.py
```diff
@@ -526,49 +523,21 @@ def get_selected_filters(self, form: Form) -> Dict[str, List[Tuple[str, str]]]:
...
-            for dict_values in error_dict.values():
-                for item in dict_values:
-                    form_error_messages.append(item["message"])
-
         if opening_start_date := form.cleaned_data.get("opening_start_date"):
-            # if both dates have valid values but invalid when together
-            if (
-                CUSTOM_ERROR_MESSAGES.get("invalid_date_range")
-                not in form_error_messages
-            ):
```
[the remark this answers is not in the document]
ababic: Because 'opening_start_date' will only be present in `form.cleaned_data` if that value is valid. There's no need to check for range validity explicitly here.

etna/search/views.py
```diff
@@ -526,49 +523,21 @@ def get_selected_filters(self, form: Form) -> Dict[str, List[Tuple[str, str]]]:
...
+                    "Record opening from: " + opening_start_date.strftime("%d %m %Y"),
                 )
+            ]
 
         if opening_end_date := form.cleaned_data.get("opening_end_date"):
-            # if both dates have valid values but invalid when together
-            if (
-                CUSTOM_ERROR_MESSAGES.get("invalid_date_range")
-                not in form_error_messages
-            ):
```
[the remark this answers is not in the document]
ababic: Because 'opening_end_date' will only be present in `form.cleaned_data` if that value is valid. There's no need to check for range validity explicitly here.

templates/search/includes/filter.html
ababic: Previously, you had some [strange code within the date filter output](https://github.com/nationalarchives/ds-wagtail/blob/develop/templates/search/blocks/search_filters.html#L94-L95) to account for this, but this is a better solution to that problem.

etna/search/forms.py
ababic: Why would an error code be field specific? It should only really be used for debugging - in which case, you'll already know which field threw it.

etna/search/views.py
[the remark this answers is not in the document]
TNA-Allan: When either of the date fields have errors, they are not in the selected field labels
When both of them (valid) but errors together only one of them show up in the selected filters. 
So the condition when removed allows one of them of show and not both, and when condition is put both don's show in filters labels - which is the same behaviour as if any one of them have errors.
Any thoughts on this?

etna/search/forms.py
TNA-Allan: There could be range between numbers as well, so I thought of specific.

ababic:
@TNA-Allan in response to your question about cases for the date filters not showing when they should, I included one in my original statement:

> when using them, they hide if there are no results, making it impossible to tweak the value

To make that a bit clearer:

1. Do a valid search
2. Enter valid values into the date fields, using a range that does not yield any results

The fields are not available on the page to adjust the dates, when they really should be.

etna/search/forms.py
ababic: Well, that's why I changed the code to `date_range_invalid`, as it's specifically about a date range

templates/search/includes/filter.html
TNA-Allan: Ok, That bit of code was not introduced in the [PR](https://github.com/nationalarchives/ds-wagtail/pull/991/files) but I've explained and removed/adjusted them with  `render_form_fields_as_hidden` for the filter section of the form.

etna/search/views.py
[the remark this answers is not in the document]
ababic: > When either of the date fields have errors, they are not in the selected field labels

In my view, it's fine for one or the other to show as a 'selected filter' (and be included to the API request) if it is valid but the other is not.

> When both of them (valid) but errors together only one of them show up in the selected filters.

Yep, that's still true. We make the 'start' date invalid, so only the 'end' date is shown as a selected filter

TNA-Allan:
> @TNA-Allan in response to your question about cases for the date filters not showing when they should, I included one in my original statement:
> 
> > when using them, they hide if there are no results, making it impossible to tweak the value
> 
> To make that a bit clearer:
> 
> 1. Do a valid search
> 2. Enter valid values into the date fields, using a range that just so happens to not yield any results
> 
> The fields are not available on the page to adjust the dates, when they really should be.

Thanks for clarifying that, also this bit was covered in the [PR-991](https://github.com/nationalarchives/ds-wagtail/pull/991)

PR #1047, opened 2023-06-15
DF-731: Use the 'catalogue reference' url for records by default

ababic:
https://national-archives.atlassian.net/browse/DF-731


jamesbiggs (approved):
Nice work Andy! 🚀

PR #1058, opened 2023-06-19
DF-725: Other archives hierarchy

jamesbiggs:
## Ticket 

[Ticket: DF-725: Address issue with extra levels in hierarchy on other archives details pages](https://national-archives.atlassian.net/browse/DF-725) 

## About these changes

This ticket address an issue which created a 500 error when a user tried to view a record that was at level 8, 9, 10 or 11. This ticket resolves this issue via the following improvements:

- Adding a new constant which maps the levels for 'Other archives'.
- Re-working the logic to display Other archives levels. 
- Setting different IDs in the HTML for TNA vs Non-TNA records to be used by the CSS.
- Re-working the CSS to provide different width percentages for TNA vs Non-TNA records.
- Removing the need to add classes manually in the CSS by using :nth-child and calc() to apply the width decrements automatically instead.
- Truncating the reference numbers and archive names within the different areas of the hierarchy to avoid breaking the layout for long reference number or archive name entries. 
- Added tests to test the template tags created

## How to check these changes

The reviewer should look at a sample of records details pages (both TNA and Non-TNA) to check the compressed and detailed hierarchy are displaying correctly, percentage bars and accompanying text next to them remain on the same line for larger displays and that the layout looks intact at the different responsive breakpoints.

They should also check the archive name and reference numbers are truncating for longer reference numbers at:

- 47 characters for the detailed hierarchy view
- 27 characters for the compressed hierarchy view
- 22 characters for the next/current/previous links at the foot of the detailed hierarchy

## Merging PR guidance

## Deployment guidance


etna/records/templatetags/records_tags.py
```diff
@@ -46,6 +46,31 @@ def as_label(record_field_name: str) -> str:
...
+        return NonTNALevelKeys["LEVEL_" + str(level_code)].value
+
+
+@register.simple_tag
+def breadcrumb_items(hierarchy: list, is_tna: bool, current_item: Record) -> list:
+    """Returns breadcrumb items depending on position in hierarchy
+    Update tna_breadcrumb_levels or oa_breadcrumb_levels to change the levels displayed
+    """
+    items = []
+    tna_breadcrumb_levels = [1, 2, 3]
```
TNA-Allan, line 63: Can make use of constants, instead of hardcoding, eg LevelKeys.LEVEL_1

etna/records/templatetags/records_tags.py
```diff
@@ -46,6 +46,31 @@ def as_label(record_field_name: str) -> str:
...
+    oa_breadcrumb_levels = [1, 2, 5]
```
TNA-Allan, line 64: Better to make use of constants, instead of hardcoding, eg NonTNALevelKeys.LEVEL_1

etna/records/templatetags/records_tags.py
```diff
@@ -36,7 +36,7 @@ def record_url(
 @register.simple_tag
 def is_page_current_item_in_hierarchy(page: Record, hierarchy_item: Record):
     """Checks whether given page matches item from a record's hierarchy"""
-    return page.reference_number == hierarchy_item.reference_number
+    return page.iaid == hierarchy_item.iaid
```
TNA-Allan, line 39: Could be nice to have some test cases here, specially because of mentions like 
Re: 
> because item and sub-item seem to share the same reference number.

etna/records/templatetags/records_tags.py
```diff
@@ -46,6 +46,31 @@ def as_label(record_field_name: str) -> str:
...
     """returns level as a human readable string"""
-    return LevelKeys["LEVEL_" + str(level_code)].value
+    if is_tna:
+        return LevelKeys["LEVEL_" + str(level_code)].value
+    else:
+        return NonTNALevelKeys["LEVEL_" + str(level_code)].value
+
+
+@register.simple_tag
+def breadcrumb_items(hierarchy: list, is_tna: bool, current_item: Record) -> list:
```
TNA-Allan, line 58: It would be nice to see some test cases for this method.

etna/records/templatetags/records_tags.py
jamesbiggs: Hi @TNA-Allan, this is so that levels can be specifically chosen to be displayed in the "breadcrumb"/non-detailed view (at the top of the hierarchy)

I'm not sure using the constants will work here as that would return a string - doing this allows us to pick specific levels (e.g. level 1, 2 and 3 as we have here) 😃

TNA-Allan (changes requested):

etna/records/templatetags/records_tags.py
TNA-Allan: @jamesbiggs , you're right it does return strings :)

etna/records/templatetags/records_tags.py
TNA-Allan: @jamesbiggs , you're right it does return strings :)

etna/records/templatetags/records_tags.py
jamesbiggs: Hi Allan, I've added a test for this but not sure how useful it is as it's only really comparing one value - it also comes from the loop inside the HTML which I think might be difficult to replicate here... :)

jamesbiggs: Actually, I've updated this - ignore my comment :)

jamesbiggs:
Hi @TNA-Allan, I've addressed all the comments and made the changes 🙂 Back over to you for review! 😃

etna/records/tests/test_templatetags.py
```diff
@@ -267,3 +623,59 @@ def test_non_reference_number_url(self):
...
+                if record.iaid == current_record.iaid:
+                    self.assertEqual(
+                        is_page_current_item_in_hierarchy(current_record, record),
+                        True,
+                    )
+                else:
+                    self.assertEqual(
+                        is_page_current_item_in_hierarchy(current_record, record),
+                        False,
+                    )
```
TNA-Allan, line 644:
```suggestion
    def test_is_page_current_item_in_hierarchy(self):
page = self.tna_long_hierarchy_record_instance
        hierarchy = page.hierarchy
        for name, level_item, expected_result in (("item_level_1", hierarchy[0], False),
                                                  ("item_level_3", hierarchy[1], False),
                                                  ("item_level_6", hierarchy[2], False),
                                                  ("item_level_7", hierarchy[3], True)
                                                  ):
            with self.subTest(name):
                self.assertEqual(is_page_current_item_in_hierarchy(page, level_item),expected_result)
```
@jamesbiggs , really good to see a test case for this.
1) should avoid conditions like these `if record.iaid == current_record.iaid:` in a test case, we try on only check with the data provided 
2) ` is_page_current_item_in_hierarchy(current_record, record)`  its actually the page, item which passes through see `def is_page_current_item_in_hierarchy(page: Record, hierarchy_item: Record):`

See my suggested change for the above.

jamesbiggs: Thanks Allan, I was trying to think of a way to do this for ages 😅 Thanks for the suggestion! 😃

etna/records/tests/test_templatetags.py
```diff
@@ -267,3 +623,54 @@ def test_non_reference_number_url(self):
...
+            with self.subTest(self):
+                # We pass in the record hierarchy, is_tna value and the current record
+                # and this function then retrieves the breadcrumb items associated
+                # with that hierarchy, depending on the state of is_tna
+                for i, item in enumerate(
+                    breadcrumb_items(
+                        current_record.hierarchy, current_record.is_tna, current_record
+                    )
+                ):
+                    self.assertEqual(item.iaid, expected_result[i])
```
TNA-Allan, line 676:
```suggestion
    def test_breadcrumb_items(self):
record = self.tna_long_hierarchy_record_instance
        hierarchy = page.hierarchy
        item_level_1 = hierarchy[0]
        item_level_3 = hierarchy[1]
        expected_breadcrumb_items = [item_level_1, item_level_3, record]
        self.assertEqual(breadcrumb_items(hierarchy, page.is_tna, record), expected_breadcrumb_items)
```
@jamesbiggs , nice to see this test case.
Similar to is_page_current_item_in_hierarchy test, lets avoid looping `for i, item in enumerate( breadcrumb_items(..` 
to call breadcrumb_items as breadcrumb_items is called only once

See suggested changes for that

etna/records/templatetags/records_tags.py
TNA-Allan:
```suggestion
def breadcrumb_items(hierarchy: list, is_tna: bool, record: Record) -> list:
```

I dont think current_item is passed see `{% breadcrumb_items record.hierarchy record.is_tna record as breadcrumbs %}` it should be named as just record instead of current_item.

etna/records/templatetags/records_tags.py
```diff
@@ -54,6 +54,31 @@ def as_label(record_field_name: str) -> str:
...
+    oa_breadcrumb_levels = [1, 2, 5]
+    for record in hierarchy:
+        if record.level_code != current_item.level_code:
+            if is_tna:
+                if record.level_code in tna_breadcrumb_levels:
+                    items.append(record)
+            else:
+                if record.level_code in oa_breadcrumb_levels:
+                    items.append(record)
+    items.append(current_item)
```
TNA-Allan, line 81:
```suggestion
    items.append(record)
```

etna/records/templatetags/records_tags.py
```diff
@@ -54,6 +54,31 @@ def as_label(record_field_name: str) -> str:
...
+
+@register.simple_tag
+def breadcrumb_items(hierarchy: list, is_tna: bool, current_item: Record) -> list:
+    """Returns breadcrumb items depending on position in hierarchy
+    Update tna_breadcrumb_levels or oa_breadcrumb_levels to change the levels displayed
+    """
+    items = []
+    tna_breadcrumb_levels = [1, 2, 3]
+    oa_breadcrumb_levels = [1, 2, 5]
+    for record in hierarchy:
```
TNA-Allan, line 73:
```suggestion
    for hierarchy_record in hierarchy:
```

use `hierarchy_record`  to make it more clear in the loop.

TNA-Allan (changes requested):
@jamesbiggs, nice to see test -  a couple of suggestions for them.

etna/records/tests/test_templatetags.py
TNA-Allan: In that case, I did for 1 bit nonTna, I  would suggest add TNA/NonTNA in the similar suggested way - record_tna, record_non_tna

etna/records/templatetags/records_tags.py
jamesbiggs: By putting record here it won't add the current item which is what we want the final item in the list to be 🙂

jamesbiggs: By adding `record` here it adds a different item than expected - I think `current_item` is best to keep it clear, but I'm going to change the variable name to `current_record` for extra clarity :)

etna/records/tests/test_templatetags.py
jamesbiggs: Done 😃

TNA-Allan (approved):
Looks good in terms of BE. Thanks @jamesbiggs @matt-blair .

JohnHeeryTNA (approved):
Approving from an FE perspective

PR #1071, opened 2023-06-20
UN-736: CE soft launch prep

jamesbiggs:
Ticket URL: https://national-archives.atlassian.net/browse/UN-736

## About these changes

Changes are listed on the ticket

## Merging PR guidance

## Deployment guidance


etna/articles/blocks.py
```diff
@@ -197,6 +197,7 @@ class FeaturedCollectionBlock(SectionDepthAwareStructBlock):
     description = blocks.TextBlock(max_length=200)
     items = PageListBlock(
         "articles.ArticlePage",
+        "articles.RecordArticlePage",
```
ababic, line 203: Do these two values need to be in a list?

etna/articles/models.py
```diff
@@ -369,7 +373,7 @@ class RecordArticlePage(
     TopicalPageMixin, ContentWarningMixin, NewLabelMixin, BasePageWithIntro
 ):
     template = "articles/record_article_page.html"
-    parent_page_types = ["collections.ExplorerIndexPage"]
+    parent_page_types = ["collections.ExplorerIndexPage", "articles.ArticleIndexPage"]
```
ababic, line 376: I would just go with the following here:
```suggestion
    parent_page_types = ["articles.ArticleIndexPage"]
```

These attributes are only looked at when adding new or moving existing pages - If we don't want articles to live under the explorer index page any longer, we shouldn't include it.

jamesbiggs: Hi Andy, sorry - I just messaged you on Slack about this one 😄 


I've left the original parent/sub-page types in because when I tried doing "move" on my local after removing the original parent/sub-page type I kept getting an error saying the page didn't exist or something to that effect

etna/articles/blocks.py
jamesbiggs: I don't think so, it works as is and when I tried putting it into a list the list wasn't getting resolved and threw an error

etna/articles/models.py
jamesbiggs: NOTE: Resolved on slack - this is temporary due to the error/problem mentioned above, an update will be put in for the next deployment (Tues 27th) to remove this

etna/articles/blocks.py
jamesbiggs: https://github.com/nationalarchives/ds-wagtail/blob/develop/etna/core/blocks/page_list.py#L21-L29

ababic: So it does :) I'm having a brain-fart.

etna/articles/blocks.py
```diff
@@ -200,6 +200,7 @@ class FeaturedCollectionBlock(SectionDepthAwareStructBlock):
...
```
ababic, line 203: @jamesbiggs Can we just double-check that we're using `.specific` when pulling these values out for display (to get the actual `ArticlePage` or `RecordArticlePage`)? It only does that for you automatically when you specify a single option

ababic (approved):
Just one question RE the block rendering. Other than that, this looks good to go :)

etna/articles/blocks.py
jamesbiggs: Hey Andy - looks like specific is being used already, here's an example of the element:
image

😄

jamesbiggs: NOTE: Resolved on slack - added `.specific` to the template as it was coming up funny in preview mode 🙂

PR #1083, opened 2023-06-22
UN 737 beta site banner mobile view

krishan-torchbox:
Ticket URL: [UN 737](https://national-archives.atlassian.net/browse/UN-737)

## About these changes

Adjusted top position values and added position relative on the parent header element

## How to check these changes

To enable the beta banner go to Wagtail > Snippets > Alerts > Beta banner and check active then save
Any page can be viewed to see the beta banner

## Merging PR guidance

## Deployment guidance


janehughes:
This is working well and looks good. My only concern is that if the beta banner is removed, this will break the menu again until the css is adjusted. Could we make this more dynamic by applying a class when the beta banner is added, and apply the adjusted styles to that class? This way we don't have to rely on someone updating the css every time the banner is added/removed.

Appreciate the banner is likely to stay for a while, but worth doing just in case.

janehughes (changes requested):
This is working well and looks good. My only concern is that if the beta banner is removed, this will break the menu again until the css is adjusted. Could we make this more dynamic by applying a class when the beta banner is added, and apply the adjusted styles to that class? This way we don't have to rely on someone updating the css every time the banner is added/removed.

Appreciate the banner is likely to stay for a while, but worth doing just in case.

krishan-torchbox:
I've tested it with and without the beta banner and also with multiple banners. It appears to be ok without the need for a additional class. Also ticket UN-720 will change the burger menu to use flex not absolute position so this will need to be checked once that branch is merged @ahosgood

PR #1119, opened 2023-06-29
UN-759: Unpublished pages still showing in Featured Article

jamesbiggs:
Ticket URL: [UN-759]

## About these changes

I've chosen to just prevent pages from showing if they're not public (and by that, live) in the front-end. This is because editors can select a page as a featured article, and then make the page private/un-publish it. This then either takes the user to a password page, or to an error that the page doesn't exist. This eliminates that by checking if the selected page is public - if the page isn't public or live then it won't show.

## How to check these changes

Add a page to a featured article field, then unpublish the selected page or make it private - then view the original page where you added the featured article and see that it is missing.

## Merging PR guidance

## Deployment guidance

[UN-759]: https://national-archives.atlassian.net/browse/UN-759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/articles/article_index_page.html
+++ b/templates/articles/article_index_page.html
@@ -14,7 +14,7 @@
         <h2 class="sr-only">{% trans "Each collection insight takes an in-depth look into a group of the records we hold." %}</h2>
     </div>
 
-    {% if page.featured_article.public %}
+    {% if page.featured_article.live %}
         {% include "includes/article-spotlight.html" with page=page.featured_article.specific %}
     {% endif %}
 
--- a/templates/articles/article_index_page.html
+++ b/templates/articles/article_index_page.html
@@ -14,7 +14,7 @@
         <h2 class="sr-only">{% trans "Each collection insight takes an in-depth look into a group of the records we hold." %}</h2>
     </div>
 
-    {% if page.featured_article %}
+    {% if page.featured_article.live %}
         {% include "includes/article-spotlight.html" with page=page.featured_article.specific %}
     {% endif %}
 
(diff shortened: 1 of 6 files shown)


templates/articles/article_index_page.html
```diff
@@ -14,7 +14,7 @@
         <h2 class="sr-only">{% trans "Each collection insight takes an in-depth look into a group of the records we hold." %}</h2>
     </div>
 
-    {% if page.featured_article %}
+    {% if page.featured_article.live %}
```
TNA-Allan, line 17: > or make it private - then view the original page where you added the featured article and see that it is missing.

When its made private, its still live and can be seen. 

You might want to check a method [get_view_restrictions().exists()](https://github.com/wagtail/wagtail/blob/main/wagtail/models/__init__.py#L2474-L2482) if it exists or not - maybe something as a template tag or so?

TNA-Allan (changes requested):

templates/articles/article_index_page.html
jamesbiggs: Hey @TNA-Allan, I had a think about this, and I think just having `.live` will be fine. The ticket only asks for unpublished pages, and I think it should be down to editors to not put a page in the featured article field if they're going to make it private at some point... I'm happy with this code as is, so if you're happy with it too then this is done 😃

TNA-Allan: @jamesbiggs , I going with the view you have taken.

TNA-Allan (approved):
Looks good. Thanks @jamesbiggs

PR #1150, opened 2023-07-05
UN-785: Topic and Time Period pages to have related Focused Articles

jamesbiggs:
Ticket URL: [UN-785]

## About these changes

Re-written the `related_articles` attribute for the Topic and Time Period pages to include `ArticlePage` and `FocusedArticlePage`.

Added docstrings for a bit of explanation :) 

## How to check these changes

Go to http://localhost:8000/explore-the-collection/explore-by-topic/migration-and-citizenship/ or http://localhost:8000/explore-the-collection/explore-by-time-period/early-20th-century/ and ensure that Focused Articles are also in the articles section

## Merging PR guidance

## Deployment guidance

[UN-785]: https://national-archives.atlassian.net/browse/UN-785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/etna/collections/models.py
+++ b/etna/collections/models.py
@@ -247,7 +247,7 @@ class TopicExplorerPage(RequiredHeroImageMixin, AlertMixin, BasePageWithIntro):
         """Return a list of related pages for rendering in the related articles section
         of the page. To add another page type, add it to the `exact_type` query below.
         """
-        
+
         from wagtail.models import Page
 
         from etna.articles.models import ArticlePage, FocusedArticlePage
--- a/etna/collections/models.py
+++ b/etna/collections/models.py
@@ -245,28 +245,23 @@ class TopicExplorerPage(RequiredHeroImageMixin, AlertMixin, BasePageWithIntro):
     @cached_property
     def related_articles(self):
         """Return a list of related pages for rendering in the related articles section
-        of the page. To add another page type, import it and add it to the related_page_types list.
+        of the page. To add another page type, import it and add it to the list.
         """
 
         from etna.articles.models import ArticlePage, FocusedArticlePage
 
-        related_page_types = [ArticlePage, FocusedArticlePage]
         page_list = []
 
-        for page_type in related_page_types:
-            pages = (
+        for page_type in [ArticlePage, FocusedArticlePage]:
+            page_list.extend(
                 page_type.objects.exclude(pk=self.featured_article_id)
                 .filter(pk__in=self.related_page_pks)
                 .live()
                 .public()
                 .select_related("teaser_image")
             )
-            if pages:
-                page_list.extend(pages)
 
-        page_list.sort(key=lambda x: x.first_published_at, reverse=True)
-
-        return page_list
+        return sorted(page_list, key=lambda x: x.first_published_at, reverse=True)
 
     @cached_property
     def related_record_articles(self):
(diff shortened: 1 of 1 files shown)


etna/collections/models.py
```diff
@@ -244,14 +244,22 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
         return (
-            ArticlePage.objects.exclude(pk=self.featured_article)
+            Page.objects.exclude(pk=self.featured_article.pk)
             .live()
             .public()
+            .exact_type(ArticlePage, FocusedArticlePage)
             .filter(pk__in=self.related_page_pks)
             .order_by("-first_published_at")
+            .specific()
             .select_related("teaser_image")
```
ababic, line 263: I don't think this `select_related()` statement will work from here, since `Page` does not have a `teaser_image` field. It's probably best to remove it all together.

etna/collections/models.py
```diff
@@ -400,14 +408,22 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
```
ababic, line 427: Same here :)

etna/collections/models.py
```diff
@@ -244,14 +244,22 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+        of the page. To add another page type, add it to the `exact_type` query below.
+        """
+
+        from wagtail.models import Page
+
+        from etna.articles.models import ArticlePage, FocusedArticlePage
 
         return (
-            ArticlePage.objects.exclude(pk=self.featured_article)
+            Page.objects.exclude(pk=self.featured_article.pk)
```
ababic, line 256: I believe this will raise an AttributeError if `self.featured_article` is `None` (which it can be)

ababic: What you could do though, is `self.featured_article_id`

jamesbiggs: Hi Andy, I tried removing a featured article on a page and it seems to have worked :)

etna/collections/models.py
ababic: Actually, if you wanted to keep things nice and efficient here, you could consider querying each model separately, sticking everything into a list, then sort that list in Python. I think you might see some performance gains by adding `prefetch_related("teaser_image__renditions")` to each query, too.

jamesbiggs: Okay! I wasn't sure which was better (if you see the commit history you'll see I've re-done it 3 times 😆 )
If that's the best then I'll do that :)

etna/collections/models.py
ababic: Honestly, I would use `self.featured_article_id` here. It's something Django provides for free for every FK field, and the value is always going to be `None` or and `int` - which is what we want.

etna/collections/models.py
```diff
@@ -245,22 +245,22 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
-            .exact_type(ArticlePage, FocusedArticlePage)
-            .filter(pk__in=self.related_page_pks)
-            .order_by("-first_published_at")
-            .specific()
-        )
+        related_page_types = [ArticlePage, FocusedArticlePage]
+        page_list = []
+
+        for page_type in related_page_types:
+            pages = page_type.objects.exclude(pk=self.featured_article_id).filter(pk__in=self.related_page_pks).live().public().prefetch_related("teaser_image__renditions")
```
ababic: NOTE: We'll want to keep the `select_related("teaser_image")` if going back to specific page type queries :)

etna/collections/models.py
```diff
@@ -400,16 +413,29 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+
+        for page_type in related_page_types:
+            pages = (
+                page_type.objects.exclude(pk=self.featured_article_id)
+                .filter(pk__in=self.related_page_pks)
+                .live()
+                .public()
+                .select_related("teaser_image")
+            )
+            if pages:
```
ababic, line 433: The `if` here will actually be introducing an additional `exists()` query here, which we don't want. `page_list.extend(pages)` will happily handle an empty result, so the conditional can be ditched.

etna/collections/models.py
```diff
@@ -400,16 +413,29 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+        page_list = []
+
+        for page_type in related_page_types:
+            pages = (
+                page_type.objects.exclude(pk=self.featured_article_id)
+                .filter(pk__in=self.related_page_pks)
+                .live()
+                .public()
+                .select_related("teaser_image")
+            )
```
ababic, line 432: I would shorten this whole thing to:
```suggestion
        page_list = []
        for page_type in [ArticlePage, FocusedArticlePage]:
            page_list.extend(
                page_type.objects.exclude(pk=self.featured_article_id)
                .filter(pk__in=self.related_page_pks)
                .live()
                .public()
                .select_related("teaser_image")
                .prefetch_related("teaser_image__renditions")
            )
            return sorted(page_list, key=lambda x: x.first_published_at, reverse=True)
```

jamesbiggs: Hi Andy, thanks for all these suggestions :) Do I still need the `prefetch_related` as well as `select_related`?

ababic (approved):
Approved. Though, you seemed to have lost the `prefetch_related("teaser_image__renditions")` line that was in there at one point. Maybe fire up django debug toolbar locally and see what the impact is on query numbers / page speed?

PR #1151, opened 2023-07-05
DF-746: Website search: Page type label tweaks

ababic:
Ticket URL: https://national-archives.atlassian.net/browse/DF-746

## About these changes

Implements the changes requested by Charlie from [this comment](https://national-archives.atlassian.net/browse/DF-746?focusedCommentId=52403).

## How to check these changes

Head to http://0.0.0.0:8000/search/website/ locally. 

The 'Page type' filter should now be labelled 'Format', and the option labels should reflect those included in comment. 

When applied, 'selected filters' should appear at the top of the page, prefixed with "Format:" instead of "Page type:".

In results (both here and on http://0.0.0.0:8000/search/featured/), the "Page type" label should also have changed to "Format"

## Merging PR guidance

## Deployment guidance


config/settings/base.py
```diff
@@ -283,8 +283,12 @@
 # By default, a version of the model's verbose_name is used (with " page"
 # removed from the end).
 PUBLIC_PAGE_TYPE_LABEL_OVERRIDES = {
-    "articles.articlepage": "Story",
+    "articles.articlepage": "The story of",
+    "articles.focusedarticlepage": "Focus on",
     "articles.recordarticlepage": "Record revealed",
+    "collections.topicexplorerpage": "Explore by topic",
+    "collections.timeperiodexplorerpage": "Explore by time period",
+    "collections.highlightgallerypage": "In pictures",
```
jamesbiggs, line 291: SUGGESTION: On this bit, could we not use the `title_label` that the pages have? We could add this to the other page types (e.g. topic explorer, highlight gallery, etc). Story of, Focus on, and Record revealed are already on these pages. Might make things easier?

jamesbiggs: We would need to make these title case of course, which I meant to do at some point anyways - currently they're all caps in the BE, but really we should be capitalising them in the FE :)

jamesbiggs (approved):
Hi Andy, nice work, just a possible suggestion around how we're getting the names of the page labels that might be a better way of getting it. Otherwise happy to approve this! 😃

config/settings/base.py
ababic: Okay, leave it with me :)

ababic: Okay, I've managed to come up with something that works. I had to come up with a way to only show the labels in places when desired (for example, we don't want the 'Explore by topic' label to output for every item on the topic list page), hence why it took so long.

jamesbiggs (changes requested):
Hi Andy, nice work, just two things:

The label on http://localhost:8000/explore-the-collection/stories/ has disappeared on the top cards (editor selected), but is still showing on the auto-generated cards
image
image

And then the filters on the search page are showing this:
image

etna/search/tests/test_native_website_search.py
```diff
@@ -231,14 +231,14 @@ def test_pagination(self):
             html=True,
         )
 
-    def test_filter_by_page_type(self):
+    def test_filter_by_format(self):
         """
-        Tests the view with only the 'page_type' filter applied.
+        Tests the view with only the 'formt' filter applied.
```
jamesbiggs, line 236:
```suggestion
        Tests the view with only the 'format' filter applied.
```

templates/articles/article_index_page.html
jamesbiggs, line 1:
```
<div class="container">
 {% include_block page.featured_pages with show_type_label=True %}
</div>```

on lines 21-23 (I can't seem to do an in-line comment?!)

jamesbiggs (approved):
Nice work Andy! 😃🚀

PR #1138, opened 2023-07-03
DF-761-fixing colours for feedback mechanism

JohnHeeryTNA:
Ticket URL: https://national-archives.atlassian.net/browse/DF-761

## About these changes

Change of colours for feedback mechanism

## How to check these changes

Check visually the colours of the feedback mechanism

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly before handing over to reviewer.
- [x] Checked PR title starts with ticket number as per project conventions to help us keep track of changes.
- [x] Ensured that PR includes only commits relevant to the ticket.
- [x] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

## Merging PR guidance

Follow [docs\developer-guide\contributing.md](https://nationalarchives.github.io/ds-wagtail/developer-guide/contributing/)

## Deployment guidance

Follow [docs\infra\environments.md](https://nationalarchives.github.io/ds-wagtail/infra/environments/)

PR #1127, opened 2023-06-30
Initial concept of reusable components

ahosgood:
## About these changes

This is to introduce the use of reusable components, starting with the hero (a.k.a. "highlight intro" as it is currently named).

## How to check these changes

Topic pages such as `/explore-the-collection/explore-by-topic/international-trade-and-affairs/` should have the new hero banner which should be nearly identical to the existing banner but with fewer layout issues and sets us up to start including more components.

## Before assigning to reviewer, please make sure you have

- [x] Checked things thoroughly before handing over to reviewer.
- [ ] ~~Checked PR title starts with ticket number as per project conventions to help us keep track of changes.~~
- [ ] ~~Ensured that PR includes only commits relevant to the ticket.~~
- [x] Waited for all CI jobs to pass before requesting a review.
- [x] Added/updated tests and documentation where relevant.

PR #1168, opened 2023-07-10
Reinstate frontend styles, fix issues, update footer classes

ahosgood:
https://national-archives.atlassian.net/browse/UN-742

## About these changes

Reinstates updates from https://github.com/nationalarchives/ds-wagtail/pull/1127 (undone in https://github.com/nationalarchives/ds-wagtail/pull/1138) but fixes issues:

- `tna-footer` namespace updated to `etna-footer` so as not to cause conflicts
- Images requiring `display: inline-block;` fixed
- Focus styles fixed

This should then allow us to move forward with making a first stable version (`v1.0.0`) of `@nationalarchives/frontend`

Also added `load-path` parameter into the SASS compilation to allow import not to have to use relative `../node_modules` path.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "7.1.0",
       "license": "ISC",
       "dependencies": {
+        "@nationalarchives/frontend": "^0.1.8-prerelease",
         "jest": "~29.5",
         "jquery": "~3.7",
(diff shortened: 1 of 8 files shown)


jamesbiggs (approved):
Cheers AJ! 😃

PR #1218, opened 2023-07-27
Add healthcheck route for Docker

ahosgood:
Adding a `/healthcheck/` URL that simply returns a `200` response with "OKAY" gives us a way to query the health of the container.

We can fiddle with the options available (https://docs.docker.com/engine/reference/builder/#healthcheck) once this is in AWS.


jamesbiggs (approved):
This is working so happy to approve it, but a question for my own understanding - what is the use of this?

ahosgood:
@jamesbiggs - this adds an endpoint (http://localhost:8000/healthcheck or https://beta.nationalarchives.gov.uk/healthcheck) which literally just responds with a `200` status code and a body of "okay".

If I visit that URI and get a `200` with "okay" then I know the app is up and ready to respond to requests. If I get anything else (`400`, `404`, `500` etc.) then I know something is up.

Docker can use this to periodically check on the health of the application and restart it if it doesn't get the `200` it was expecting.

We can also potentially use this in production to do a blue/green deployment - we start up the new version and as soon as it's ready (responds on `/healthcheck` with a `200`) we switch the previous instance of the application with the new version with hopefully as little downtime as possible.

PR #1230, opened 2023-07-31
CHORE: Fix CI running twice on main

jamesbiggs:
## About these changes

Removed step in CI where it ran if it was a push to `main` or `release/*`

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,10 +1,6 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - main
-      - "release/**"
   pull_request:
 
 env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 on:
-  pull_request:
+  push:
 
 env:
   PYTHON_VERSION: "3.11"


.github/workflows/main.yml
```diff
@@ -1,10 +1,6 @@
 name: CI
```
ahosgood, line 1: If this only runs on a pull request then maybe we should update the name and the filename of this workflow?

Although, should we keep running this on pushes to `develop` to ensure the working branch is always healthy?

jamesbiggs: Maybe instead of just `pull-request` we do `push` to branches `develop` and `main`? I think we should have CI run on any PR, to make sure we're not breaking what we push it into... But if you think we should only run it on `develop` and `main` then I am happy for that instead of every PR.

I'll change the filename to `ci.yml`?

ahosgood: Maybe we could just do it on every push to any branch? That way pull requests would be open against commits in that branch that had already run it rather than running it on a pull request as well?

I'm thinking it might be best to rename this to `ci.yml` and just have:

```yml
on:
  push:
```

Keen to see what other devs think...

jamesbiggs: Gotcha, yep, that sounds best to me!😄

PR #1232, opened 2023-08-01
Local development permissions

ahosgood:
- Moved media storage to a Docker volume
- CSS and JS development compilation doesn't start automatically unless `AUTO_START_SERVER=true` for faster startup time
- Updated docs for pulling media from platform.sh


jamesbiggs (approved):
Cheers AJ! 😃

PR #1237, opened 2023-08-01
Create two Docker instances to update dependencies

ahosgood:
This allows us to run `fab update-deps` to update all dependencies for npm and Poetry.

Two Docker instances are created and the updates happens inside them. This ensures a consistent environment for updating (identical versions of Python, Node etc.)

Behind the scenes, `fab update-deps` runs `docker-compose --profile update up -d` which starts up the two Docker instances defined in `docker-compose.yml`. Their volumes are mapped to the current working directory so they can write changes back and then they shut down after updating.

These two images are not affected by `fab build` or `fab start`.

In the future we can look to remove duplicated version definitions, possibly use nvm and maybe pyenv.

### DIFF
--- a/.docker/update-npm/Dockerfile
+++ b/.docker/update-npm/Dockerfile
@@ -1,0 +1,4 @@
+FROM node:18.16
+WORKDIR /home
+COPY package.json package-lock.json ./
+CMD ["npm", "update"]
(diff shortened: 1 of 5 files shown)


jamesbiggs (approved):
Nice work AJ! 😃 Updated both npm and poetry files and shut-down after!

PR #1242, opened 2023-08-03
UN-781: Focused Article "You may be interested in" and "More stories"

jamesbiggs:
Ticket URL: [UN-781]

## About these changes

- Added `similar_items` property to `FocusedArticlePage` to retrieve pages (`FocusedArticle`, `RecordArticle`, `Article`) with similar tags to the current page.
- Added `latest_items` property to `FocusedArticlePage` to retrieve the latest published pages (`FocusedArticle`, `RecordArticle`, `Article`).
- Added these to the HTML to display the pages.
- Created tests for the property.

## How to check these changes

Open a Focused Article page, scroll to the bottom and see that there is a section for similar items, and latest items.

The FA page must be tagged in the "article tags" section on the promote admin panel - similar items will only work if there are other pages tagged with the same tag.

## Merging PR guidance

## Deployment guidance

[UN-781]: https://national-archives.atlassian.net/browse/UN-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/etna/articles/models.py
+++ b/etna/articles/models.py
@@ -271,11 +271,11 @@ class ArticlePage(
 
         # Identify other live pages with tags in common
         related_tags = (
-            Page.objects.live()
+            Page.objects.filter(tagged_items__tag_id__in=tag_ids)
+            .exact_type(ArticlePage, FocusedArticlePage, RecordArticlePage)
+            .live()
             .public()
             .not_page(self)
-            .exact_type(ArticlePage, FocusedArticlePage, RecordArticlePage)
-            .filter(tagged_items__tag_id__in=tag_ids)
         )
 
         return tuple(
@@ -419,11 +419,11 @@ class FocusedArticlePage(
 
         # Identify other live pages with tags in common
         related_tags = (
-            Page.objects.live()
+            Page.objects.filter(tagged_items__tag_id__in=tag_ids)
+            .exact_type(ArticlePage, FocusedArticlePage, RecordArticlePage)
+            .live()
             .public()
             .not_page(self)
-            .exact_type(ArticlePage, FocusedArticlePage, RecordArticlePage)
-            .filter(tagged_items__tag_id__in=tag_ids)
         )
 
         return tuple(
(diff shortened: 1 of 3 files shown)


etna/articles/models.py
```diff
@@ -399,6 +399,62 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+            # Avoid unncecssary lookups
+            return ()
+
+        # Identify other live pages with tags in common
+        related_tags = (
+            Page.objects.live()
+            .public()
+            .not_page(self)
+            .exact_type(ArticlePage, FocusedArticlePage, RecordArticlePage)
+            .filter(tagged_items__tag_id__in=tag_ids)
```
ahosgood, line 426: Do we know if there is any optimisation to be gained by changing this order? I'm just thinking that this is run against all pages on every page load (I think?)

If it does have an impact then maybe we should be applying the filter that omits most pages at the start so that subsequent filters have less and less to sift through?

templates/articles/focused_article_page.html
```diff
@@ -45,13 +45,35 @@
         </div>
     </div>
 
-    {% if page.primary_topic or page.primary_time_period %}
-        <div class="related-content related-content--borderless">
-            <div class="container">
+    <div class="related-content related-content--borderless">
```
ahosgood, line 48: Would we always get this `<div>` regardless of whether there is anything in it?

jamesbiggs: I was thinking that @ahosgood, but decided that running the logic to check if 1 of 3 scenarios would be occurring is more than it's worth. There will _always_ be `latest_items` by nature of how it works, and there will usually be a related topic/time period too 😃

etna/articles/models.py
jamesbiggs: Hey AJ, sorry it took me a while to get back to you on this one - I _think_ what I've just pushed should make it a bit better, but tbh it's very hard to track the performance time with the Django Debug Toolbar, because it keeps changing the results. Running the same query 5 times gave me 5 very different numbers 😂

ahosgood (approved):

PR #1259, opened 2023-08-09
Allow no-interaction create-superuser

ahosgood:
Rather than running `python manage.py createsuperuser` and entering some values for development, the command `fab create-superuser` now takes values from the environment variables `DJANGO_SUPERUSER_PASSWORD`, `DJANGO_SUPERUSER_USERNAME` and `DJANGO_SUPERUSER_EMAIL` to avoid entering them manually.

The default login is:

- Username: `wagtail`
- Password: `wagtail`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,9 +33,9 @@ services:
       - DATABASE_HOST=db
       - DATABASE_NAME=postgres
       - DATABASE_USER=postgres
-      - DJANGO_SUPERUSER_PASSWORD=app
-      - DJANGO_SUPERUSER_USERNAME=app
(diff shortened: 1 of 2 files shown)


JohnHeeryTNA:
Hi @ahosgood - as long as this applies just to dev, I _think_ it would be ok but having differing user policies from dev and production might be a risk..?

ahosgood:
@JohnHeeryTNA - The command is intended to be run in dev and the environment variables wouldn't exist in production 👍🏼

jamesbiggs:
@ahosgood Looks good, but you've got some conflicts 😃

jamesbiggs (approved):

PR #1264, opened 2023-08-11
Add more components from tna-frontend

ahosgood:
## About these changes

- Updated `@nationalarchives/frontend` to `0.1.14-prerelease`
- Removed `box-sizing` override (in two places)
- Changed page language attribute from `en` to `en-GB` (International English -> British English)
- Updated components to latest version from `tna-frontend`:
    - Header (solves some accessibility issues and layout bugs)
    - Breadcrumbs
    - Warning component ("This article contains content that some people may find offensive")
- Updated grid for generic intro to the new `tna-frontend` grid system (matches header, footer and breadcrumbs)
- Removed some conflicting rules (e.g. links should have underline...no they shouldn't...okay but no underline when you hover on them...actually yes an underline on hover)
- Added `tabindex="-1"` to card images which means you can't tab to them (there is already a link in the card which you can tab to so we don't need a duplicate link) - this doesn't remove the ability to still click on the image to go to the link

Site is now a more consistent width and font sizes are slightly more consistent. Padding and alignment has improved. The new styles also increase the legibility of the fonts with improved antialiasing.

Before
Screenshot 2023-08-11 at 16 56 52

After
Screenshot 2023-08-11 at 16 56 56

Before
Screenshot 2023-08-11 at 16 57 30

After
Screenshot 2023-08-11 at 16 57 28


ahosgood:
@sdwilkes and @matt-blair - sorted out some of the layout on the search landing page. Hope this all looks good...

Before
Screenshot 2023-08-17 at 12 34 39

After
Screenshot 2023-08-17 at 12 35 54

PR #1279, opened 2023-08-23
Fix search menu item

ahosgood:
The search menu item won't work unless it is made visible across the whole site which we don't want on the beta site but was fine on the development site.


sdwilkes (approved):
Approved for testing

PR #1310, opened 2023-09-08
CHORE: Update CI

jamesbiggs:


ahosgood (approved):
This all looks good and should keep us updated with a CodeCov score 👍🏼

PR #1340, opened 2023-09-20
Updates to tooling

helenb:
Ticket URL: paste URL here

## About these changes

- Adds an option to run watch tasks for compiling JS and CSS concurrently
- Adds stylelint and eslint to the linting, including automated fixing
- All automated fixes are applied
- All fixes that needed manual intervention are applied, or the rule disabled in cases where I am wary of a change of style.
- Adds documentation about the linting
- Renames CI job currently called 'Tests / Prettier format' as this now covers stylelint and eslint too

## How to check these changes

- Check out the branch and test that the linting runs cleanly.
- Review all the code changes, and my comments, and flag if you think any might change the final styling - I have not tested these changes in the browser, as the ones I have done I have been fairly confident will have no ending impact. The only ones that might are where I have removed duplicate properties, so the final ordering of the CSS may have adjusted.

## Merging PR guidance

## Deployment guidance


.eslintrc.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+    env: {
+        browser: true,
+        es2021: true,
+        jquery: true,
```
helenb, line 5: differences to tna-frontend: we have jquery and jest to support in this repo

.stylelintrc
```diff
@@ -0,0 +1,29 @@
+{
+    "extends": [
+        "stylelint-config-standard-scss"
+    ],
+    "plugins": [
+        "stylelint-selector-bem-pattern"
+    ],
+    "ignoreFiles": ["sass/tna-toolkit/**/*.scss"],
```
helenb, line 8: ignore as discussed with AJ

.stylelintrc
```diff
@@ -0,0 +1,29 @@
...
+        "value-no-vendor-prefix": null,
+        "value-keyword-case": null,
+        "scss/dollar-variable-empty-line-before": null,
+        "scss/double-slash-comment-empty-line-before": null,
+        "selector-class-pattern": null,
+        "plugin/selector-bem-pattern": {
+            "preset": "bem"
+        },
+        "declaration-block-no-duplicate-properties": {
+            "ignore": ["consecutive-duplicates-with-different-values"]
```
helenb, line 23: This is to allow IE11 property fallbacks which we have in places

helenb: It is throwing an error message when running: "Invalid Option: Unexpected option value "{"ignore":["consecutive-duplicates-with-different-values"]}" for rule "declaration-block-no-duplicate-properties" - I am not sure why as it works and is valid: https://stylelint.io/user-guide/rules/declaration-block-no-duplicate-properties

.stylelintrc
```diff
@@ -0,0 +1,29 @@
...
+        },
+        "scss/at-extend-no-missing-placeholder": null,
```
helenb, line 25: I added this because there seems to be a widely used pattern in the codebase where instead of using `class="alert alert--warning"` we just have `class="alert--warning"` - and the BEM modifier class `@extends` the block class.

.stylelintrc
```diff
@@ -0,0 +1,29 @@
...
+        "scss/dollar-variable-pattern": null,
```
helenb, line 26: added due to the variable patterns used

.stylelintrc
```diff
@@ -0,0 +1,29 @@
...
+        "selector-id-pattern": null
```
helenb, line 27: added because django uses underscores in ids

package.json
```diff
@@ -8,9 +8,12 @@
     "compile": "npm run compile:css && npm run compile:js",
     "dev:css": "sass --load-path=node_modules --watch sass/etna.scss:templates/static/css/dist/etna.css",
     "dev:js": "webpack --config webpack.config.js --mode=development --watch",
+    "dev": "npm run dev:css & npm run dev:js",
```
helenb, line 11: For convenience: allow watching of the js and css at the same time

package.json
```diff
@@ -8,9 +8,12 @@
...
     "test": "jest --coverage",
-    "lint": "prettier \"{scripts,sass}/**/*.{scss,js}\" --check",
-    "lint:fix": "prettier \"{scripts,sass}/**/*.{scss,js}\" --write"
+    "lint:js": "eslint 'scripts/**/*.js'",
```
helenb, line 13: These have crept in by accident from when I was testing, and I can remove them - or if preferred I could add individual linting for prettier, js and css, and make the lint command call them.

package.json
```diff
@@ -8,9 +8,12 @@
...
+    "lint:js:fix": "eslint --fix 'scripts/**/*.js'",
+    "lint": "prettier --check '{scripts,sass}/**/*.{scss,js}' & stylelint 'sass/**/*.scss' & eslint 'scripts/**/*.js'",
+    "lint:fix": "prettier --write '{scripts,sass}/**/*.{scss,js}' & stylelint --fix 'sass/**/*.scss' & eslint --fix 'scripts/**/*.js'"
```
helenb, line 16: Uses `&` rather than `&&` so that if there are errors from more than just one tool, we see all of them. It does mean the task doesn't exit cleanly and the user has to use `cntrl c`

sass/etna.scss
```diff
@@ -236,12 +237,15 @@ $largest-small-device: 992px;
     }
 }
 
+/* stylelint-disable no-descending-specificity */
```
helenb, line 240: I've used disables where I don't want to risk any changes that might affect the styling result

sass/includes/_highlight-gallery.scss
```diff
@@ -77,13 +77,6 @@
         }
     }
 
-    &__richtext {
```
helenb, line 80: This was a duplicate of what we had above

sass/includes/_maintenance-503.scss
```diff
@@ -17,7 +17,7 @@
         margin: 3rem 5rem;
     }
 
-    @media only screen and (max-width: 37.5rem) {
+    @media only screen and (width <= 37.5rem) {
```
helenb, line 20: @ahosgood these are the media query changes that the automated fix added as discussed yesterday

sass/includes/_media-embed.scss
```diff
@@ -82,18 +82,14 @@
         width: calc(100% + 4rem);
     }
 
-    &__transcript-heading {
```
helenb, line 85: property moved to duplicate rule below

sass/includes/search/_search-results-hero.scss
```diff
@@ -118,17 +121,17 @@
             height: 100%;
             padding: 0.5rem;
             text-decoration: none;
-            color: #000000; // IE11
+            color: #000; // IE11
```
helenb, line 124: Here you can see that the option for duplicate properties applied in `.stylelintrc` is working as expected

scripts/src/image-viewer.js
```diff
@@ -1,3 +1,5 @@
+/* global image_source */
```
helenb, line 1: where variables are being declared in html files I have declared them as globals. Ideally they'd be passed as data attributes rather than being declared in separate files.

scripts/src/modules/analytics/article_tracking/remove_mobile_tracking.js
```diff
@@ -11,7 +11,9 @@ export default function remove_mobile_tracking() {
         item.removeAttribute("data-link");
     });
 
+    /* eslint-disable no-undef */
     article_body.removeEventListener("click", (e) => {
         mobile_section_tracking(e);
     });
+    /* eslint-enable no-undef */
```
helenb, line 18: I cannot find this function declared anywhere, but am wary of just removing this code

scripts/src/modules/focus_element_if_found.js
```diff
@@ -8,7 +8,7 @@
 const focus_element_if_found = (needle_class = "") => {
     const needle_elements = document.getElementsByClassName(needle_class);
 
-    if (!!needle_elements.length) {
+    if (needle_elements.length) {
```
helenb, line 11: The automated linter changed this, but I don't think it should be a problem. Length will either be 0 or a positive value, which will convert to a boolean.

package.json
ahosgood: The reason I used `&&` was to ensure the second command only runs after the first has succeeded.

Just using one `&` will force the command to run in the background and can't be easily quit.

Using `&&` also ensured we we only linting what had already been nicely formatted and that we wouldn't get conflicts from the two processes both trying to write changes back to the file system.

.stylelintrc
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "plugins": [
+        "stylelint-selector-bem-pattern"
+    ],
+    "ignoreFiles": ["sass/tna-toolkit/**/*.scss"],
+    "rules": {
+        "at-rule-empty-line-before": null,
+        "block-no-empty": null,
+        "declaration-empty-line-before": null,
+        "property-no-vendor-prefix": null,
```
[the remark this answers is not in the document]
ahosgood: Just out of interest, which properties do we have that are vendor prefixed?

.stylelintrc
ahosgood: Sure thing. I would like to eventually remove this because some styles have not strictly adhered enough to the BEM strategy but if we need it for now then this is fine. 👍🏼

package.json
ahosgood: See here: https://github.com/nationalarchives/ds-wagtail/pull/1340/files#r1347244698

package.json
ahosgood: I would honestly prefer it all in one command because I can't imagine many scenarios where we'd just want to run one in isolation.

sass/includes/_buttons.scss
```diff
@@ -81,10 +83,11 @@
 .response-btn {
     margin-right: 1em;
 }
+
 .thumb-image {
     width: 25px;
     height: 25px;
-    //margin-left:1em;
```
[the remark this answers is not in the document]
ahosgood: All of it will be deleted eventually so I wouldn't worry too much about it right now 👍🏼

sass/includes/_maintenance-503.scss
ahosgood: It's a really weird one. This is not the standard syntax so I have no idea why this change is being made. It's not part of Prettier, is it? I don't seem to get the same change on the `tna-frontend` repository...

scripts/src/modules/analytics/article_tracking/remove_mobile_tracking.js
ahosgood: Again, this might all be deleted or re-written eventually. Leave it in for now to save the headaches!

scripts/src/modules/focus_element_if_found.js
ahosgood: It's no issue. We don't need to cast to a boolean - like you said, the value will be evaluated as truthy or falsy anyway.

sass/includes/search/_search-results-hero.scss
ahosgood: These look like semi-transparent colour values? If they are, could we change these to a single `rgba()` definitions which work in IE11?

.stylelintrc
helenb: I saw a few being removed when the linting rules didn't have that - here are a few examples

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/tna-toolkit/bootstrap/_reboot.scss#L374-L379

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/tna-toolkit/bootstrap/_tables.scss#L170-L172

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/includes/search/_featured-search.scss#L10-L29

.stylelintrc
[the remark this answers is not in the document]
helenb: I saw a few being removed when the linting rules didn't have that - here are a few examples

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/tna-toolkit/bootstrap/_reboot.scss#L374-L379

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/tna-toolkit/bootstrap/_tables.scss#L170-L172

https://github.com/nationalarchives/ds-wagtail/blob/cfa2dd90a25764fd4d03f8749a87be8dd610bf3b/sass/includes/search/_featured-search.scss#L10-L29

sass/includes/_maintenance-503.scss
helenb: I'll investigate this one again.

package.json
helenb: I've removed the extra ones.

package.json
helenb: I'll change these ones. It does mean that linting may be slightly confusing, as it will only report css changes until you've fixed all of them, and only then report js ones - I think it might be slightly clearer if we split them out into two separate tasks.

helenb: Switched running them consecutively - @ahosgood let me know if you're happy for me to split the linting into two tasks - js and css.

sass/includes/search/_search-results-hero.scss
helenb: Done - turned out this was the only one for ie11 - all the others the linter picked up were duplicates that should not have been there - so that rule is reinstated.

.stylelintrc
helenb: I've removed this rule now as it was allowing more than just consecutive duplicate properties. ie11 issue resolved.

.stylelintrc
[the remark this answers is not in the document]
ahosgood: Okay - those are some interesting examples! 😅

The first one isn't even a supported property any more and the second was to fix an issue from 10 years ago. Both are just Bootstrap and that will be going at some point.

The third one is difficult because MSIE's grid rendering was fundamentally different from everyone else's and couldn't just be solved with an `-ms-` prefix. We'll have to look at improving that one at some point.

Thank you for the examples. We'll leave the prefixed versions in and remove the `property-no-vendor-prefix` rule for now to make things easier. 👍🏼

package.json
ahosgood: We don't even use the `dev` command - I think it needs to go. We run this script from inside the dev container: https://github.com/nationalarchives/ds-wagtail/blob/develop/dev/bin/dev

package.json
ahosgood: As long as we end up with a single `lint` command that exits a non-zero status if either command fails (no running in the background) so we can use it in the CI, I'm happy 😊

package.json
helenb: I added it as I couldn't see a way to compile the css and js at the same time, which seemed odd - I'm used to running these commands on my local machine - is there a way to run them both together in the dev container then?

package.json
helenb: I think I'll leave it as one now with the double `&&` - it's not ideal but we probably want to keep it just one command for the CI pipeline.

.stylelintrc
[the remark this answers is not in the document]
helenb: autoprefixer might be a good solution for ms-grid.

package.json
ahosgood: To compile and watch both you can just use two terminals or run one in the background like you have here with the `&` but I don't think there was a huge need to jump back and forth between developing the CSS and JavaScript without utilising what the `dev` container has to offer.

.stylelintrc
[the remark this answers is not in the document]
ahosgood: It might be if we do look to keep the grid stuff from Bootstrap. I wouldn't worry about it too much now though.

helenb:
@ahosgood I'm just checking over any open PRs in my name, and this one is still sitting around. It looks like we got all the feedback points addressed. Would you like me to get it updated against develop, and then merge it?

ahosgood:
@helenb - Do it! 😊👍🏼

helenb:
> @helenb - Do it! 😊👍🏼

@ahosgood can you approve so I can merge?

ahosgood (approved):

PR #1347, opened 2023-09-28
PROTOTYPE: What's On - Event Page

jamesbiggs:
## About these changes

This PR is for the BE integration of the Event Page.

Field spec can be found here: https://national-archives.atlassian.net/wiki/spaces/DSWhatson/pages/145883142/Event+page+specification

I added an "event sessions" InlinePanel so that we can add the sessions for an event from the API.

The Host and Speaker fields will eventually be updated with a ForeignKey to an AuthorPage, which will allow for either a link to an AuthorPage, or the fields we have here currently to be filled in. I'll also update the clean method so that it can only be one or the other.

Also created a What's On page to list events.

**Note:** Guidance to be given around the `description`, `useful_info` and `target_audience` fields. I couldn't work out from the spec but it might be that these will be part of a StreamField with some different blocks. I've left the blocks.py there just in case, and we can always swap the fields out for the StreamField instead. Ashley is on leave until Tuesday(?) I believe.

## How to check these changes

Create a What's On page under the Home page, then create Events pages under the Home page.

## Merging PR guidance

## Deployment guidance


etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    HYBRID = "hybrid", _("Hybrid")
+
+
+@register_snippet
+class EventType(models.Model):
+    """
+    This snippet model is used so that editors can add event types,
+    which we use via the event_type ForeignKey to add event types
+    to event pages.
+    """
```
ababic, line 35: @jamesbiggs How many options are there? Does this need to be a model, or would `TextChoices` do here as well? Another good reason for using `TextChoices` is when you have bits of conditional code based around the values.

ababic: If we DO indeed need a model here, we should add a `slug` field for use in public URLs

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    class Meta:
+        verbose_name = _("Event type")
+        verbose_name_plural = _("Event types")
+
+    def __str__(self):
+        return self.name
+
+
+@register_snippet
+class AudienceType(models.Model):
```
ababic, line 52: Same feedback for this model:
- Use `TextChoices` if it doesn't need to be editable, or values are used in conditional logic
- Add `slug` field if we do indeed need a model

etna/whatson/models.py
jamesbiggs: I asked about having hard-coded EventTypes or having a snippet so that editors can update/create new ones.

Currently it's:
Event
Course
Workshop
Training
Talk
Conference
Tour
Professional development

I shall add a slug! 😄

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    This model is used to add multiple audience types to event pages.
+    """
+
+    page = ParentalKey(
+        "wagtailcore.Page",
+        on_delete=models.CASCADE,
+        related_name="event_audience_types",
+    )
+
+    audience_type = models.ForeignKey(
```
ababic, line 84: If `AudienceType` did become a `TextChoices` instance, this would just need to be a `CharField` with `choices=AudienceType.choices`

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        "whatson.AudienceType",
+        on_delete=models.CASCADE,
+        related_name="audience_types",
+    )
+
+
+@register_snippet
+class AccessType(models.Model):
+    """
```
ababic, line 93: Same feedback for this model:

- Use `TextChoices` if it doesn't need to be editable, or values are used in conditional logic
- Add `slug` field if we do indeed need a model

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+
+    page = ParentalKey(
+        "wagtailcore.Page",
+        on_delete=models.CASCADE,
+        related_name="host_information",
+    )
+
+    # When we have an author page, we will add a ForeignKey to that here.
+    # The below fields will remain when we have an author page, but will be
+    # optional, if there is no page for the host/author.
```
ababic, line 148: I think we can drop this comment and just make the change when the time comes... things can have a bizarre way of changing shape, and you'll then end up with a confusing comment in your codebase :P

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+
+    host_name = models.CharField(
```
ababic, line 150:
```suggestion
    name = models.CharField(
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        max_length=100,
+        verbose_name=_("host name"),
```
ababic, line 152:
```suggestion
        verbose_name=_("name"),
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        help_text=_("The name of the host."),
```
ababic, line 153: This isn't really providing any new information, so I would ditch it.

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        blank=False,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    panels = [
+        FieldPanel("host_name"),
+        FieldPanel("host_description"),
+        FieldPanel("host_image"),
+    ]
```
ababic, line 178:
```suggestion
     description = models.CharField(
        max_length=200,
        verbose_name=_("description"),
        blank=True,
        null=True,
    )

    image = models.ForeignKey(
        get_image_model_string(),
        verbose_name=_("image"),
        null=True,
        blank=False,
        on_delete=models.SET_NULL,
        related_name="+",
    )

    panels = [
        FieldPanel("name"),
        FieldPanel("description"),
        FieldPanel("image"),
    ]
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        max_length=100,
+        verbose_name=_("host name"),
+        help_text=_("The name of the host."),
+        blank=True,
+        null=True,
+    )
+
+    host_description = models.CharField(
+        max_length=200,
+        verbose_name=_("host description"),
```
ababic, line 160: Nice consistant use of `gettext` wrapping here! If you ever need to translate the site to a different language (or reuse the code in a multi-lingual projec), you'll be in a much better position

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    )
+
+    access_type = models.ForeignKey(
+        "whatson.AccessType",
+        on_delete=models.CASCADE,
+        related_name="access_types",
+    )
+
+
+class Host(Orderable):
```
ababic, line 135: Is there ever more than one host? Maybe these fields should be added directly to the page model instead (in which case, you would keep the `host_` prefix on the fields)

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+
+class Speaker(Orderable):
+    """
+    This model is used to add speaker information to event pages.
+    """
+
+    page = ParentalKey(
+        "wagtailcore.Page",
+        on_delete=models.CASCADE,
+        related_name="speaker_information",
```
ababic, line 189:
```suggestion
        related_name="speakers",
```

ababic: might as well keep things simple here

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        blank=False,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    panels = [
+        FieldPanel("speaker_name"),
+        FieldPanel("speaker_description"),
+        FieldPanel("speaker_image"),
+    ]
```
ababic, line 224:
```suggestion
    name = models.CharField(
        max_length=100,
        verbose_name=_("name"),
        blank=True,
        null=True,
    )

    speaker_description = models.CharField(
        max_length=200,
        verbose_name=_("description"),
        blank=True,
        null=True,
    )

    image = models.ForeignKey(
        get_image_model_string(),
        verbose_name=_("image"),
        null=True,
        blank=False,
        on_delete=models.SET_NULL,
        related_name="+",
    )

    panels = [
        FieldPanel("name"),
        FieldPanel("description"),
        FieldPanel("image"),
    ]
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        verbose_name=_("session end date"),
+        null=True,
+        blank=False,
+        help_text=_("The date and time the session ends."),
+    )
+
+    panels = [
+        FieldPanel("session_start_date"),
+        FieldPanel("session_end_date"),
+    ]
```
ababic, line 257:
```suggestion
    start = models.DateTimeField(
        verbose_name=_("starts at"),
    )

   end = models.DateTimeField(
        verbose_name=_("ends at"),   
    )

    panels = [
        FieldPanel("start"),
        FieldPanel("end"),
    ]
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    )
+
+    panels = [
+        FieldPanel("speaker_name"),
+        FieldPanel("speaker_description"),
+        FieldPanel("speaker_image"),
+    ]
+
+
+class EventSession(Orderable):
```
ababic, line 227: It might actually be a good idea to drop the `Orderable` class here and make this a regular `models.Model` with the following `Meta` class:
```
    class Meta:
        verbose_name = _("session")
        verbose_name_plural = _("sessions")
        ordering = ["start"]
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+    end_date = models.DateTimeField(
+        verbose_name=_("end date"),
+        null=True,
+        blank=True,
+        help_text=_("The date and time the event ends."),
+    )
+
+    description = RichTextField(
+        verbose_name=_("description"),
+        null=True,
```
ababic, line 337: Don't use `null=True` on text values... there's not really anything to gain from it.

etna/whatson/models.py
jamesbiggs: I wondered this and apparently we can have more than one host, so I've put it in as an Orderable here. I'll remove the prefixes for most of the models, bad habit sorry! 😬

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+
+    # Booking information
+    booking_type = models.CharField(
+        max_length=20,
+        verbose_name=_("booking type"),
+        null=False,
+        blank=False,
+        default="Drop in",
+        help_text=_("The type of booking for the event."),
+    )
```
ababic, line 405: I have some worries about this one. Definitely feels like a `TextChoices` would be handy here for consistency

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+                .order_by("session_start_date")
+                .first()
+                .session_start_date
+            )
+            self.end_date = (
+                self.event_sessions.all()
+                .order_by("session_end_date")
+                .last()
+                .session_end_date
+            )
```
ababic, line 593: It's currently a little unclear to me whether `start_date` and `end_date` come from EventBrite (as mentioned elsewhere), or whether we always want to set them from the session data. If the latter, that's fine, but let's make sure there's no conflicting information further up.

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        blank=True,
+        help_text=_("The date and time the event starts."),
+    )
+
+    end_date = models.DateTimeField(
+        verbose_name=_("end date"),
+        null=True,
+        blank=True,
+        help_text=_("The date and time the event ends."),
+    )
```
ababic, line 333: If these fields are never for direct editing, it's usually clearer to add `editable=False`

```suggestion
    start_date = models.DateTimeField(
        verbose_name=_("start date"),
        null=True,
        editable=False,
    )

    end_date = models.DateTimeField(
        verbose_name=_("end date"),
        null=True,
        editable=False,
    )
```

etna/whatson/models.py
```diff
@@ -0,0 +1,633 @@
...
+        FieldPanel("lead_image"),
+        MultiFieldPanel(
+            [
+                FieldPanel("event_type"),
+                FieldPanel("start_date", read_only=True),
+                FieldPanel("end_date", read_only=True),
+                InlinePanel(
+                    "event_sessions",
+                    heading=_("Event sessions"),
+                    help_text=_("List of event sessions, ordered from first to last."),
```
ababic, line 464: Ordering doesn't matter to the editor, technically.... we'll always display them in date order

etna/whatson/models.py
ababic: If unclear, you could just remove for now and add later, once better defined

jamesbiggs: I think on this one, as we're populating it via the API - I wasn't going to make it a TextChoices or anything (or have it editable)

I was going to populate it via enums from the API, and make it a non-editable field. Do you think we should make this editor-editable?

etna/whatson/models.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+class BookingType(models.TextChoices):
+    """
+    This model is used for the Eventbrite booking type.
+    TODO: Find out what the other booking types are from Eventbrite.
+    """
+
+    DROP_IN = "drop_in", _("Drop in")
+    BOOKING = "booking", _("Booking")
```
ababic, line 26: Based on [scanning the docs](https://www.eventbrite.com/platform/api#/reference/event-teams/create/retrieve-an-event), I can't think what this would map to in Eventbrite. It feels like something we've invented. Maybe this is just something we should work out in a property method or something, depending on whether there is any booking option data saved?

etna/whatson/models.py
```diff
@@ -0,0 +1,618 @@
...
+    """
+
+    name = models.CharField(
+        max_length=255,
+        verbose_name=_("name"),
+    )
+
+    slug = models.SlugField(
+        max_length=255,
+        verbose_name=_("slug"),
```
ababic, line 54: We should add a `unique=True` to all of these `slug` fields

etna/whatson/models.py
```diff
@@ -0,0 +1,618 @@
...
+    )
+
+    class Meta:
+        verbose_name = _("Event type")
+        verbose_name_plural = _("Event types")
```
ababic, line 59: Django and Wagtail are pretty consistent in uppercasing the first character when needed, so there's no need to do that here. It's not explicitly mentioned in the Django docs to my knowledge, but there's a reason the examples for these aren't title-cased in the [reference docs](https://docs.djangoproject.com/en/4.2/ref/models/options/#label).

```suggestion
        verbose_name = _("event type")
        verbose_name_plural = _("event types")
```

etna/whatson/models.py
jamesbiggs: You're right, it most likely is something we've invented. It'd be good to have this as an option though, so I'm guessing it's down to editor-discretion. I think I may leave it out for now, then have a chat with everyone? What do you think?

ababic: Yep - this was never going to be the end of the work anyway... I think it makes sense to take time and be more certain about it

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        verbose_name=_("slug"),
+        unique=True,
+    )
+
+    class Meta:
+        verbose_name = _("Access type")
+        verbose_name_plural = _("Access types")
+
+    panels = [
+        FieldPanel("name"),
```
ababic, line 135: I think we can just remove this `panels` attribute and let it pick up `name` and `slug` automatically

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+    page = ParentalKey(
+        "wagtailcore.Page",
+        on_delete=models.CASCADE,
+        related_name="event_access_types",
+    )
+
+    access_type = models.ForeignKey(
+        "whatson.AccessType",
+        on_delete=models.CASCADE,
+        related_name="access_types",
```
ababic, line 152: The `related_name` here is for identifying `AccessTypeOrderable` objects from the `AccessType` model, e.g. 
```
AccessType.objects.filter(access_types__page=SomeEvent)
```

I would probably go with `event_access_types` for both fields here.

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+
+    panels = [
+        FieldPanel("name"),
+    ]
+
+    def __str__(self):
+        return self.name
+
+
+class AccessTypeOrderable(Orderable):
```
ababic, line 138: `EventAccessType` might make for a more sensible name here.

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+
+    access_type = models.ForeignKey(
+        "whatson.AccessType",
+        on_delete=models.CASCADE,
+        related_name="access_types",
+    )
+
+
+class Host(Orderable):
```
ababic, line 156: `EventHost` most make for a more sensible name here, as they are tied to event pages

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+
+    panels = [
+        FieldPanel("name"),
+        FieldPanel("description"),
+        FieldPanel("image"),
+    ]
+
+
+class Speaker(Orderable):
```
ababic, line 194: `EventSpeaker` might make for a more sensible name here, as they are tied to event pages

etna/whatson/models.py
jamesbiggs: I'm not entirely sure but I think there might be an ExhibitionPage to come soon, because I was thinking we could have the relation be 
`page = ParentalKey(
        "whatson.EventPage",
        on_delete=models.CASCADE,
        related_name="hosts",
    )`

instead of `wagtailcore.Page`. Is there any benefit of that though? And if there was an ExhibitonPage, would that then need to go back to `wagtailcore.Page`?

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        null=True,
+        blank=False,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    start_date = models.DateTimeField(
+        verbose_name=_("start date"),
+        null=True,
+        blank=True,
```
ababic, line 324: NOTE: There's no need to specify `blank` for non-editable fields, as `blank` has to do with form field behaviour, which is no longer applicable

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        verbose_name=_("who it's for"),
+        blank=True,
+        help_text=_("Info about the target audience for the event."),
+    )
+
+    # Venue information
+    venue_type = models.CharField(
+        max_length=15,
+        verbose_name=_("venue type"),
+        null=False,
```
ababic, line 357: `False` is always the default value for `null`, so you don't have to specify this here.

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        blank=False,
+        choices=VenueType.choices,
+        default=VenueType.IN_PERSON,
+    )
+
+    venue_website = models.URLField(
+        max_length=255,
+        verbose_name=_("venue website"),
+        null=True,
```
ababic, line 366:
```suggestion
```

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        blank=False,
+        choices=BookingType.choices,
+        default=BookingType.DROP_IN,
+        editable=False,
+    )
+
+    registration_url = models.URLField(
+        max_length=255,
+        verbose_name=_("registration url"),
+        null=True,
```
ababic, line 403:
```suggestion
```

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+
+    # Promote tab
+    short_title = models.CharField(
+        max_length=50,
+        verbose_name=_("short title"),
+        help_text=_(
+            "A short title for the event. This will be used in the event listings."
+        ),
+        blank=False,
```
ababic, line 435: `False` is always the default value for `blank`, so you don't need to specify that here

```suggestion
```

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+        obj.start_date = self.start_date
+        obj.end_date = self.end_date
+        return obj
+
+    def save(self, *args, **kwargs):
+        """
+        Set the event start date to the earliest session start date.
+        Set the event end date to the latest session end date.
+        """
+        sessions = [session for session in self.sessions.all()]
```
ababic: Technically, we should always use a `tuple` when we don't intend to alter the list. This would do the trick...

```suggestion
        sessions = tuple(self.sessions.all())
```

etna/whatson/models.py
```diff
@@ -0,0 +1,621 @@
...
+
+        self.start_date = sessions[0].start
+        self.end_date = sessions[-1].end
```
ababic: There's actually nothing in place to enforce that the last item is the one with the latest 'end' value, so we might want to be a bit more cautious here. Maybe something like this would be better in this scenario:

```
min_start = None
max_end = None
for session in self.sessions.all():
    if min_start is None or session.start < min_start:
        min_start = session.start
    if max_end is None or session.end > max_start:
        max_end = session.max_end

self.start_date = min_start
self.end_date = max_end
```

etna/whatson/models.py
```diff
@@ -0,0 +1,578 @@
...
+        ),
+        MultiFieldPanel(
+            [
+                FieldPanel("venue_type"),
+                FieldPanel("venue_website"),
+                FieldPanel("venue_address"),
+                FieldPanel("venue_space_name"),
+                FieldPanel("video_conference_info"),
+            ],
+            heading=_("venue information"),
```
ababic, line 465: Wagtail doesn't automatically capitalise the heading unfortunately, so this should probably be:
```suggestion
            heading=_("Venue information"),
```

etna/whatson/models.py
```diff
@@ -0,0 +1,578 @@
...
+        ),
+        MultiFieldPanel(
+            [
+                FieldPanel("booking_type", read_only=True),
+                FieldPanel("registration_url", read_only=True),
+                FieldPanel("registration_cost", read_only=True),
+                FieldPanel("registration_info"),
+                FieldPanel("contact_info"),
+            ],
+            heading=_("booking information"),
```
ababic, line 475:
```suggestion
            heading=_("Booking information"),
```

etna/whatson/models.py
```diff
@@ -0,0 +1,578 @@
...
+        FieldPanel("lead_image"),
+        MultiFieldPanel(
+            [
+                FieldPanel("event_type"),
+                FieldPanel("start_date", read_only=True),
+                FieldPanel("end_date", read_only=True),
+                InlinePanel(
+                    "sessions",
+                    heading=_("event sessions"),
+                    help_text=_("List of event sessions"),
```
ababic:
```suggestion
                    heading=_("sessions"),
```

etna/whatson/models.py
```diff
@@ -0,0 +1,578 @@
...
+                ),
+                InlinePanel(
+                    "speakers",
+                    heading=_("speaker information"),
+                    help_text=_(
+                        "If the event has more than one speaker, please add these in order of relevance from most to least."
+                    ),
+                ),
+            ],
+            heading=_("event information"),
```
ababic:
```suggestion
            heading=_("Event information"),
```

etna/whatson/models.py
ababic: `wagtailcore.Page` keeps things flexible (you can use the orderable for any page type just by adding `InlinePanel("hosts")` to the panels. Another benefit is that you can prefetch the data even if you are just working with `Page` objects (you don't need specific objects), e.g.:

`events = Page.objects.type(EventPage, ExhibitionPage).prefetch_related("hosts", "speakers")`

jamesbiggs: Coolio - that's what I thought. I shall leave it as is! 😃

etna/whatson/models.py
ababic: Yes, we can use the same pattern there too :)

ababic (approved):
Nice work @jamesbiggs! We'll definitely want some unit tests to cover off the custom validation / save / revision-related logic here, but no reason why that can't be added at a later time.

PR #1353, opened 2023-10-03
Fix featured article hero colours, column widths

ahosgood:


jamesbiggs (changes requested):
Hi AJ, nice work, just two changes:

image

The author info `li` needs `text-decoration: none;`

and

image

The intro text should have a bit of a margin between itself and the title/image like this:
image

PR #1354, opened 2023-10-03
Styles for the non-js version of the what's on listing filters

helenb:
[Ticket URL: paste URL here](https://national-archives.atlassian.net/wiki/spaces/DSWhatson/pages/151420936/What+s+on+listing+filters+spec)

Design: https://www.figma.com/proto/pcIxOpwvJ1YRYbt9lbeKUG/TNA-Etna---Design-system?node-id=5218-74239

## About these changes
These changes add the filter styles for the what's on listing, along with the pink header area
The styles for the breadcrumb, title and intro follow existing patterns in the site rather than following the design.

As we agreed not to use a detail / summary element for the dropdowns for the non-js version, the design has been adapted (in conversation with Nick)

The content is all hard-coded for now.

## How to check these changes
- Check out the `feature/filters-front-end` branch
- Create a 'what's on listing' page in your local build
- Review the styles

## Screenshots
<details>
<summary>click to expand</summary>

Desktop:
Screenshot 2023-10-03 at 15 12 36

Mobile:
Screenshot 2023-10-03 at 15 12 57

</details>

## Merging PR guidance

## Deployment guidance


helenb:
@ahosgood tagging you for reference

sass/includes/_filters.scss
```diff
@@ -0,0 +1,127 @@
+.filters {
+    $root: &;
+    display: flex;
+    gap: 2rem;
+    padding: 1.5rem 0;
+    flex-direction: column;
+
+    @media only screen and (min-width: #{$screen__md + 1px}) {
```
ahosgood, line 8: Could we import the new media queries (`@use "@nationalarchives/frontend/nationalarchives/tools/media";`) and replace these rules with queries like `@include media.on-mobile {}`?

The full list can be found here: https://github.com/nationalarchives/tna-frontend/blob/main/src/nationalarchives/tools/_media.scss#L3-L15

templates/whatson/whats_on_page.html
```diff
@@ -5,12 +5,71 @@
...
 {% block content %}
 
+  <div class="full-width-header full-width-header--pink">
+    {% include 'includes/breadcrumb.html' with classes='breadcrumb-content--reduce-space' %}
+
+    {% include "includes/generic-intro.html" with title=page.title intro=page.intro %}
+
+    <div class="tna-container">
+      <div class="tna-column tna-column--full">
+        <form class="filters" method="GET">
```
ahosgood, line 21: It would be good to specify the `action` parameter (which would be the current URL) here else any URL fragments such as `#maincontent` (if the skip link is used, for example) will persist after the form is posted and the focus will be there rather than at the top of the page (from what I can recall - you may need to test this).

templates/whatson/whats_on_page.html
```diff
@@ -5,12 +5,71 @@
...
+
+          <fieldset class="filters__fieldset">
+            <legend class="filters__super-title">When</legend>
+            <div class="filters__panel">
+              <label for="event-date" class="filters__label filters__label--top">Choose a date</label>
+              <p class="filters__help">Format: dd/mm/yyyy</p>
+              {# we need a default value for the date in iOS or it is displayed tiny #}
```
nicklee, line 28: Nice, TIL!

sass/includes/_filters.scss
```diff
@@ -0,0 +1,127 @@
...
+            justify-content: flex-start;
+        }
+
+        &::after {
+            content: "";
+            display: inline-block;
+            margin-left: 1rem;
+            width: 54px;
+            height: 30px;
+            background-image: url(/static/images/icons/toggle.svg);
```
nicklee, line 94: A future enhancement might be to create the toggle switch in pure css, this would allow us to have a subtle animation of the dot sliding across, but it feels like we could look at this when we do the JS enhancements? (see https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/static_src/sass/components/_toggle-switch.scss / https://rca-development.herokuapp.com/pattern-library/pattern/patterns/atoms/toggle-switch/toggle-switch.html (tbx/tbx), however note that this example is a bit of a frankenstein since it was updated with two "labels"…)

nicklee:
LGTM @helenb, there are a couple of comments from AJ re: form action and how we use media queries. I've also added a note about potentially enhancing the toggle with an animation when we move onto the JS enhanced version.

sass/includes/_filters.scss
helenb: Unfortunately it looks like that list doesn't have quite the same granularity which was useful for the filters - it was hard to fit them all across but I also didn't want to leave them stacked for too long, so I had a spot between 992px and 1200px where they sat horizontally but with a smaller gap.

Is there any scope, if we use the `tna-frontend` media queries, for a more granular custom breakpoint in between? If not I'll need to rejig the layout a bit, perhaps having the fields 2-across at medium breakpoints.

templates/whatson/whats_on_page.html
helenb: Ah thanks - I assumed that detail would come when the back-end is hooked in - I added the GET parameter as Leanne mentioned a need to link straight to filtered lists so I was thinking ahead with that. But it makes sense that the action would be the current URL, so I'll add that in.

sass/includes/_filters.scss
helenb: The reason I went with the svg was because I was adding it as an `after` element on the label that needs its style to be updated when the sibling element (the checkbox) is clicked or focused, and the style looked too complex to do with one element rather than two. I'm not sure off the top of my head how I'd resolve that but I'll look at your example.

templates/whatson/whats_on_page.html
helenb: Me too!

sass/includes/_filters.scss
helenb: Ah Ok I can see how it all works in that RCA example now - the label wraps the input and you have a separate element for the switch.

templates/whatson/whats_on_page.html
helenb: I've set the action to `#` for now - down the line that might become a specific anchor point on the page so the user lands at the start of the listing.

sass/includes/_filters.scss
ahosgood: Okay, let's keep it for now. Ideally we wouldn't have too many breakpoints as there then becomes a lot more scenarios to test but we might have to think about adding a bit more flexibility in the future.

ahosgood (approved):

PR #1373, opened 2023-10-12
Add basic author profile styling

ahosgood:
Ticket URL: [UN-798](https://national-archives.atlassian.net/browse/UN-798)

## About these changes

- Added markup and styling for author profile page

## How to check these changes

[Example page from develop data](http://localhost:8000/authors/thomas-paine/)

[UN-798]: https://national-archives.atlassian.net/browse/UN-798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/authors/author_page.html
```diff
@@ -1,57 +1,48 @@
...
                     {% endfor %}
                 </ul>
-            {% endif %}
-        </div>
-    </div>
-
+            </div>
+        </aside>
+        {% endif %}
+    </article>
```
jamesbiggs, line 47: For my own understanding: why put it in an article (and an aside)? I understand semantics > divs but why specifically an article?

sass/includes/_author-intro.scss
```diff
@@ -1,3 +1,26 @@
...
+
+        img {
+            width: 100%;
+            height: 100%;
+        }
+
+        @include media.on-tiny {
+            max-width: 15rem;
+            margin-bottom: 3rem;
+        }
```
jamesbiggs, line 20: Another one for my own understanding: What does `media.on-tiny` do? Smallest screensize media query?

templates/authors/author_page.html
ahosgood: MDN is really helpful when I have to remind myself of the semantics.

This profile just kind of fitted the description of the article element (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) and you should always try and use semantic elements where possible.

Again, the `<aside>` just kind of fitted the pattern of being an aside to the main "article" about the author, as kind of "related to this thing" content.

You'll see that there is also a `<header>` element in this `<article>` because headers aren't just for the page header. When used inside an article, they are the header of that article. Again, I remind myself using MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#article_header

sass/includes/_author-intro.scss
ahosgood: In the `tna-frontend`, we have four classifications of device for media queries; `large`, `medium`, `small` and `tiny`. I picked these rather than `desktop`, `tablet` and `mobile` because I don't care about the device, only the available screen size. It is meant to abstract the *real* media queries away from you.

However, I needed names for "groups" of these so I could apply styles to more than one size at a time. `mobile` was what I decided to call the `small` and `tiny`-sized devices. Maybe a poor choice but I probably need to review this in the future.

The way media queries are named and grouped is described here is a crude ASCII diagram: https://github.com/nationalarchives/tna-frontend/blob/main/src/nationalarchives/tools/_media.scss

The actual values used are defined here: https://github.com/nationalarchives/tna-frontend/blob/main/src/nationalarchives/variables/_media.scss

At a point, the screen is too narrow to squeeze the image next to the title so I stack them. The CSS above just gives some extra space when that situation occurs and sets a maximum image size so it doesn't fill the whole mobile screen.

templates/authors/author_page.html
jamesbiggs: Awesome, thank you. I never knew that about headers!

jamesbiggs (approved):
Lovely work AJ 🚀

PR #1374, opened 2023-10-13
EDEV-6 - Remove dev and move commands into root web

ahosgood:
https://national-archives.atlassian.net/browse/EDEV-6

- Removed the `dev` Docker container
- Using a new variant of the `tna-docker-django` image; `tna-docker-django-root` which has the user set to `root`
- Can now change Docker image and tag for development without affecting builds on GitHub CI which continue to use the non-root version
- Commands can now be run in the web container with `fab dev` followed by things like `dev-css` - hopefully really similar to how they are now
- Running as `root` during development means we can write changes back to the file system such as the `format` command without the changes being owned by the container user
- The admin user `admin` is created on boot with the password `admin` which should speed up development
- Add configuration for `isort`, `black` and `flake8`


jamesbiggs (commented):
Hi AJ,
I'm still getting the permissions error when running format - is there a way that you can run `sudo chown -R user:user .` after the format commands are run? That'd give the permissions back to the user we're currently on (not sure how to directly get the user though, but I've been running `sudo chown -R james:james .`) Only issue is this would need to be run outside the container, I think.

I'm also not getting images load after running `fab pull-staging-media`, I see them in the ds-wagtail-media volume but just getting blank images on the FE.

It looks like the management commands are getting run inside the container only as well, so if I do, for example, `manage startapp appname`, it creates the new app folder inside the container's folders but doesn't get replicated on my local machine so I can't see/edit the files

jamesbiggs:
```bash
WARN[0000] The "UID" variable is not set. Defaulting to a blank string. 
WARN[0000] The "GID" variable is not set. Defaulting to a blank string. 
time="2023-11-07T15:15:19Z" level=warning msg="The \"UID\" variable is not set. Defaulting to a blank string."
time="2023-11-07T15:15:19Z" level=warning msg="The \"GID\" variable is not set. Defaulting to a blank string."
```

I think the `user: "${UID}:${GID}"` isn't getting set properly on my machine

jamesbiggs:
Just tested this again in WSL and on normal Windows and I'm getting this error on both

```
failed to solve: ghcr.io/nationalarchives/tna-python-django-root:rooted-images: ghcr.io/nationalarchives/tna-python-django-root:rooted-images: not found
#5 [web internal] load metadata for ghcr.io/nationalarchives/tna-python-django-root:rooted-images
#5 ERROR: ghcr.io/nationalarchives/tna-python-django-root:rooted-images: not found

#7 [cli internal] load metadata for docker.io/pjcdawkins/platformsh-cli:latest
#7 ERROR: failed to authorize: Unavailable: error reading from server: EOF
```

jamesbiggs (approved):
After I have set my WSL user to `root`, I can now use this perfectly 👌 Same for Windows, that works without any changes needed. Happy to approve this now, but we will need to sort out the other Windows (WSL) dev laptops as well 😃

jamesbiggs:
@ahosgood before merging could you run the formatters here please? I'm getting some test errors when using this

ahosgood:
Replaced with #1743

PR #1404, opened 2023-10-24
Whats on filters - front-end

helenb:
I don't think there is a ticket for this work.

## About these changes

- Adds JavaScript filters to live reload results
- Creates a feature flag to enable testing without JavaScript
- Updates the browser history state when filters are applied with JavaScript, and makes sure the page is consistent with the url when the back and forward buttons are used

Still to come in a later MR:
- Clear filters and list of applied filters
- Accessibility considerations - e.g. an aria-live region for the results
- Hide the featured event if it is not part of the listing
- Tests for filtering with JS

## How to check these changes
To check without JavaScript:
- edit your `.env` file and add `FEATURE_DISABLE_JS_WHATS_ON_LISTING=True`, then `fab stop` and `fab start`.
- Make sure the submit button appears, and that changing the filters doesn't update the listing until you click 'submit'
To check with JavaScript
- make sure the line above is removed from your `.env` file, then `fab stop` and `fab start`
- Make sure the submit button is hidden, and that updating the filters changes the listing immediately
- Check the url updates when the listing changes
- Ensure that you can use the back button and forward button in the browser, and that the filters update as you would expect.

## Merging PR guidance

## Deployment guidance


etna/whatson/models.py
```diff
@@ -289,6 +290,20 @@ def filter_form_data(self, filter_form_cleaned_data):
...
+        # If so, return just the event listing, not the full page
+        if request.headers.get("JS-Request"):
+            print(self.get_context(request))
+
+            return render(
+                request, "includes/whats-on-listing.html", self.get_context(request)
+            )
+        else:
+            # Display whats on page as usual
+            return super(WhatsOnPage, self).serve(request)
```
jams2, line 305:
```suggestion
            return super().serve(request)
```

You should be fine to use the short form here.

ahosgood (approved):

PR #1410, opened 2023-10-26
Filtering of featured event card

helenb:
No ticket

## About these changes

The featured event card is meant to have the filtering applied, just like the other event cards, and should not be repeated in the list below.

This adds some extra queries to work with the featured card, if it has been added. It also moves more of the markup from the whats on template to the whats on listing include, so that the featured card can be updated with JavaScript.

## How to check these changes

Add a few events to a whats on listing page, and also add a featured event. Check that the filtering behaves as you expect, and that the featured event is not repeated in the listing below.

Also check the page when there is no featured event selected.

You could also test that it works without JS by setting the `FEATURE_DISABLE_JS_WHATS_ON_LISTING` to True in your `.env` file (note that you need to run `fab stop` and `fab start`)

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/whatson/models.py
+++ b/etna/whatson/models.py
@@ -305,7 +305,7 @@ class WhatsOnPage(BasePageWithIntro):
         return events
 
     # we need a separate query to check if the featured event is filtered out or not
-    def should_show_featured_event(self, filtered_events, featured_event):
+    def should_show_featured_event(self, filtered_events):
         try:
             # checks if the featured event is part of the filtered events queryset
             filtered_events.filter(id=self.featured_event_id).get()
@@ -337,11 +337,12 @@ class WhatsOnPage(BasePageWithIntro):
             else:
                 context["events"] = filtered_events
         else:
+            filtered_events = self.events
             context["events"] = self.events
         context["filter_form"] = filter_form
         if self.featured_event:
             context["show_fetured_event"] = self.should_show_featured_event(
-                filtered_events, self.featured_event
+                filtered_events
             )
 
         return context
--- a/etna/whatson/models.py
+++ b/etna/whatson/models.py
@@ -288,7 +288,8 @@ class WhatsOnPage(BasePageWithIntro):
         """
         Returns a queryset of events that are children of this page.
         """
-        return EventPage.objects.child_of(self).live().public().order_by("start_date")
+        # .exclude and check page id matches featured_event.id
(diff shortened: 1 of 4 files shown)


helenb:
@jams2 this is ready for a re-review.
@jamesbiggs you'll want to also have a look because this changes the values for the filtered events that we are using in the template.

etna/whatson/models.py
```diff
@@ -290,6 +290,15 @@ def filter_form_data(self, filter_form_cleaned_data):
 
         return events
 
+    # we need a separate query to check if the featured event is filtered out or not
+    def should_show_featured_event(self, filtered_events, featured_event):
```
jams2, line 294:
```suggestion
    def should_show_featured_event(self, filtered_events):
```

Issue: unused argument

etna/whatson/models.py
```diff
@@ -308,9 +317,18 @@ def get_context(self, request, *args, **kwargs):
...
+            if self.featured_event:
+                context["events"] = filtered_events.exclude(id=self.featured_event_id)
+            else:
+                context["events"] = filtered_events
+        else:
+            context["events"] = self.events
         context["filter_form"] = filter_form
+        if self.featured_event:
+            context["show_fetured_event"] = self.should_show_featured_event(
+                filtered_events, self.featured_event
```
jams2, line 330: Issue: in the event that `filter_form.is_valid()` evaluates to `False`, `filtered_events` will not have been declared, so this will cause an exception.

jams2, line 330: Question: is this needed? We are filtering the featured event out of the listing, so shouldn't it always be shown in the featured card style?

jams2 (changes requested):
Nice work so far @helenb , a few more comments from me.

etna/whatson/models.py
helenb: Is it best to just declare `filtered_events` to equal `self.events` in that scenario? I'm not sure if the form can be invalid as it is a simple get.

helenb: I think it is. In the template, we first render the featured event, then loop over the listing below to display the other events. We need to know whether the featured event should be filtered out or not.

helenb:
@jamesbiggs maybe you could deal with the last points on this and merge it now?

etna/whatson/models.py
jams2: Got it, this was a little confusing, thanks!

jamesbiggs (approved):
Nice work @helenb and @jams2! 😃

PR #1416, opened 2023-10-27
Include nationalarchives-frontend-django module support

ahosgood:
Add support for https://github.com/nationalarchives/tna-frontend-django

Templates defined in this repository will be used first and if none are found, the `nationalarchives-frontend-django` module is used.

`components/card.html` is taken from `templates/components/card.html` first because it is defined. The one from `nationalarchives-frontend-django` is not used.

`components/skip-link.html` doesn't have a template available in this repository so it defaults to the one in `nationalarchives-frontend-django`.


jamesbiggs (approved):
Looks good AJ! Nice use of overriding the default templates folder with a package's templates - I'd not seen that before! 😄 

Might be an idea to run `update-poetry` or something just to re-update the `poetry.lock` file - I ran it to quickly install the new package but when I did it re-added a few lines that were deleted in this one. Other than that, all good! 🚀

PR #1418, opened 2023-10-30
Tweak introduction text on articles

ahosgood:
Increase size of introduction paragraph on articles and increase column width to make more prominent.

https://national-archives.atlassian.net/browse/UN-815?focusedCommentId=59554


JohnHeeryTNA (approved):
approving change.

PR #1425, opened 2023-10-31
Add tracking for image block transcripts and tabs

ahosgood:
https://national-archives.atlassian.net/browse/UN-580

Tracking added to picture elements with a transcript when:

- opening/closing the transcript dropdown
- changing tabs within the transcript panel


JohnHeeryTNA (approved):
approving changes.

PR #1444, opened 2023-11-13
Switch data-link and data-link-type in transcript

ahosgood:
https://national-archives.atlassian.net/browse/UN-580

It was picked up in testing that these two properties were not right in the Jira ticket.

The Jira ticket has now been updated and this PR is a reflection of it.


ahosgood:
Replaced with https://github.com/nationalarchives/ds-wagtail/pull/1451

PR #1451, opened 2023-11-15
Switch data-link and data-link-type in transcript

ahosgood:
Copy of https://github.com/nationalarchives/ds-wagtail/pull/1444 for `main`


TNA-Allan (approved):
Thanks @ahosgood

PR #1458, opened 2023-11-15
[DSWO-46] Internal feedback adjustments

janehughes:
Ticket URL: https://national-archives.atlassian.net/browse/DSWO-46

## About these changes

Addresses AJ's feedback following review of the What's On landing page and Event page.

Note: I haven't addressed the breadcrumbs comment, as this probably needs some further design consideration. From local testing the position of the breadcrumbs on the event hero isn't too far from the whats on page - perhaps we need to user test before changing the design?

**Event page feedback**

* The main content ideally needs to be in an `<article>` element and the "Need to know" box should be marked up as an `<aside>` within it. Similarly, the hero (excluding the breadcrumbs) should be a `<header>` element.

* There are already helper classes for the aside which could remove all need for custom CSS in the “Need to know” box:

* The gutter on the left side of the “Booking information” (3rem) is not the same as the gutter between the columns (2rem) - is this intentional?

* The breadcrumbs in the header are much further from the header than other pages - we’d ideally want to keep them in the same position to avoid users losing their place.

* The “Book now” button has some extra styling that makes it different from other buttons on the site. I suggest we just stick with the single standard tna-button class.

**What's on landing feedback**

* I think the “tags” at the top of each event card (“Talk”, “Professional development” etc.) could do with some visually hidden text to explain what they are.

* The “Our pick” label is a weird distance from the top left corner. One of the dimensions is 1.563rem and according to the Figma file we should be working on a 4px grid. I would suggest a top of 1rem and a left of 1rem as we have done for the other cards.

* Use suggested markup for the event-details-banner on the event details page to avoid the container collapsing into the button

## How to check these changes

- [ ] Checked things thoroughly before handing over to reviewer
- [ ] Ensured that PR includes only commits relevant to the ticket

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/sass/includes/_event-meta.scss
+++ b/sass/includes/_event-meta.scss
@@ -7,7 +7,7 @@
     margin-bottom: 2rem;
 
     @include media.on-larger-than-mobile {
-        padding-left: 3rem;
+        padding-left: 2rem;
         border-left: 1px solid colour.brand-colour("light-grey");
         margin-bottom: 0;
     }
--- a/sass/includes/_event-meta.scss
+++ b/sass/includes/_event-meta.scss
@@ -3,7 +3,6 @@
 @use "@nationalarchives/frontend/nationalarchives/variables/grid" as gridVars;
 @use "@nationalarchives/frontend/nationalarchives/tools/typography";
 
-
 .event-meta {
     $root: &;
     margin-bottom: gridVars.$gutter-width;
--- a/sass/includes/_event-meta.scss
+++ b/sass/includes/_event-meta.scss
@@ -1,13 +1,15 @@
 @use "@nationalarchives/frontend/nationalarchives/tools/media";
 @use "@nationalarchives/frontend/nationalarchives/tools/colour";
+@use "@nationalarchives/frontend/nationalarchives/variables/grid" as gridVars;
 @use "@nationalarchives/frontend/nationalarchives/tools/typography";
+
 
 .event-meta {
     $root: &;
-    margin-bottom: 2rem;
+    margin-bottom: gridVars.$gutter-width;
 
     @include media.on-larger-than-mobile {
-        padding-left: 2rem;
+        padding-left: gridVars.$gutter-width;
         border-left: 1px solid colour.brand-colour("light-grey");
         margin-bottom: 0;
     }
--- a/templates/components/event_tags.html
+++ b/templates/components/event_tags.html
@@ -1,10 +1,14 @@
 {% if event_type or access_descriptor %}
     <ul class="event-tags">
         {% if event_type %}
-        <li class="event-tags__tag event-tags__tag--event-type">{{ event_type }}</li>
+        <li class="event-tags__tag event-tags__tag--event-type">
+            <span class="sr-only">Event type:</span> {{ event_type }}
+        </li>
         {% endif %}
         {% if access_descriptor and not hide_access_descriptor %}
-        <li class="event-tags__tag event-tags__tag--access">{{ access_descriptor }}</li>
+        <li class="event-tags__tag event-tags__tag--access">
+            <span class="sr-only">Accessibility:</span> {{ access_descriptor }}
+        </li>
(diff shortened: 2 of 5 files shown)


templates/includes/event_hero.html
```diff
@@ -1,5 +1,5 @@
 {% load wagtailimages_tags %}
-<div class="event-hero">
+<header class="event-hero">
```
helenb, line 2: This now means there are two `<header>` elements on the page - is that correct? I am guessing this is OK because it is wrapped in an `<article>`?

templates/whatson/event_page.html
```diff
@@ -9,154 +9,156 @@
...
-                                    <p>{{ page.start_date|date:"l j F, H:i" }}</p>
-                                </div>
-                            </li>
-                        {% endif %}
-                        {% if page.price_range %}
+    <article>
+        {% include "includes/event_hero.html" with title=page.title image=page.lead_image event_type=page.event_type access_descriptor=page.primary_access_type %}
+
+        <div class="event-details-banner">
+            <div class="tna-container tna-container--align-center tna-!--padding-vertical-s">
```
helenb, line 16: I don't think `!` is valid in a  class name

sass/includes/_event-meta.scss
```diff
@@ -7,7 +7,7 @@
     margin-bottom: 2rem;
 
     @include media.on-larger-than-mobile {
-        padding-left: 3rem;
+        padding-left: 2rem;
```
helenb, line 10: I know @ahosgood that you queried this one, but I've just double checked the designs and it is 3 rem rather than 2 rem here.

templates/components/event_tags.html
```diff
@@ -1,10 +1,14 @@
 {% if event_type or access_descriptor %}
     <ul class="event-tags">
         {% if event_type %}
-        <li class="event-tags__tag event-tags__tag--event-type">{{ event_type }}</li>
+        <li class="event-tags__tag event-tags__tag--event-type">
+            <span class="sr-only">Event type:</span> {{ event_type }}
```
helenb, line 5: good catch

helenb:
Thanks @janehughes just a few queries. Not sure if we can avoid the invalid classname if this comes from `tna-frontend` but perhaps that is one that can be fixed there in the future.

templates/includes/event_hero.html
janehughes: Yeah it's fine to have more than one, and as you say this one is separated by the article tag as well which distinguishes it from the main template header

templates/whatson/event_page.html
janehughes: I can see we're escaping the ! in the `overrides.scss` file but would be good to understand the reasoning behind this @ahosgood

ahosgood: Using an "!" in a class name is fine - I've copied this approach from GOV.UK Frontend.

`tna-!--` classes are ones that have `!important` rules so we know just by looking that the padding is defined by this class and we shouldn't need to overwrite it. If we did need to change the bottom padding (for example) then we'd change to using the `tna-!--padding-top-s` class and explicitly define the padding for the bottom.

sass/includes/_event-meta.scss
ahosgood: I raised it because the guttering in other pages and all throughout the grid system is `2rem` so it didn't make sense to do something different. I'm trying to standardise this so we don't have a whole load of spacing that varies by tiny amounts.

What might be even better is if we use the gutter definition itself:

```scss
@use "@nationalarchives/frontend/nationalarchives/variables/grid" as gridVars;

.rule {
  padding-left: gridVars.$gutter-width
}
```

janehughes: That's fair enough thanks @ahosgood - I've updated this instance but sounds like it'll need a more in depth review to replace existing rems later.

helenb (approved):

PR #1461, opened 2023-11-16
Off-black header colour for whats on pages

helenb:
[ part of this ticket](https://national-archives.atlassian.net/jira/software/projects/DSWO/boards/121?selectedIssue=DSWO-46)

## About these changes

This adds the correct header colour for the what's on pages. Note that the solution was not simply to completely remove the `tna-header--accent ` class from the header, as that left the background colour as black, whereas the designs use the off-black `#1e1e1e` which isn't currently available in `tna-frontend`. Therefore I've added another override in `etna.scss` to set the correct background class for the whats on header - this can be moved to `tna-frontend` in the future.

## How to check these changes

In your local build, test the whats on listing page, and an individual event page.

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/templates/includes/header.html
+++ b/templates/includes/header.html
@@ -2,7 +2,7 @@
 
 {% include 'includes/login-status.html' %}
 
-<header class="tna-header tna-header--accent" data-module="tna-header">
+<header class="tna-header tna-header--accent {{ classes }}" data-module="tna-header">
   <div class="tna-header__exit">
     <div class="tna-container">
       <div class="tna-column tna-column--full">
--- a/sass/etna.scss
+++ b/sass/etna.scss
@@ -594,3 +594,8 @@ Modified to match the existing site (for now)
         }
     }
 }
+
+// off-black header
+.tna-header--off-black {
+    background-color: $color--off-black;
+}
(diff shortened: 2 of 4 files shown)


templates/includes/header.html
```diff
@@ -2,7 +2,7 @@
 
 {% include 'includes/login-status.html' %}
 
-<header class="tna-header tna-header--accent" data-module="tna-header">
+<header class="tna-header tna-header--accent {{ classes }}" data-module="tna-header">
```
ahosgood, line 5: Would it be easier to use a boolean for this switch to avoid adding more classes and overrides?

```
<header class="tna-header{% '' if noAccent else ' tna-header--accent' %}" data-module="tna-header">
```

Then we could use the simpler:

```
{% include 'includes/header.html' with noAccent=True %}
```

⚠️ (untested code) ⚠️

helenb: I think thought that this would then end up with the black background, rather than the off-black one?

ahosgood: Should it not be black? It will be eventually anyway when the new header is added. For ease then, let's keep your solution.

ahosgood (approved):

PR #1470, opened 2023-11-22
DSWO-51: Format date and time ranges correctly for event pages and event cards

helenb:
Ticket URL: https://national-archives.atlassian.net/browse/DSWO-51

## About these changes

Adds the correct formatting for events according to the number of sessions, the start and end date, and the start and end time. This is applied to the event page key details bar, the event card in the listing, and the featured event card.

The whatson listing tests have also been updated to test the date range formatting. I needed to add some more events in order to test the different scenarios, which means the filter test also had to be updated.

## How to check these changes

On a local build, create a few different events, covering the following scenarios:
- One session on one date, and start and end time are exactly the same
- One session on one date, and start and end times are different
- Multiple sessions on one date
- Multiples dates

Check that the featured event, event cards in the listing, and the key details on the event page match the rules specified here: https://national-archives.atlassian.net/wiki/spaces/DSWhatson/pages/142573589/Time+date+format+display+rules

## Merging PR guidance

## Deployment guidance


helenb:
@jamesbiggs this is ready for review. @jams2 if you have time to look too, that would be appreciated.

etna/whatson/models.py
```diff
@@ -685,6 +685,34 @@ def primary_access_type(self):
         if primary_access := self.event_access_types.first():
             return primary_access.access_type
 
+    @cached_property
+    def date_time_range(self):
+        delta = self.end_date - self.start_date
```
jams2, line 690: Suggestion: move this to after the first if clause so we only calculate it if we need it

etna/whatson/models.py
```diff
@@ -685,6 +685,34 @@ def primary_access_type(self):
...
+        if (delta.days == 0) and (len(self.sessions.all()) > 1):
+            return self.start_date.strftime("%A %-d %B %Y")
+        # Event has multiple dates
+        # Eg. 1 January 2024 to 5 January 2024
+        if delta.days > 0:
+            return (
+                self.start_date.strftime("%-d %B %Y")
+                + " to "
+                + self.end_date.strftime("%-d %B %Y")
+            )
```
jams2, line 714: Suggestion: assign the format strings to variables so that they can be reused and are self-documenting. This should make this method a little easier to read at a glance.

It might be nice to use f-strings also, e.g.:

```python
f"{self.start_date.strftime(format_no_time)} to {self.end_date.strftime(format_no_time)}"
```

jams2 (commented):
The code looks good to me, just a couple of suggestions.

etna/whatson/models.py
```diff
@@ -685,6 +685,34 @@ def primary_access_type(self):
         if primary_access := self.event_access_types.first():
             return primary_access.access_type
 
+    @cached_property
+    def date_time_range(self):
+        delta = self.end_date - self.start_date
+        # One session on one date where start and end times are the same
+        # return eg. Monday 1 January 2024, 19:00
+        if (self.start_date == self.end_date) and (len(self.sessions.all()) == 1):
```
jams2, line 693: Question: `start_date` and `end_date` are `DateTimes`, not `Dates` - it's possible that this comparison could fail if the times are different. What effect will this have?

etna/whatson/models.py
```diff
@@ -685,6 +685,34 @@ def primary_access_type(self):
...
+            return self.start_date.strftime("%A %-d %B %Y, %H:%M")
+        # One session on one date where there are values for both start time and end time
+        # eg. Monday 1 January 2024, 19:00–20:00 (note this uses an en dash)
+        if (delta.days == 0 and delta.seconds > 0) and (len(self.sessions.all()) == 1):
```
jams2, line 697: Suggestion: `datetime` has `date` and `time` methods. Comparing `self.start_date.date() == self.end_date.date() and ... ( time not same )` might be clearer than using a `timedelta`.

Question: is there validation that enforces that events start before they end? If not, this might result in a negative delta which would cause this comparison to fail.

etna/whatson/models.py
helenb: If the times are different, we want this comparison to fail - instead we move to the logic of one date with different start and end times.

etna/whatson/models.py
helenb: Unfortunately there is no validation to stop events having an end date before the start date - it feels like there should be though, so is that actually a separate problem to fix?

etna/whatson/models.py
helenb: This is removed now. I have added the format variables here instead - I think it reads clearer to declare them all together at the start rather than as they are needed, but let me know if you'd like me to change that.

etna/whatson/models.py
helenb: variables added and f-strings added in https://github.com/nationalarchives/ds-wagtail/pull/1470/commits/1238a0ca1875a7be135bf6da42d4f00d7d83375b

etna/whatson/models.py
helenb: `.date()` and `.time()` methods used instead of `delta` in https://github.com/nationalarchives/ds-wagtail/pull/1470/commits/1238a0ca1875a7be135bf6da42d4f00d7d83375b

helenb:
@jams2 ready for a re-review

etna/whatson/models.py
```diff
@@ -685,6 +685,33 @@ def primary_access_type(self):
...
+        if (
+            dates_same
+            and (self.start_date.time() != self.end_date.time())
+            and (len(self.sessions.all()) == 1)
+        ):
+            return f"{self.start_date.strftime(format_day_date_and_time)}–{self.end_date.strftime(format_time_only)}"
+        # Multiple sessions on one date
+        # Eg. Monday 1 January 2024
+        if dates_same and len(self.sessions.all()) > 1:
+            return self.start_date.strftime("%A %-d %B %Y")
```
jams2, line 709: Suggestion: declare variable for this format string too, for consistency.

etna/whatson/models.py
```diff
@@ -685,6 +685,33 @@ def primary_access_type(self):
...
+        # Event has multiple dates
+        # Eg. 1 January 2024 to 5 January 2024
+        if self.start_date.date() != self.end_date.date():
```
jams2, line 712:
```suggestion
        if not dates_same:
```

etna/whatson/models.py
```diff
@@ -685,6 +685,33 @@ def primary_access_type(self):
...
+            return f"{self.start_date.strftime(format_date_only)} to {self.end_date.strftime(format_date_only)}"
+
```
jams2, line 714: Question: do we need a fallback case? Is it possible that there are 0 sessions?

etna/whatson/models.py
helenb: whoops, missed that one

etna/whatson/models.py
helenb: You have to add at least one session. Would there be any others?

etna/whatson/models.py
helenb: Added in https://github.com/nationalarchives/ds-wagtail/pull/1470/commits/3b4dfa3deaa36afd1c22dac630c7805595577fea

etna/whatson/models.py
jams2: I think that covers it:

clause 1: date=, time=, 1 session
clause 2: date=, time!=, 1 session
clause 3: date=, 1+ session
    - implied: date=, time=, 1+ session
    - implied: date=, time!=, 1+ session
clause 4: date!=

jams2:
I don't seem to have the option to approve, paging @jamesbiggs

jamesbiggs (approved):
🚀

PR #1503, opened 2024-01-09
Optimise image renditions; change to JPG, set quality

ahosgood:
Change image renditions in the API - change to JPEG and set lower quality to save bandwidth


jamesbiggs (approved):
All looks good to me! 😃

PR #1504, opened 2024-01-10
Introduce APIRichTextBlock

ahosgood:
From https://github.com/wagtail/wagtail/issues/2695#issuecomment-627226782, `RichTextBlock` instances from Wagtail don't compile well for the API.

See http://localhost:8000/api/v2/pages/3/, `body[0].value.paragraph` comes out as `<a id=\"5\" linktype=\"page\">Explore the collection</a>`.

This change transforms these Wagtail-specific elements to HTML (i.e. `<a href=\"/explore-the-collection/\">Explore the collection</a>`)


jamesbiggs (commented):
Nice work AJ, does this not happen with the Django `RichTextField` as well? Or is it just an issue for Wagtail's `RichTextBlock`?

ahosgood:
@jamesbiggs - I'm not sure. I only noticed it in the homepage body but maybe it affects `RichTextField`s as well.

jamesbiggs:
image

@ahosgood are you okay to extend this to the `RichTextField` that we use as well? Tested this on the `intro` field and it has the same issue as the `Block` version

jamesbiggs:
I've just tested again and it doesn't look like Wagtail's `RichTextField` actually has a `get_api_representation` method - so extending it doesn't do anything, annoyingly :(

I think we'll have to do this at the serializer level... I've put a question in the Wagtail Slack to find out

ahosgood:
@jamesbiggs - Let's just do this for the `RichTextBlock` for now then 👍🏼

jamesbiggs (approved):
Happy to approve this as long as you're okay with the changes I made too @ahosgood 😃

PR #1532, opened 2024-02-23
Mount media folder locally

tm-kn:
## About these changes

I was getting permission denied (Django does not have permission to write a file to the media volume) error when trying to upload an image on my local build.

I wonder if we should mount the media folder locally rather than use the volume. That seemed to clear the issue for me. Is there any advantage of creating a separate volume for the media files?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       - ./pyproject.toml:/app/pyproject.toml
       - ./package.json:/app/package.json
       - ./package-lock.json:/app/package-lock.json
-      - media:/media
+      - ./media:/media
     depends_on:
       - db
     ports:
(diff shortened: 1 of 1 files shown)


tm-kn:
This also required a gitgnore addition.

ahosgood:
@tm-kn - Out of interest, are you working on Windows or Mac? If Windows, are you using WSL or not? We've had a few issues with the compatibility between Mac and Windows development environments in the past.

tm-kn:
@ahosgood I use OrbStack on MacOS. What do you use? Can you successfully upload images in Wagtail admin without permission issues?

ahosgood:
@tm-kn - I don't have any issues uploading from http://localhost:8000/admin/images/multiple/add/ and I don't think anyone else has. We aren't using OrbStack because we haven't had any reason to move away from Docker Desktop.

PR #1543, opened 2024-03-04
[DOR-47] Serve IIIF manifest from Rosetta

tm-kn:
Ticket URL: https://national-archives.atlassian.net/browse/DOR-47

## About these changes

- Add an API call to the new IIIF manifest endpoint.
- The prefix varies to the existing API endpoint, so had to add a new environment variable: `_CLIENT_IIIF_MANIFEST_BASE_URL`.
- Add a fetch call at the record detail view to determine if the record has IIIF manifest generated. This is a bit inefficient as we have to make a request even if the record has no IIIF manifest (Raised https://national-archives.atlassian.net/browse/DOR-53 to address it).
- ~Add a new view at `/catalogue/id/<IAID>/iiif-manifest/` to serve the manifests to the IIIF viewer JavaScript front-end.~
- Decided against implementing the proxy view. Sounds like a counterintuitive idea at the moment. The manifest can link out to a lot of different resources that the viewer may want to call itself. Proxying the view, would mean having to proxy all other resources and that would require implementing quite a complicated router. It feels like this complexity shouldn't belong in the Django web application.

## How to check these changes

View record detail page locally, e.g. http://localhost:8000/catalogue/id/C4700551/

## Merging PR guidance

## Deployment guidance


etna/records/views/records.py
```diff
@@ -124,3 +147,34 @@ def record_detail_view(request, id):
...
+            "Unexpected error happened when trying to fetch the IIIF manifest for a record: requested_record_id=%s",
+            id,
+            exc_info=True,
+        )
+        # Return "503 Service Unavailable" if the IIIF manifest cannot be fetched
+        # for reasons other than it does not exist.
+        return HttpResponse(status=503)
+
+    # Proxy the IIIF manifest as-is from the CIIM (Rosetta) API to the client.
+    return JsonResponse(iiif_manifest.content)
```
tm-kn, line 180: **Issue (blocking):** I think there's a big problem with using this proxy. The IIIF manifest contains links to the API that must be resolvable by the end user. I'm not sure how we want to proceed with that.

jamesbiggs: Sorry Tomasz, I can't view the link you shared as it's behind Torchbox's slack. I've not seen the manifests etc so not sure what the real difference is apart from them being separate (for now)

tm-kn: Sorry! It was a thread on our shared channel with Lee. I've got rid of the proxy view for now because it does not seem feasible to implement it now.

etna/ciim/client.py
```diff
@@ -296,6 +300,37 @@ def fetch(
...
+
+        Used to fetch a single item by its identifier.
+
+        Keyword arguments:
+
+        id:
+            Matches a single IAID.
+            Ex: returns match on Information Asset Identifier - iaid.
+        """
+        response = self.make_request(self.get_public_iiif_manifest_url(id=id))
```
tm-kn, line 318: **Question:** Is there a need to implement a fake request of some kind for the CI tests to not hit the real server?

jamesbiggs: I think we have some `records.json`s that we use for this - I haven't been too involved in the "search" applications though, so not sure what state we're in at the minute. If you want to make some tests that you feel are appropriate, feel free! 😃

jamesbiggs (approved):
LGTM! 🚀

etna/records/views/records.py
```diff
@@ -87,11 +89,38 @@ def record_detail_view(request, id):
...
+        #
+        #       We could make a HEAD request instead to check that, but it feels
+        #       short-sighted given that the progressively enhanced view will
+        #       require all this content to be fetched anyway.
+        #
+        #       This information could also be returned in the fetch endpoint
+        #       so we know in advance if we need to call the IIIF manifest
+        #       endpoint at all. The raised ticket:
+        #       https://national-archives.atlassian.net/browse/DOR-53
+        records_client.fetch_iiif_manifest(id=record.iaid)
```
tm-kn, line 113: **Question:** Should this only run for certain record types? E.g. we might not want to make the request for `ARCHON` and `CREATEORS` record types. They also have different template which means that the viewer is not implemented there.

jamesbiggs: Yes, I think we only want this for `digitised` records which cover all digitised records at TNA. I don't think ARCHON or CREATORS will have any images/digitised stuff. Not sure if it'd be easy to prevent it running on ARCHON etc?

tm-kn: Just added an if-statement, same as for the template selection. An observation: It feels like the view as a function-based view looks quite messy given it does so much right now. There's a lot of noise all in one function. It may be beneficial to implement it as a class-based view or break down the function at some point.

etna/ciim/client.py
tm-kn: Mhm... I am hesitant to add tests I can't run with confidence because the existing unit tests are not ready for Rosetta. I don't want to silently introduce broken tests. I'll push the testing to #1554.

tm-kn:
@jamesbiggs Sorry to be asking you to review this multiple times. I made an update to the code:
- Only fetching IIIF manifest for the digitised records.
- Only fetching IIIF manifest for templates that have the IIIF viewer.
- Added tests for the records client in this PR.
- Added additional tests for the record detail view in #1558. Can't do this as part of this as the unit testing for Rosetta is not working so I can't verify my tests are correct.

jamesbiggs (approved):
All looks good, cheers Tomasz! 😄

PR #1565, opened 2024-03-15
Make strtobool() return a boolean value, not an integer (0, 1)

tm-kn:
## About these changes

`strtobool` function was returning an integer.

I realised this because we were passing `1` instead of `True` to the verify certificates option in HTTP requests in the CIIM client. It only broke when making requests to HTTPS, because requests were looking at a custom certificate at a filesystem path `1`, rather than just using the default verify certificate behaviour.

It might be because this is the first time we used HTTPS instead of HTTP using this API client.

`strtobool` is used heavily in the codebase, and if the rest of the code was not changing an integer to the boolean, there may be side effects once this is merged.

Alternatively, we could ensure whenever this function is used, we explicitly cast return values to the boolean at the call time. I don't know the original reason for returning 0s and 1s in the first place.

## How to check these changes

- CIIM client works as expected over HTTPS.

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -70,7 +70,7 @@ if DEBUG and DEBUG_TOOLBAR_ENABLED:
         "debug_toolbar.middleware.DebugToolbarMiddleware",
     ]
 
-    def show_toolbar(request):
+    def show_toolbar(request) -> bool:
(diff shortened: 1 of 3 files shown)


jamesbiggs (approved):
LGTM!

ahosgood (approved):

PR #1580, opened 2024-03-25
Restrict image sizes in record revealed articles

ahosgood:
Images in record revealed articles are displayed in their original size. This is too large.

This change restricts the image to a maximum dimension of 1024px.


JohnHeeryTNA (approved):
approving change

PR #1584, opened 2024-03-26
[DOR-83] HTML-only record viewer

tm-kn:
Ticket URL: https://national-archives.atlassian.net/browse/DOR-83

## About these changes

This is a basic implementation of the HTML-only viewer.

- See more notes on this implementation here: https://national-archives.atlassian.net/wiki/spaces/DE/pages/347144228/Progressive+Enhancement+implementation

- Simplistic IIIF manifest parser that looks for images on an IIIF canvas. If images are stored in different structures, this won't work.
   - We load the image from the `id` key. We don't reach out to the image API so we don't know what other formats are available. That means that by default we may be fetching quite a big image. This can be improved upon.
- Add query parameters to the record detail view:
   - `record_viewer_index` contains the current canvas index (the zero-indexed IIIF manifest's `items` array index).
   - `show_record_viewer` query parameter that either takes `js` (the default) or `html-only`. That way you can easily force an HTML-only viewer to show for testing.
 - If JavaScript viewer (i.e. Universal Viewer) loads with the value in the `record_viewer_index` query parameter, then the Universal Vierwer loads that canvas index on the initial load. However, after the initial load nothing happens to it. An improvement may be to get UV to update the URL parameter to update as the canvas is changed in UV by the user.

## How to check these changes

You can see the HTML-only version by forcing it in the query parameter:

- http://localhost:8000/catalogue/id/C11050044/?show_record_viewer=html-only

- [ ] Checked things thoroughly before handing over to reviewer
- [ ] Ensured that PR includes only commits relevant to the ticket

## Merging PR guidance

## Deployment guidance

### DIFF
--- a/etna/ciim/client.py
+++ b/etna/ciim/client.py
@@ -24,7 +24,8 @@ from django.utils.timezone import get_current_timezone
 import requests
 
 from etna.ciim.constants import Aggregation
-from etna.records.models import IIIFManifest, Record
+from etna.records import iiif
+from etna.records.models import Record
 
 from .exceptions import (
     ClientAPIBadRequestError,
@@ -194,6 +195,10 @@ class ResultList:
         return f"<{self.__class__.__name} {self.hits}>"
 
 
+class ReturnedResourceIsNotIIIFManifest(Exception):
+    pass
+
+
 class ClientAPI:
     """Client used to Fetch and validate data from Client API."""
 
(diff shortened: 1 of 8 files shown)


janehughes:
FE styling based on this branch: https://github.com/nationalarchives/ds-wagtail/pull/1588

PR #1610, opened 2024-04-16
EDEV-51: Better page serialisation

jamesbiggs:
Ticket URL: [EDEV-51]

## About these changes

These changes go a bit outside of the scope of the ticket but are necessary changes to improve the way we serialise data.

I've made a few functions that will be used in various serializers to get fields without throwing errors if they don't exist. For example, if we want to get the `author` field from an item that may be any sort of page - only the `FocusedArticlePage` will have this field. So, these functions will check if a page has `author` in their `api_fields`. It then retrieves the list of `APIField` items, which contains both the field name, and serializer object.

`default_api_fields` does not get displayed in the API representation of the page itself, but will be displayed if the page is referenced anywhere else (e.g. `featured_article`). These will *always* be displayed (not in the actual API representation though), regardless of the fields that are passed in the `required_api_fields` parameter to the `get_api_data` function.

I am using `default_api_fields` rather than just putting them into the `api_fields` as the pages already come with `id`, `title`, etc. So if we have it in `api_fields` (and just use that for these "secondary" representations) it'll be duplicated in the "primary" representations.

To specify which fields you want to return from a page/item, pass them as a list of strings to `required_api_fields`, and if you are overriding the `get_api_representation` or `to_representation` of something, use `get_api_data` if necessary.

## How to check these changes

Check that "linked" pages are being serialized correctly.

## Merging PR guidance

## Deployment guidance

[EDEV-51]: https://national-archives.atlassian.net/browse/EDEV-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


etna/articles/models.py
```diff
@@ -271,9 +266,19 @@ class Meta:
         + NewLabelMixin.api_fields
         + ArticleTagMixin.api_fields
         + [
-            APIField("similar_items", serializer=PageSerializer(many=True)),
-            APIField("latest_items", serializer=PageSerializer(many=True)),
             APIField("body"),
+            APIField(
+                "similar_items",
+                serializer=DefaultPageSerializer(
+                    required_api_fields=["newly_published_at", "type_label"], many=True
```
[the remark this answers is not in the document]
ahosgood: ...and the image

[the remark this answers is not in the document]
jamesbiggs: Sorry, yes, I should've said - a lot of the fields will be wrong/missing, I needed a bit of help trying to find the right ones (image sizes, correct fields, etc)!

etna/core/serializers/images.py
```diff
@@ -0,0 +1,121 @@
...
+                f"{self.rendition_size}|format-jpeg|jpegquality-{self.jpeg_quality}"
+            )
+            webp_image = value.get_rendition(
+                f"{self.rendition_size}|format-webp|webpquality-{self.webp_quality}"
+            )
+
+            return {
+                "id": value.id,
+                "title": value.title,
+                "image_jpeg": {
```
ahosgood, line 48: Please could we add the `alt` attribute to all images?

ahosgood, line 48: Maybe this just needs to be `jpeg` because this block always appears in an `image` property?

ahosgood: Same with the `image_webp` -> `webp`?

ahosgood: ...but outside of the `image_jpeg` or `image_webp` - alongside the `id` and `title`

etna/core/blocks/image.py
```diff
@@ -40,47 +41,8 @@ def __init__(
...
-                        "heading": value.get_translation_heading_display(),
-                        "text": value.translation,
-                    }
-                    if value.translation
-                    else None
-                ),
-            }
-        return None
+        serializer = DetailedImageSerializer()
+        return serializer.to_representation(value)
```
ahosgood, line 45: I think this is making renditions of all picture elements on the page and cropping them in the process

image

Somewhere, these need to be set as a max dimension of 900px (https://nationalarchives.github.io/design-system/styles/images/#picture-elements)

etna/core/blocks/image.py
```diff
@@ -40,47 +41,8 @@ def __init__(
...
-                    {
-                        "heading": value.get_translation_heading_display(),
-                        "text": value.translation,
-                    }
-                    if value.translation
-                    else None
-                ),
-            }
-        return None
+        serializer = DetailedImageSerializer()
```
ahosgood, line 44:
```suggestion
        serializer = DetailedImageSerializer(rendition_size="fill-900x900")
```

Is this what we need?

etna/authors/models.py
```diff
@@ -204,4 +170,9 @@ def author_names(self):
         if self.authors:
             return ", ".join([author.title for author in self.authors])
 
-    api_fields = [APIField("authors", serializer=AuthorSerializer(many=True))]
+    api_fields = [
```
ahosgood, line 173: Could we also get the author's role in here please?

etna/core/models/mixins.py
```diff
@@ -160,31 +159,14 @@ class Meta:
...
     api_fields = [
-        # APIField("hero_image"),
         APIField("hero_image_caption", serializer=RichTextSerializer()),
         APIField(
-            "hero_image_jpg",
-            serializer=ImageRenditionField(
-                "fill-1200x480|format-jpeg|jpegquality-60", source="hero_image"
-            ),
+            "hero_image",
+            serializer=DetailedImageSerializer("fill-1200x480"),
```
ahosgood, line 165:
```suggestion
            serializer=ImageSerializer("fill-1200x480"),

```

I don't think we need all the fields for a hero image.

etna/core/models/mixins.py
```diff
@@ -160,31 +159,14 @@ class Meta:
...
-                "fill-600x400|format-jpeg|jpegquality-60", source="hero_image"
-            ),
-        ),
-        APIField(
-            "hero_image_webp_small",
-            serializer=ImageRenditionField(
-                "fill-600x400|format-webp", source="hero_image"
-            ),
+            "hero_image_small",
+            serializer=DetailedImageSerializer("fill-600x400", source="hero_image"),
```
ahosgood, line 169:
```suggestion
            serializer=ImageSerializer("fill-600x400", source="hero_image"),

```

Again, not all the fields are needed.

etna/core/serializers/images.py
ahosgood: I'm fine with it being in each of the `image_jpeg` and `image_webp` then - just trying to reduce duplication 👍🏼

etna/core/blocks/image.py
ahosgood: The same goes for images in highlight galleries

etna/core/serializers/images.py
jamesbiggs: Our images don't have any alt text, only if they're used in the page, then we add some at that level rather than at image level - so if an image is used it'll be like this:
image
`caption` and `alt_text` are part of the `ImageBlock` StreamField, so we can't add them to the `image` part because they're like separate fields, if that makes sense 😅

etna/core/models/basepage.py
```diff
@@ -146,45 +145,29 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+        APIField("full_url"),
+        APIField("type_label"),
+        APIField("teaser_text"),
+    ]
+
     api_fields = [
         APIField("type_label"),
-        APIField("teaser_image"),
         APIField("teaser_text"),
         APIField(
```
ahosgood, line 160: Are we able to get a `teaser_image_alt_text` as a field as well?

etna/articles/models.py
```diff
@@ -512,6 +540,48 @@ def latest_items(
...
+
+    panels = [
+        FieldPanel("image"),
+        FieldPanel("alt_text"),
+        FieldPanel("caption"),
+    ]
+
+
+class GallerySerializer(serializers.ModelSerializer):
+    image = HighlightImageSerializer(rendition_size="max-1024x1024")
```
ahosgood, line 573: Until we properly design the gallery component, these images need a specific PNG rendition with a `max-900x900` size.

etna/core/serializers/images.py
```diff
@@ -0,0 +1,121 @@
...
+    def to_representation(self, value):
+        if value:
+            jpeg_image = value.get_rendition(
+                f"{self.rendition_size}|format-jpeg|jpegquality-{self.jpeg_quality}"
+            )
+            webp_image = value.get_rendition(
+                f"{self.rendition_size}|format-webp|webpquality-{self.webp_quality}"
+            )
+
+            return {
```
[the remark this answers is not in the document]
jamesbiggs: Our images don't have alt_text on the actual images, just on the `ImageBlock` where we use them in StreamFields - so we won't be able to get alt text at this level, or where we use `teaser_image` either as we don't have a `teaser_image` alt text field

etna/articles/models.py
jamesbiggs: We're using 1024x1024 at the minute in the Highlight Gallery
```
{% image first_gallery_item.image max-1024x1024 format-webp as image_webp %}
{% image first_gallery_item.image max-1024x1024 as image_png %}
{% image first_gallery_item.image max-1024x1024 as base_img %}
```

I'll change how this gets the images to allow for PNG and other image types :)

ahosgood: It's fine - we'll stick with 1024 x 1024 for now until we can rebuild the component 👍🏼

etna/core/serializers/images.py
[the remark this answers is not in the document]
ahosgood: image

We're missing the `alt_text` from topics and time periods.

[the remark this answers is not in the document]
jamesbiggs: Topics and time periods will be in the same boat as just images I think... Image renditions do generate alt text but it's just the image title, so we won't get much use out of that

I also considered the alt text on images a while ago but the content team (and I believe Andy Babic) said that alt text could be done on a case-by-case basis :(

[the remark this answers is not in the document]
ahosgood: I wonder why we've ended up with things like the transcript as part of the image then, rather than putting that down to a case-by-case thing? 🤔🤔🤔

ahosgood:
@jamesbiggs With the content stream block `featured_record_article` (e.g. http://localhost:8000/api/v2/pages/282/), I think we are missing a `date_text` field...

jamesbiggs:
> @jamesbiggs With the content stream block `featured_record_article` (e.g. http://localhost:8000/api/v2/pages/282/), I think we are missing a `date_text` field...

@ahosgood That might be for `record_links` (which won't be done yet until we do the record serializers) - `featured_record_article` just has a page link:
image

This is the record links:
image

etna/core/serializers/images.py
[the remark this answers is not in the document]
ahosgood: We can loop back on this in the future if it's not working for us 😊

etna/core/models/basepage.py
ahosgood: _For now_, we can just have these as decorative images but I think this is a good example of when the `alt` attribute needs to go in the image model because we don't have the ability to set it right now.

The image is attached to the page and we are just linking the page so there's no opportunity to add alternative text. 🥲

Another thing to raise with content designers and see what they want to do...

etna/core/blocks/cta.py
```diff
@@ -2,11 +2,15 @@
...
+
 
 class LargeCardLinksBlock(blocks.StructBlock):
     heading = blocks.CharBlock(max_length=100, required=False)
-    page_1 = blocks.PageChooserBlock(label=_("Link one target"))
-    page_2 = blocks.PageChooserBlock(label=_("Link two target"))
+    page_1 = APIPageChooserBlock(
+        label=_("Link one target"), required_api_fields=["teaser_image"]
+    )
+    page_2 = APIPageChooserBlock(label=_("Link two target"))
```
ahosgood, line 13:
```suggestion
    page_2 = APIPageChooserBlock(label=_("Link two target"), required_api_fields=["teaser_image"])
```

ahosgood (approved):
All looks fantastic - nice one!

Docs/comments can be added in a separate PR just so we can get this up and tested.

PR #1632, opened 2024-05-30
Add redis config for services not on platform.sh

ahosgood:
Set up redis config for services not on platform.sh, ready for AWS

Setup is described here: https://docs.wagtail.org/en/stable/advanced_topics/performance.html


jamesbiggs (approved):
Cheers AJ!

PR #1638, opened 2024-06-05
DSRC-62: Document block

jamesbiggs:
Ticket URL: [DSRC-62]

## About these changes

Adds a `DocumentBlock` to the `ArticlePageStreamBlock` so that you can add a link to a document with your desired text.

#### Page view:
image

#### Document view:
image

### In the API:
image

## Merging PR guidance

## Deployment guidance

[DSRC-62]: https://national-archives.atlassian.net/browse/DSRC-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ahosgood:
There is a DocumentSerializer: https://github.com/wagtail/wagtail/blob/5a1ba5abe2511aad29d927cef85e5a73916f2375/wagtail/documents/api/v2/serializers.py

I can't remember where the design is but I think we need the file name, mime type and file size. The model looks like it has a few of these fields ready to go: https://github.com/wagtail/wagtail/blob/5a1ba5abe2511aad29d927cef85e5a73916f2375/wagtail/documents/models.py#L40

I think there was a requirement to allow a list of downloads - was this your understanding as well?

ahosgood:
Another thought - are these going into article pages or the yet-to-be-defined "standard page"?

jamesbiggs:
> There is a DocumentSerializer: https://github.com/wagtail/wagtail/blob/5a1ba5abe2511aad29d927cef85e5a73916f2375/wagtail/documents/api/v2/serializers.py
> 
> I can't remember where the design is but I think we need the file name, mime type and file size. The model looks like it has a few of these fields ready to go: https://github.com/wagtail/wagtail/blob/5a1ba5abe2511aad29d927cef85e5a73916f2375/wagtail/documents/models.py#L40

Perfect 😎 
 
> I think there was a requirement to allow a list of downloads - was this your understanding as well?

I haven't seen that anywhere on any of the tickets, just the ability to add one file... Was it in a discussion somewhere?
image

ahosgood:
We have too many tickets per component! 😅

The design one mentions one or more:
https://national-archives.atlassian.net/jira/software/projects/DES/boards/138?selectedIssue=DES-5

There is a design somewhere that shows a list as well but I've only seen it on other's screens.

config/settings/base.py
```diff
@@ -338,6 +338,9 @@
     "ol",
     "ul",
 ]
+GENERIC_RICH_TEXT_FEATURES = RESTRICTED_RICH_TEXT_FEATURES + [
+    "document-link",
+]
```
jamesbiggs, line 343: Will add these settings to some RichText fields/blocks when we have somewhere to use them 🙂

ahosgood: Yeah, I think this should probably be more a case-by-case thing as we get more different "types" or rich text areas...

etna/core/blocks/document.py
```diff
@@ -0,0 +1,53 @@
...
+
+class DocumentBlock(blocks.StructBlock):
+    """
+    A block for embedding a document file in a page.
+    """
+
+    link_text = blocks.CharBlock(
+        required=True,
+        help_text="The text to display for the link to the document",
+    )
```
jamesbiggs, line 13: Not sure we actually need this anymore... Could this just be the file title? We have "file link text" in the requirements, but in the designs we are basically just using the title
image
image

etna/core/blocks/document.py
```diff
@@ -0,0 +1,53 @@
...
+    def get_api_representation(self, value, context=None):
+        representation = super().get_api_representation(value, context)
+
+        file = value.get("file")
+
+        if file:
+            if file.file_size >= 1000000:
+                file_size = f"{((file.file_size / 1024) / 1024):.2f}MB"
+            else:
+                file_size = f"{(file.file_size / 1024):.2f}KB"
```
jamesbiggs, line 25: Used 1024 (binary) representation for MB/KB instead of just 1000 - Windows file system (and I think all file systems) use 1024.

etna/core/blocks/document.py
ahosgood: Yes - we should just use the document title 👍🏼

ahosgood: (as long as you can change the name of the file after uploading to Wagtail)

etna/core/blocks/document.py
ahosgood: I _think_ the current standard is 1,000 bytes = 1 kilobyte and 1,000 kilobytes = 1 megabyte which I personally hate because it has always been 1,024 for me. I think that 1,024 is for networking (when you talk about transfer speeds) and 1,000 is for storage.

There are a "new" set of silly sounding prefixes like "kibibyte", "mebibyte" and "gibibyte" which we are meant to use but honestly, I don't think a member of the public will know or care about these. (https://en.wikipedia.org/wiki/Binary_prefix#Definitions)

https://xkcd.com/394/ 😅

TL;DR - Can we use 1,000 base for now please?

Could we also please have two fields returned in the API; `file_size` which is simply the number of bytes and `pretty_file_size` which is the file size with suffix? I might need the "raw" number in the frontend.

We can use something like the below to produce a pretty size. (Please amend as necessary):

```python
def pretty_file_size(file_size):
    suffixes = ["B", "kB", "MB", "GB"]
    suffix_index = 0
    while file_size > 1000 and suffix_index < len(suffixes) - 1:
        file_size = file_size / 1000
        suffix_index = suffix_index + 1
    return f"{round(file_size, min(suffix_index, 2))}{suffixes[suffix_index]}"
```

jamesbiggs: I was always taught kibibyte and kilobyte, but I think most people call a kibibyte (1024) kilobyte 😂 When I use 1000 base it can be off by quite a bit on my machine...

image
image

In this example 621.69MB is the same as the size of the file when it's on my machine, but if I do base 1000 it would be 651.87MB... I'm happy to go with either but I do think we should be keeping in line with the file system which is using kibi/mebi/gibibyte (1024), as it puts it out by quite a bit...

jamesbiggs: image

image

ahosgood: I had to do a bit more reading up on where everything is with this flame war 😅

It looks like basically everyone except Microsoft use decimal (1,000) - this includes Mac OS, iOS, Ubuntu and Android. Also I just found out that weirdly the online Microsoft services like Office 365 use decimal as well.

https://www.simonrjones.net/2022/01/a-little-bit-about-filesize-units/#:~:text=Today%20Ubuntu%20Linux%2C%20Mac%20OS%2C%20iOS%20and%20Android%20use%20decimal%20for%20file%20storage%20sizes.%20Windows%2C%20as%20far%20as%20I%E2%80%99m%20aware%2C%20still%20uses%20binary%20units.%20However%2C%20to%20spice%20things%20up%20Microsoft%E2%80%99s%20cloud%20office%20service%20365%20uses%20decimal%20units%20when%20referring%20to%20cloud%20storage%20size!

I honestly think we should stick with decimal (1,000) and maybe once Microsoft stop turning Windows into an advertising platform and actually fix some stuff, they might move over to the correct units... 🤞🏼

jamesbiggs: Good old Windows... 😅

I'll swap it over to 1000!

etna/core/models/documents.py
```diff
@@ -0,0 +1,23 @@
...
+from wagtail.documents.models import AbstractDocument, Document
+
+
+class CustomDocument(AbstractDocument):
+    extent = models.CharField(blank=True, null=True)
+    description = models.TextField(blank=True, null=True)
+
+    @property
+    def pretty_file_size(self):
+        suffixes = ["B", "KB", "MB", "GB"]
```
ahosgood, line 12:
```suggestion
        suffixes = ["B", "kB", "MB", "GB"]
```

etna/core/models/documents.py
```diff
@@ -0,0 +1,23 @@
...
+        i = 0
+        pretty_file_size = self.file_size
+        while pretty_file_size >= 1000 and i < len(suffixes) - 1:
+            pretty_file_size /= 1000
+            i += 1
+        return f"{pretty_file_size:.1f}{suffixes[i]}"
```
ahosgood, line 18: What would be great here is a different number of decimal places for different sizes.

For B and kB we don't really need any decimal places but as the file gets larger, those decimal places get more important.

Example:

- `50B` (0 DP for B)
- `50kB` (0 DP for kB)
- `50.1MB` (1 DP for MB)
- `5.12GB` (2 DP for GB)

Another thing that would be nice is not to end with a `.0` - e.g. `50.0MB` should ideally just be `50MB`.

ahosgood: Something like this would take care of the decimal points:

```python
return f"{round(file_size, min(suffix_index - 1, 2))}{suffixes[suffix_index]}"
```

Not sure how is best to remove the `.0` or `.00` if there is one.

ahosgood (approved):

PR #1640, opened 2024-06-07
Image size tweaks and add headless mixin

ahosgood:
- Update intro image size in `RecordArticlePage`
- Change all default `webp_quality` from `80` to `60`
- Add `HeadlessMixin` to `BasePage` and remove `HeadlessPreviewMixin`
- Removed unused `teaser_image_large`

### DIFF
--- a/etna/core/models/basepage.py
+++ b/etna/core/models/basepage.py
@@ -166,10 +166,6 @@ class BasePage(MetadataPageMixin, DataLayerMixin, Page, HeadlessPreviewMixin):
             serializer=ImageSerializer("fill-600x400"),
         ),
         APIField(
-            "teaser_image_large",
-            serializer=ImageSerializer("fill-1200x480", source="teaser_image"),
-        ),
-        APIField(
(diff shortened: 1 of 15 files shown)


etna/core/models/basepage.py
```diff
@@ -40,7 +40,7 @@
 
 @method_decorator(apply_default_vary_headers, name="serve")
 @method_decorator(apply_default_cache_control, name="serve")
-class BasePage(MetadataPageMixin, DataLayerMixin, Page, HeadlessPreviewMixin):
+class BasePage(MetadataPageMixin, DataLayerMixin, HeadlessMixin, Page):
```
ahosgood, line 43: This breaks some tests but is the recommended way according to https://github.com/torchbox/wagtail-headless-preview?tab=readme-ov-file#usage

ahosgood: `HeadlessPreviewMixin` works for now...

jamesbiggs (approved):
🚀

PR #1641, opened 2024-06-07
Add API endpoint for password protected pages

ahosgood:
This PR brings a new API endpoint for accessing password protected pages.

There is a new property in the page `meta` of API detail pages called `privacy` which can either be `password` or `public`.

If set to `password` then the new frontend can prompt for a password before deferring the request to the new `private_page` endpoint.

To test, change a page's visibility to "Private, accessible with a shared password" with the password `mypassword`:

image

## API responses for page 352:

- http://localhost:8000/api/v2/private_page/352/ - should return minimal data (just an ID), some meta data ("privacy" and `"locked": true`) a message about not having provided a password
- http://localhost:8000/api/v2/private_page/352/?password=nope - should return minimal data (just an ID), some meta data ("privacy" and `"locked": true`) a message to say the password is not correct
- http://localhost:8000/api/v2/private_page/352/?password=mypassword - should return the full and correct page

## Notes

- Public pages should still be accessible
- Non-existent IDs should return a `404`
- Other privacy modes ("Private, accessible to any logged-in users" and "Private, accessible to users in specific groups") should result in a `403` response
- Protected pages should not appear in listing results (e.g. http://localhost:8000/api/v2/pages/?search=Ireland)
- Wagtail page redirects should be applied (http://localhost:8000/admin/redirects/)

### DIFF
--- a/etna/api/tests/test_pages.py
+++ b/etna/api/tests/test_pages.py
@@ -339,7 +339,6 @@ class APIResponseTest(WagtailPageTestCase):
         return data
 
     def compare_json(self, path: str, json_file: str):
-        self.maxDiff = None
         if api_data := self.get_api_data(path):
             if api_data.startswith("Endpoint") or api_data.startswith("Unable"):
                 self.fail(api_data)
--- a/etna/api/tests/test_pages.py
+++ b/etna/api/tests/test_pages.py
@@ -339,6 +339,7 @@ class APIResponseTest(WagtailPageTestCase):
         return data
 
     def compare_json(self, path: str, json_file: str):
+        self.maxDiff = None
         if api_data := self.get_api_data(path):
             if api_data.startswith("Endpoint") or api_data.startswith("Unable"):
                 self.fail(api_data)
--- a/etna/api/urls.py
+++ b/etna/api/urls.py
@@ -1,10 +1,14 @@
+import logging
+
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.http import Http404
 from django.utils.crypto import constant_time_compare
 
 from wagtail.api.v2.router import WagtailAPIRouter
 from wagtail.api.v2.utils import BadRequestError
 from wagtail.api.v2.views import PagesAPIViewSet
+from wagtail.contrib.redirects.models import Redirect
 from wagtail.images.api.v2.views import ImagesAPIViewSet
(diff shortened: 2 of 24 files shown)


etna/api/urls.py
```diff
@@ -1,16 +1,134 @@
...
+                for translation in site.root_page.get_translations():
+                    queryset |= base_queryset.descendant_of(translation, inclusive=True)
+
+        else:
+            # No sites configured
+            queryset = queryset.none()
+
+        return queryset
+
+    # TODO: Why do I have to redefine these after meta_fields in etna/core/models/basepage.py?
```
ahosgood, line 124: ???

etna/api/tests/test_pages.py
```diff
@@ -339,6 +339,7 @@ def replace_placeholders(self, data: str):
         return data
 
     def compare_json(self, path: str, json_file: str):
+        self.maxDiff = None
```
ahosgood, line 342: TODO: Remove

etna/api/urls.py
jamesbiggs: I thought it might've been because you're querying on `Page` rather than `BasePage` in line 89, but that doesn't work because `BasePage` is abstract. I tried it with `ArticlePage` and same thing there - no `meta_fields` ☹️ 

Did `meta_fields` ever work on `BasePage` with the default serializer? I tried adding it to what we have on develop and it made no difference...

etna/core/models/basepage.py
```diff
@@ -145,6 +145,19 @@ def get_datalayer_data(self, request: HttpRequest) -> Dict[str, Any]:
...
+    @property
+    def privacy(self):
+        privacy = [r.restriction_type for r in self.get_view_restrictions()]
+        if privacy:
+            return privacy[0]
+        return "public"
+
+    meta_fields = [
+        "privacy",
+        "latest_revision_created_at",
```
jamesbiggs, line 157: Nice work AJ - just one question, why do you want `latest_revision_created_at`?

ahosgood: Sorry, I was trying to get `last_published_at`. Updated now.

etna/api/urls.py
ahosgood: You're right - they don't look like they work on abstract classes.

Kept the fields in the API but removed from `BasePage`.

jamesbiggs (approved):

PR #1648, opened 2024-06-12
DSRC-65: Call to action block

jamesbiggs:
Ticket URL: [DSRC-65] + [DSRC-68]

## About these changes

Adds a CTA block (DSRC-65) and a button block (DSRC-68)

In the API:
image

In the CMS:
image

## Merging PR guidance

## Deployment guidance

[DSRC-65]: https://national-archives.atlassian.net/browse/DSRC-65?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSRC-68]: https://national-archives.atlassian.net/browse/DSRC-68?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


etna/core/blocks/cta.py
```diff
@@ -29,3 +32,38 @@ def get_context(self, value, parent_context=None):
             link_pages.append(page_2.specific)
         context["link_pages"] = link_pages
         return context
+
+
+class ButtonBlock(blocks.StructBlock):
+    label = blocks.CharBlock()
+    link = APIPageChooserBlock(required=False)
+    external_link = blocks.URLBlock(required=False)
```
ahosgood, line 40: Would it be possible to get just one field back in the API?

I think these might be better as `page_link`/`internal_link` and `external_link` but returned in the API as just `href` (or something) so I don't have to make conditionals.

What do you think?

jamesbiggs: Good idea, I was trying to think of a way to simplify it but couldn't think of anything good. Will add that now!

jamesbiggs: image

How's this?

ahosgood: Perfect! 🤩

ahosgood:
Was there a requirement for the ability to have multiple CTA buttons next to one another?

We also need a boolean field for the button which is `accented`, defaulting to `False`. Setting to `True` would allow the editor to add an accented (coloured) button.

We also need an optional `title` field so editors can include a longer, more descriptive title of where the button goes.

ahosgood:
Another thought - are these going into article pages or the yet-to-be-defined "standard page"?

jamesbiggs:
> Another thought - are these going into article pages or the yet-to-be-defined "standard page"?

I just had a chat with Lizzy and she said they will be in a "standard" page, so I'll make a PR before this one is merged to add a standard page, add these components to that, and remove them from Article pages.


> Was there a requirement for the ability to have multiple CTA buttons next to one another?

Content haven't asked for it... But I think there could be a case for that (at some point). We can always allow this to add multiple `buttons` at a later point if that's the case 🙂 

> We also need a boolean field for the button which is `accented`, defaulting to `False`. Setting to `True` would allow the editor to add an accented (coloured) button.

👍 

> We also need an optional `title` field so editors can include a longer, more descriptive title of where the button goes.

They have the `sub-heading` feature at the minute, I think they would use that here instead of having a dedicated CTA "title"?

jamesbiggs:
image

ahosgood:
> > Was there a requirement for the ability to have multiple CTA buttons next to one another?
> 
> Content haven't asked for it... But I think there could be a case for that (at some point). We can always allow this to add multiple `buttons` at a later point if that's the case 🙂

Could we make it a list with a max of 1? I'm just thinking about the future - it would be more pain to change a dict to a list because the new format would "break" the frontend whereas adding more buttons, it might just not display them.

> > We also need an optional `title` field so editors can include a longer, more descriptive title of where the button goes.
> 
> They have the `sub-heading` feature at the minute, I think they would use that here instead of having a dedicated CTA "title"?

I was on about a specific `title` attribute on the button itself but looking at it more, I don't think this is needed. We'll just need give advice to devolved editors on how to properly write button text. Carry on!

etna/core/blocks/cta.py
```diff
@@ -29,3 +32,52 @@ def get_context(self, value, parent_context=None):
...
+                "You must provide either a page link or an external link."
+            )
+
+        return data
+
+    def get_api_representation(self, value, context=None):
+        representation = {
+            "label": value["label"],
+            "href": value.get("external_link") or value["link"].full_url,
+            "accented": value.get("accented") or False,
```
ahosgood, line 65:
```suggestion
            "accent": value.get("accented") or False,
```

ahosgood (approved):

PR #1665, opened 2024-06-27
DSRC-75: Notification banner

jamesbiggs:
Ticket URL: [DSRC-75]

## About these changes

This PR spruces up the `Alert` model we've currently got on the site, but enables it to be utilised properly in the API. Also added the `AlertMixin` to every page (via `BasePage`) so that you can use the "notification banner" on different pages.

`alert_check` will get any "cascading" alerts from parent pages. If a parent page, anywhere up the hierarchy, has a cascading alert enabled, then this will be picked up on all child pages.

Home page with an alert on it (cascading false):
image

A great-great-great-grandchild page:
image

With cascading true on the alert, on the great-great-great-grandchild page:
image

Cascading on parent page, differing active alert on great-great-great-grandchild page:
image

## How to check these changes

Alerts are registered in http://localhost:8000/admin/snippets/alerts/alert/, if you want to change or create any, go to that URL, and if you want to add them to a page, you can go into the page and then the `Settings` tab to add or remove these from the pages.

## Merging PR guidance

## Deployment guidance

[DSRC-75]: https://national-archives.atlassian.net/browse/DSRC-75?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


etna/alerts/models.py
```diff
@@ -72,11 +75,22 @@ class AlertMixin(models.Model):
...
+    def alert_check(self):
+        if self.alert and self.alert.active:
+            return self.alert
+        elif parent := self.get_parent():
+            if type(parent.specific) is not Page:
+                if parent_alert := parent.specific.alert_check:
+                    if parent_alert.cascade:
+                        return parent_alert
+        return None
+
```
jamesbiggs, line 88: Do we want the "closest" alert to show (like it is currently)
e.g.
Page 1 has a child, Page 2. Page 1 has an alert on, Page 2 also has an alert on.
Page 2 has a child, Page 3. Should Page 2 and 3 take Page 1's alert? Or should Page 2 and 3 use Page 2's alert?

Or should we do it based on "alert level"?
e.g.
Page 1 has a "high" alert, Page 2 has a "medium" alert - should Page 2 use its own "medium" alert? Or should it take the highest level of alert (Page 1)

jamesbiggs: Answer: We want the parent-most alert to show regardless of level.

jamesbiggs:
@ahosgood I realise `alert_check` probably isn't the best name for this. Any other suggestions..? Just `notification`?

ahosgood:
@jamesbiggs Is this meant to be an array?

I think something like `alerts` might be better than `alert_check`.

jamesbiggs:
> @jamesbiggs Is this meant to be an array?
No, just the one item

> I think something like `alerts` might be better than `alert_check`.
Okay!

ahosgood:
> No, just the one item

If there is only ever going to be just one, could we call it `global_alert` instead please? Sorry for flip-flopping on this!

ahosgood (approved):

PR #1666, opened 2024-06-27
API sibling pages

ahosgood:
Add a query parameter `sibling_of` (e.g. http://localhost:8000/api/v2/pages/?sibling_of=26) to allow querying of pages at the same level.

This will help when we need to add other pages in a section to the sidebar (e.g. the right column of https://www.nationalarchives.gov.uk/about/our-role/what-we-do/ - "Also in Our role")

### DIFF
--- a/etna/api/urls.py
+++ b/etna/api/urls.py
@@ -14,6 +14,7 @@ from wagtail.images.api.v2.views import ImagesAPIViewSet
 from wagtail.models import Page, PageViewRestriction, Site
 
 from rest_framework import status
+from rest_framework.filters import BaseFilterBackend
 from rest_framework.response import Response
 from wagtail_headless_preview.models import PagePreview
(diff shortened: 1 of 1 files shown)


etna/api/urls.py
```diff
@@ -23,8 +24,40 @@
...
+    def filter_queryset(self, request, queryset, view):
+        if "sibling_of" in request.GET:
+            try:
+                parent_page_id = int(request.GET["sibling_of"])
+                if parent_page_id < 0:
+                    raise ValueError()
+                parent_page = view.get_base_queryset().get(id=parent_page_id)
+                grandparent_page = view.get_base_queryset().get(
+                    id=parent_page.get_parent().id
+                )
```
ahosgood, line 42: Is there a more efficient way to do this?

PR #1683, opened 2024-07-19
Retain fields query parameter when redirecting from find_view in API

ahosgood:
Currently,

- http://localhost:8000/api/v2/pages/find/?html_path=/&fields=_,global_alert redirects to
- http://localhost:8000/api/v2/pages/5/

Until https://github.com/wagtail/wagtail/pull/12141 is accepted, this ensures:

- http://localhost:8000/api/v2/pages/find/?html_path=/&fields=_,global_alert redirects to
- http://localhost:8000/api/v2/pages/5/?fields=_,global_alert


jamesbiggs:
I can't seem to get this to work, it just keeps redirecting me to http://host.docker.internal:8000/api/v2/pages/3/fields=_,alert_check and saying the site isn't reachable... Also, what is this needed for?

jamesbiggs:
My detail_url is showing up as `"http://host.docker.internal:8000/api/v2/pages/5/"`

ahosgood:
The redirect will work inside one of the Docker containers. `http://host.docker.internal` is the address used when you are inside a container and want to get to the host machine. This works fine when the frontend requests the URL from the CMS - it can follow it but for us humans outside, we just need to replace the "host.docker.internal" with "localhost".

The reason I made this PR (and sorry for not explaining!) is for static pages on the site or pages that aren't served by Wagtail. I would like for them to also be able to display the global alerts you added (https://github.com/nationalarchives/ds-wagtail/pull/1665).

Wagtail pages get this as part of the API response but I wanted to be able to query the top-level page and just get any global alerts.

Since I don't know the ID of the home page, I can find it by its path (http://localhost:8000/api/v2/pages/find/?html_path=/ which redirects to page `3`) but I only want the `global_alert` field.

I can return just one field with http://localhost:8000/api/v2/pages/3/?fields=_,global_alert which would be fine if I knew the page ID but because I don't, I need to redirect from the `find` endpoint.

All non-Wagtail pages can then call http://localhost:8000/api/v2/pages/find/?html_path=/&fields=_,global_alert to get the global alert (if any) on the home page.

jamesbiggs:
@ahosgood Okay, cool!

BTW, `global_alert` will be on all pages, I've got it on develop on the home page, and it can be seen on "Explore the collection"
https://develop-sr3snxi-rasrzs7pi6sd4.uk-1.platformsh.site/api/v2/pages/5/
image

and a child page https://develop-sr3snxi-rasrzs7pi6sd4.uk-1.platformsh.site/api/v2/pages/53/ (and every page that is a child of the home page)
image

You could also set this notification on "Stories from the collection", and if there wasn't one on the home page, it'd take the Stories notification and apply it to all child pages of that page 👍 

So `global_alert` is given to you on all pages! 😃

ahosgood:
That all makes sense.

I think there is another scenario though. Some pages aren't in Wagtail. For example:

- https://tna.dblclk.dev/browse/
- https://tna.dblclk.dev/legal/cookies/
- https://tna.dblclk.dev/search/

These are not in Wagtail (yet) because of complexities such as the cookies page needing to be unique and have a bunch of extra functionality with not much editable content. Nevertheless, these pages still need to be aware of the global alert so they can display it if necessary.

I've just thought though - there is an option for "Show on current and all child pages" for the "main" alert. If this is checked then all these other pages need to show the alert and if not then they don't.

Firstly, I think it should be labeled "Show on all pages" with some help text to explain if not checked, the alert will only appear on the home page. Users won't know what the "current" page it's referring to is.

Then I think we need to return that value in the API as something like `show_on_all_pages` so that when I want to display it on the non-Wagtail child pages above, I know whether it should be visible or not.

jamesbiggs:
> That all makes sense.
> 
> I think there is another scenario though. Some pages aren't in Wagtail. For example:
> 
> * https://tna.dblclk.dev/browse/
> * https://tna.dblclk.dev/legal/cookies/
> * https://tna.dblclk.dev/search/
> 
> These are not in Wagtail (yet) because of complexities such as the cookies page needing to be unique and have a bunch of extra functionality with not much editable content. Nevertheless, these pages still need to be aware of the global alert so they can display it if necessary.
> 
> I've just thought though - there is an option for "Show on current and all child pages" for the "main" alert. If this is checked then all these other pages need to show the alert and if not then they don't.
> 
> Firstly, I think it should be labeled "Show on all pages" with some help text to explain if not checked, the alert will only appear on the home page. Users won't know what the "current" page it's referring to is.
> 
> Then I think we need to return that value in the API as something like `show_on_all_pages` so that when I want to display it on the non-Wagtail child pages above, I know whether it should be visible or not.

Good point! Could you query the root page, and if there's an alert on that one (with cascade enabled) then display that? I can surface the `cascade` value in the API (the one that says apply to current and all child pages) for the non-Wagtail pages?

ahosgood:
@jamesbiggs - Yes, this exactly 👍🏼

jamesbiggs (approved):

PR #1688, opened 2024-07-25
Add setting for WAGTAILAPI_LIMIT_MAX

ahosgood:
Allow the frontend to query more child pages in a single request for things like the sitemap.

At the moment, a maximum of 20 pages can be returned at once.

E.g. http://localhost:8000/api/v2/pages/?limit=100


jamesbiggs (approved):
🚀

PR #1726, opened 2024-10-11
Add chapters to media items

ahosgood:
Allows media items to have chapters.

Chapters will be helpful when we come to migrate podcasts over from https://media.nationalarchives.gov.uk/.


etna/media/blocks.py
```diff
@@ -34,6 +34,9 @@ def get_api_representation(self, value, context=None):
             "date": value.date,
             "description": expand_db_html(value.description),
             "transcript": expand_db_html(value.transcript),
+            "chapters": value.chapters.stream_block.get_api_representation(
+                value.chapters, context
+            ),
```
ahosgood, line 39: This feels wrong but I couldn't figure out how else to do it.

I'm not sure if this is good practice at all...

etna/media/models.py
```diff
@@ -1,16 +1,38 @@
...
+        default=0,
+        validators=[MinValueValidator(0)],
+        label="Time in seconds",
+    )
+    heading = blocks.CharBlock(max_length=20)
+    transcript = APIRichTextBlock(required=False, features=["bold", "italic"])
+
+    def get_api_representation(self, value, context=None):
+        representation = super().get_api_representation(value, context)
+        representation["time"] = int(representation["time"])
```
ahosgood, line 29: I couldn't figure out how to make `DecimalBlock` only store integers - it always wanted to store one decimal place.

jamesbiggs: `IntegerBlock` will solve the decimal thing but for some reason it's saved as a string so you'll still need to wrap it in `int()`, up to you if you want Decimal or Integer block!

etna/media/blocks.py
```diff
@@ -24,6 +24,9 @@ def get_api_representation(self, value, context=None):
         We use expand_db_html to get any rich text fields as useful HTML,
         rather than the raw database representation.
         """
+        chapters = value.chapters.stream_block.get_api_representation(
+            value.chapters, context
+        )
```
jamesbiggs, line 29: Instead of using the actual StreamBlock instance and calling its own function on itself, you can get the data this way:
```
        return {
            "id": value.id,
            "file": value.url,
            "full_url": value.full_url,
            "type": value.type,
            "mime": value.mime(),
            "title": value.title,
            "date": value.date,
            "description": expand_db_html(value.description),
            "transcript": expand_db_html(value.transcript),
            "chapters": sorted(
                [
                    {
                        "time": int(chapter.value["time"]),
                        "heading": chapter.value["heading"],
                        "transcript": expand_db_html(
                            chapter.value["transcript"].source
                        ),
                    }
                    for chapter in value.chapters
                ],
                key=lambda x: x["time"],
            ),
            "width": value.width,
            "height": value.height,
            "duration": value.duration,
        }
```

ahosgood: Which do you think is better?

ahosgood: I think yours is better. Changed.

jamesbiggs (approved):
🚀

PR #1736, opened 2024-10-24
Add VTT subtitles and chapters fields to media items

ahosgood:
Allow subtitles and chapters VTT files for media items


jamesbiggs (approved):
Made a small change to the `verbose_name` so there's no underscore in the CMS. All good otherwise!
🚀

PR #1746, opened 2024-11-13
Return None if record does not exist or multiple objects returned

jamesbiggs:
Temporary fix for errors being raised if a record does not exist in CIIM. This will need some work to gracefully handle errors in the future, in case a record is taken down from CIIM, or CIIM goes down in the future.


ahosgood (approved):
If the tests pass, let's go 👍🏼

PR #1772, opened 2025-01-08
`include_aliases` filter in the API

jamesbiggs:
## About these changes

Removes alias pages from queryset, if the original is in the queryset. If two aliases of the same page are in a queryset, then the "shallowest" (i.e. highest in the tree) is kept, while the other copies are removed.

This is applied to all endpoints that inherit `CustomPagesAPIViewSet` (currently all of them), primarily for use on the blog.

`?include_aliases` can be used to include aliases

## Merging PR guidance

## Deployment guidance


jamesbiggs:
NOTE: Pretty much there, just need to exclude aliases if another alias of the same page exists within the query set
e.g.
page ID 2 and 3 are in a queryset, and they are both aliases of ID 1 - show only one

etna/api/urls.py
```diff
@@ -28,7 +28,7 @@
 
 class CustomPagesAPIViewSet(PagesAPIViewSet):
     known_query_parameters = PagesAPIViewSet.known_query_parameters.union(
-        ["password", "author"]
+        ["password", "author", "include_aliases"]
```
ahosgood, line 31: Are we guaranteed that `include_aliases` will be the last filter using `.union()`?

We need to ensure it is applied after all other filtering.

etna/api/urls.py
```diff
@@ -50,6 +50,38 @@ def listing_view(self, request):
 
         self.check_query_parameters(queryset)
         queryset = self.filter_queryset(queryset)
+
+        if "include_aliases" not in request.GET:
+            alias_pages = queryset.filter(alias_of_id__isnull=False).values(
+                "id", "alias_of_id"
+            )
+            original_ids = set(
+                queryset.filter(alias_of_id__isnull=True).values_list("id", flat=True)
```
ahosgood, line 59: Does this `queryset` have everything in it?

I'm wondering if we have filtered by an `offset` and `limit` (basically paginating the result) and the original isn't in this subset, we aren't getting the full list of IDs here.

Do we need to query a new queryset of ALL pages to ensure this is never the case?

jamesbiggs: Pagination is applied after with `queryset = self.paginate_queryset(queryset)`, this filtering is applied after the main filtering is done (e.g. page type filtering, or `descendant_of` filtering) so that we're running queries on a smaller dataset - so this should be okay 🙂

etna/api/urls.py
ahosgood: 🤦🏼‍♂️

You are right - sorry!

ahosgood (approved):

PR #1807, opened 2025-03-24
Release to main

jamesbiggs:


jamesbiggs:
^ This is to do with the `author.json` test fixture because it has someone's name in, "John Smith" - sorry John Smith!

PR #1810, opened 2025-03-24
Split all API endpoints into separate files

ahosgood:
Split out API endpoints - it was getting too busy in one big file


jamesbiggs (approved):
Good idea 👌

PR #1811, opened 2025-03-24
Add breadcrumbs metadata to CustomPagesAPIViewSet response

ahosgood:
This means the frontend won't have to make a follow-up call of something like http://localhost:8000/api/v2/pages/?ancestor_of=192&order=-depth after the initial page request.

image

### DIFF
--- a/etna/api/urls.py
+++ b/etna/api/urls.py
@@ -100,6 +100,18 @@ class CustomPagesAPIViewSet(PagesAPIViewSet):
         restrictions = instance.get_view_restrictions()
         serializer = self.get_serializer(instance)
         data = serializer.data
+        data["meta"].update(
+            {
+                "breadcrumbs": [
+                    {
+                        "text": ("Home" if page.url == "/" else page.title),
+                        "href": page.url,
+                    }
+                    for page in instance.get_ancestors().order_by("depth")
+                    if page.url
+                ],
+            }
+        )
         if not restrictions:
             return Response(data)
         restricted_data = {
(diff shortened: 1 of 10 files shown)


etna/api/urls.py
```diff
@@ -100,6 +100,18 @@ def detail_view(self, request, pk):
         restrictions = instance.get_view_restrictions()
         serializer = self.get_serializer(instance)
         data = serializer.data
+        data["meta"].update(
+            {
+                "breadcrumbs": [
+                    {
+                        "text": ("Home" if page.url == "/" else page.title),
```
ahosgood, line 107: @jamesbiggs - I want to get back the `short_title` here if the page has one but because we're using the normal `Page` serializer, it doesn't have knowledge of that field.

Any chance you could do some magic here? ✨

etna/api/urls.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import time
```
ahosgood, line 2: @jamesbiggs Is this a mistake?

jamesbiggs: Oops... Sorry!

etna/api/urls.py
jamesbiggs: I've added `.specific(defer=True)` rather than using just `.specific()` (something new I've learnt!).

`.specific()` gets _every_ field from the specific class/model. This includes StreamFields and other big bits of data that might not be needed.

We could use `.defer_streamfields()` but this only excludes StreamFields, still giving us all the extra data that we might not need.

`.specific(defer=True)` gets the specific class/model, but only fetches the additional fields when they're called. In our case, we're only using `short_title` here - so I don't think the additional query makes much of a difference. I've done some testing:
```
defer=true
page=55
0.004607 seconds
0.003863 seconds
0.004145 seconds
0.004095 seconds

page=329 (relies on CIIM)
0.007733 seconds
0.007735 seconds
0.007903 seconds
0.010870 seconds
0.005672 seconds

defer=false
page=55
0.013425 seconds
0.007046 seconds
0.008331 seconds
0.008266 seconds
0.008762 seconds

page=329 (relies on CIIM)
0.011526 seconds
0.008934 seconds
0.009031 seconds
0.008881 seconds
0.009235 seconds
```

Which suggests it works better with `defer=True`.

I've asked in the Wagtail slack what's the best way to get a specific value without calling `.specific()` because it's a pretty hefty query to make for just one value. But for now this seems to be the best option.

It would be good to go through what we've got at some point after I've had a look into caching to see if there are any chunks (like this) that could be candidates for caching.

jamesbiggs (approved):

PR #1814, opened 2025-04-02
Remove `wagtail-generic-chooser` and move to built-in Wagtail Chooser

jamesbiggs:
## About these changes

This PR moves us from using the third-party `wagtail-generic-chooser` to the built-in Wagtail chooser.

We are now going to store a string of the IAID for the Record rather than an instance of a Record. (Potentially move to JSONField to use JSON data in the future, for more resilience?)

## Merging PR guidance

## Deployment guidance


config/settings/base.py
```diff
@@ -319,7 +310,7 @@
 
 # API Client
 
-CLIENT_BASE_URL = os.getenv("KONG_CLIENT_BASE_URL")
+CLIENT_BASE_URL = os.getenv("ROSETTA_DEV_CLIENT_BASE_URL")
```
ahosgood, line 313: I think `CLIENT_BASE_URL` and `ROSETTA_DEV_CLIENT_BASE_URL` both need to be `ROSETTA_API_URL`

config/settings/base.py
```diff
@@ -319,7 +310,7 @@
...
 CLIENT_KEY = os.getenv("KONG_CLIENT_KEY")
 CLIENT_VERIFY_CERTIFICATES = strtobool(
```
ahosgood, line 315: We can remove this env var

config/settings/base.py
```diff
@@ -319,7 +310,7 @@
 
 # API Client
 
-CLIENT_BASE_URL = os.getenv("KONG_CLIENT_BASE_URL")
+CLIENT_BASE_URL = os.getenv("ROSETTA_DEV_CLIENT_BASE_URL")
 CLIENT_KEY = os.getenv("KONG_CLIENT_KEY")
```
ahosgood, line 314: We can remove this env var for now

etna/ciim/client.py
```diff
@@ -1,636 +1,70 @@
...
-        if field in field_list_to_prepare:
-            # replace special chars
-            prepared_value = re.sub(regex, subst, value, 0, re.MULTILINE)
-            # replace multiple space
-            prepared_value = re.sub(" +", subst, prepared_value, 0, re.MULTILINE)
-            filter_prepared = (
-                field + ":" + re.sub(regex, subst, prepared_value, 0, re.MULTILINE)
+    def get(self, path: str = "/", headers: dict = None) -> dict:
+        try:
+            return super().get(path=path, headers=headers)
```
ahosgood, line 22: I don't think `path` should be variable here - it is defined by the API.

```suggestion
    def get(self, headers: dict = None) -> dict:
        try:
            return super().get(path="/", headers=headers)
```

etna/ciim/client.py
```diff
@@ -1,636 +1,70 @@
...
-    ):
-        self._hits = hits or []
-        self.total_count = total_count
-        self.item_type = item_type
-        self.aggregations = aggregations
-        self.bucket_counts = bucket_counts
-
-    @cached_property
-    def hits(self) -> Tuple["APIModel"]:
+    def get_record_instance(self, path: str = "/get") -> dict:
```
ahosgood, line 29: As above, I think the path should be hard-coded

etna/ciim/templatetags/records_tags.py
```diff
@@ -0,0 +1,20 @@
+from django import template
+
+register = template.Library()
+
+
+# TODO: Reduce cyclomatic complexity from 14 to 12 or below
+# flake8: noqa: C901
```
ahosgood, line 7: Is this still the case?

etna/core/api.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+
+class APIClientError(Exception):
+    def __init__(self, message, response=None):
+        super().__init__(message)
+        self.response = response
+
+
+class JSONAPIClient:
```
ahosgood, line 19: Is this the best place for this? It feels like the naming would suggest it is the Wagtail API.

Maybe move to something like `etna/core/lib.py` or `etna/core/json_api_client.py`?

etna/ciim/client.py
ahosgood: Oh I see - this is used by the other functions. Ignore this.

ahosgood (approved):

PR #1817, opened 2025-04-16
Put alias filtering before searching

jamesbiggs:
Allows removal of aliases before searching to fix the `?search` filter. Any filtering must go before the `SearchFilter` runs because it returns a `PostgresSearchResults` object.

Also splits out existing filter functions that are in the `listing_view` function into their own `Filter` objects for more readability and lowered cyclomatic complexity.

Also lowered cylomatic complexity on a few functions too!


etna/api/urls/pages.py
```diff
@@ -39,81 +63,11 @@ def listing_view(self, request):
         for restricted_page in restricted_pages:
             queryset = queryset.not_descendant_of(restricted_page, inclusive=True)
 
-        # Check if we have a specific site to look for
```
ahosgood, line 42: We might need this in the future if we need to serve content for multiple domains. Is there a reason it was removed?

ahosgood (approved):

PR #1822, opened 2025-04-29
Refactor HighlightImageSerializer to use CIIMClient for fetching record details

ahosgood:
Functionality changed in #1814

See http://localhost:8000/api/v2/pages/248/ - record fields on highlight images weren't getting populated.

I had to use the CIIM client to get the details for each.

Right now it makes a separate query for each which is SLOOOOOOW. Might need to talk to K-Int and get a new endpoint on the API where we can request a few records at a time, although I don't know that we can use that here.

This scenario _might_ be a good reason to cache record details...


etna/core/serializers/images.py
```diff
@@ -139,18 +141,17 @@ class HighlightImageSerializer(DetailedImageSerializer):
     def to_representation(self, value):
         representation = super().to_representation(value)
         if representation:
+            record = None
+            if value.record:
+                params = {
+                    "id": value.record,
+                }
+                client = CIIMClient(params=params)
```
jamesbiggs, line 149:
```suggestion
                client = CIIMClient(params = {"id": value.record})
```

jamesbiggs (commented):
Nice work AJ! All looks good but one tiny change (ignore it if you disagree though!)

jamesbiggs (approved):

PR #1833, opened 2025-05-28
Add redirect functionality for beta subdomain URLs to main domain

ahosgood:
For when we no longer have a need for the beta subdomain.

This should redirect all URLs under `/explore-the-collection/*` and `/people/*` to the live site.


jamesbiggs (approved):
🚀

PR #1839, opened 2025-06-11
What's On pages

jamesbiggs:


etna/api/urls/events.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class EventsAPIViewSet(CustomPagesAPIViewSet):
+    known_query_parameters = CustomPagesAPIViewSet.known_query_parameters.union(
+        ["online", "at_tna"]
+    )
+
+    filter_backends = [LocationFilter] + CustomPagesAPIViewSet.filter_backends
+
+    model = EventPage
```
ahosgood, line 13: This won't include displays, right? Is that what we want?

etna/core/models/location.py
```diff
@@ -0,0 +1,209 @@
...
+
+    image = models.ForeignKey(
+        "wagtailimages.Image",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("Image"),
+        help_text=_("An image representing the location, if applicable."),
+    )
```
ahosgood, line 79: Is this the only way of achieving this? It's producing a really undesirable and long `<select>` list in the admin.

jamesbiggs: Yeah, this will be the "Plan your visit" section that we have on Exhibition pages - this change will also need to be made there, not sure if you want to do that here too?

etna/api/urls/events.py
```diff
@@ -0,0 +1,13 @@
...
+from etna.whatson.models import EventPage
+
+
+class EventsAPIViewSet(CustomPagesAPIViewSet):
+    known_query_parameters = CustomPagesAPIViewSet.known_query_parameters.union(
+        ["online", "at_tna"]
+    )
+
+    filter_backends = [LocationFilter] + CustomPagesAPIViewSet.filter_backends
+
```
ahosgood, line 12: Should we only allow the listing path? We don't need the details path (because we can just use `/pages`) or the find path (we don't need to find events):

```suggestion
    @classmethod
    def get_urlpatterns(cls):
        return [
            path("", cls.as_view({"get": "listing_view"}), name="listing"),
        ]
```

etna/whatson/models.py
```diff
@@ -106,110 +195,175 @@ class AudienceType(models.Model):
...
+    """
+    A page for listing events.
+    """
 
-    panels = [
-        FieldPanel("name"),
-        FieldPanel("description"),
-        FieldPanel("image"),
+    # DataLayerMixin overrides
+    gtm_content_group = "What's On"
```
ahosgood, line 346: Don't think we need this any more

ahosgood (dismissed):

ahosgood (approved):

PR #1850, opened 2025-07-14
DM-35: Explore the Collection landing page updates

jamesbiggs:
Ticket URL: [DM-35]

## About these changes

Adds:
Hero image
"explorer_index_page_selections" - a way to add pages of any type at the bottom of the EtC landing page

These changes are mostly temporary in order to surface the blog and podcast at the bottom of the EtC page, before a wider rework is done

[DM-35]: https://national-archives.atlassian.net/browse/DM-35?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/core/models/location.py
+++ b/app/core/models/location.py
@@ -213,7 +213,7 @@ class LocationSerializer(serializers.Serializer):
 
     def to_representation(self, instance):
         if instance:
-            return {
+            representation = {
                 "space_name": instance.space_name,
                 "at_tna": instance.at_tna,
                 "online": instance.online,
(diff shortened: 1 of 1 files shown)


ahosgood:
Do we still need a "Heading" for the "Large card links" block? I don't think it's used right now.

ahosgood:
> > Do we still need a "Heading" for the "Large card links" block? I don't think it's used right now.
> 
> We're using it at the bottom of the Topic and Time period index pages, but they both have the same text. If you're happy to hard-code the heading text on those, I can remove it here?

I'm on about the one on the Etc landing page:

image

ahosgood (approved):

PR #1852, opened 2025-07-15
DM-59 + DM-68: Remove alt_text and deprecate blocks

jamesbiggs:
Ticket URL: [DM-59] and [DM-68]

## About these changes

Removes alt_text and decorative options for images (and remove unused replicated image blocks)

Removes "featured record article", "promoted links" from record article pages, and [...]


ahosgood (approved):
🚀

PR #1857, opened 2025-08-04
DM-49: PeopleListingBlock

jamesbiggs:
Ticket URL: [DM-49]

## About these changes

Re-works the logic for PersonPage roles - so that they're searchable in the database, rather than just class attributes.

Adds a new PeopleListingBlock which allows editors to select a role and display all People tagged with that role.

PersonRoles can be created to group people only in the backend (for use in PeopleListingBlock, and potentially other features in the future) so that the role never gets displayed to the public. This can be achieved by using the "Display on card" boolean, which will show (True) the role on any Person's card (if they have the role). The tests verify this by having an "Author" role with the "Display on card" set to True, and "Board member" set to False.

[DM-49]: https://national-archives.atlassian.net/browse/DM-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ahosgood (approved):
This all looks good to me and I don't _think_ there is anything to change on the frontend 🚀

PR #1946, opened 2026-04-08
EDEV-165: Feature/grouped block picker

rzarrouk:


jamesbiggs (commented):
Nice work @rzarrouk !

I'm happy to approve, but I have a suggestion which I'm happy to go either way on.

We can set these all at instance level as you've done here (i.e. `Block(group="xyz")`), but this will become quite hard to maintain as each time we use a block, we'll have to find which group it comes under. And then to change it, we'd have to update every place it's mentioned.

If we add `group="xyz"` to the `Meta` of a block, it is used in every instance of that class. For example:
```py
class QuoteBlock(blocks.StructBlock):
. . .
    class Meta:
        icon = "openquote"
        label = "Quote"
        group = "Structured and collapsible content"
```

I believe this also prevents any migrations needing to be created/run to update the group text. But the main benefit is having a single source of truth for the group text/label 😃

rzarrouk:
> Nice work @rzarrouk !
> 
> I'm happy to approve, but I have a suggestion which I'm happy to go either way on.
> 
> We can set these all at instance level as you've done here (i.e. `Block(group="xyz")`), but this will become quite hard to maintain as each time we use a block, we'll have to find which group it comes under. And then to change it, we'd have to update every place it's mentioned.
> 
> If we add `group="xyz"` to the `Meta` of a block, it is used in every instance of that class. For example:
> 
> ```python
> class QuoteBlock(blocks.StructBlock):
> . . .
>     class Meta:
>         icon = "openquote"
>         label = "Quote"
>         group = "Structured and collapsible content"
> ```
> 
> I believe this also prevents any migrations needing to be created/run to update the group text. But the main benefit is having a single source of truth for the group text/label 😃

I get ya - that makes sense - extra makes sense cause I had asked Lizzy if we wanted it to be the same for all pages because we could make them different, but she said for now it should all be the same 👍 

Ah. never pressed sent on this comment, but have just done this on a new branch so it's easier, will close this branch and get rid of PR. https://github.com/nationalarchives/ds-wagtail/tree/feature%2Fgrouped-block-picker-meta

PR #1947, opened 2026-04-09
EDEV-158: quote block citation url

rzarrouk:


app/core/blocks/quote.py
```diff
@@ -11,11 +54,42 @@ class QuoteBlock(blocks.StructBlock):
...
+            "attribution": source.get("attribution"),
+            "citation": source.get("citation"),
+            "citation_url": citation_link.get("citation_external_link")
+            or (internal_link.get("full_url") if internal_link else None),
+        }
 
     class Meta:
         icon = "openquote"
         label = "Quote"
+        source = AttributionCitationBlock()
```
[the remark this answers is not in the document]
rzarrouk: None issue, this is part of wagtail functionality.

app/core/blocks/quote.py
```diff
@@ -1,7 +1,50 @@
...
+        required=False,
+        max_length=100,
+        help_text="The name of the person being quoted. e.g. King Edward VIII",
+    )
+
+    citation = blocks.CharBlock(
+        required=False,
+        max_length=100,
+        help_text="The title of the work or document from which the quote is taken. e.g. Instrument of Abdication, 10 December 1936. Catalogue reference: PC 11/1",
+    )
```
jamesbiggs: I think it might be good to avoid including a catalogue reference as an example here as editors may start putting absolute links to Discovery/Catalogue (which can break as we have no control over them) - we have the `RecordChooser` block/field which does this for editors. Not sure if we want to include that as a link option here, potentially, but I feel it's probably best not to. Very minor nit-pick though and happy to leave it as is! 😃

app/core/blocks/quote.py
```diff
@@ -1,7 +1,50 @@
...
+
+    source_link = CitationLink()
+
+    class Meta:
+        icon = "edit"
+        collapsed = True
+        label_format = "{attribution+", " if attribution} {citation}"
```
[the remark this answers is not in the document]
jamesbiggs: I agree w/ the comment here, it should be just `"{attribution} {citation}"` as your original only displays `attribution`. Maybe `"{attribution}: {citation}"` or with a `-` between them? Or brackets around the whole thing might be good, as `Source` stays there so it ends up as `Source attribution citation`. Adding brackets could make it `Source (attribution citation)`

app/core/blocks/quote.py
jamesbiggs: I've not seen this before, what does it do? I tried removing it but I couldn't see that anything changed in the admin

app/core/blocks/quote.py
```diff
@@ -11,11 +54,42 @@ class QuoteBlock(blocks.StructBlock):
...
+        citation_link = source.get("source_link") or {}
+        internal_link = citation_link.get("citation_internal_link")
+
+        return {
+            "quote": representation.get("quote"),
+            "attribution": source.get("attribution"),
+            "citation": source.get("citation"),
+            "citation_url": citation_link.get("citation_external_link")
+            or (internal_link.get("full_url") if internal_link else None),
+        }
```
jamesbiggs, line 88: This works, but I think just `representation.get()` instead of `pop()` is better (we're not really re-using representation, so no need to pop it) and the conditional logic in the `return` might be better being done above, e.g.
```py
def get_api_representation(self, value, context=None):
    representation = super().get_api_representation(value, context)
    source = representation.get("source") or {}
    source_link = source.get("source_link") or {}

    citation_url = source_link.get("citation_external_link")
    if not citation_url:
        internal_link = source_link.get("citation_internal_link") or {}
        citation_url = internal_link.get("full_url")

    return {
        "quote": representation.get("quote"),
        "attribution": source.get("attribution"),
        "citation": source.get("citation"),
        "citation_url": citation_url,
    }
```
Just makes it a bit clearer (to me, at least!)

jamesbiggs (commented):
Nice work! Just a few comments for my own understanding and a few small changes 😃

app/core/blocks/quote.py
rzarrouk: This was given to me by Lizzy - I'll feed back to her and let you know :)

app/core/blocks/quote.py
rzarrouk: It nests the blocks if you add collapsable=True - they've got examples in the docs e.g. settings. I just think it's less visual clutter for the user, so added it for source and source link.

rzarrouk: https://guide.wagtail.org/en-latest/releases/new-in-wagtail-7-3/

See Custom editing layouts for block-based content

app/core/blocks/quote.py
rzarrouk: Ah yeah, thanks for catching that! I was trying to simplify things and reduce what was passed through because there was some nested bits before I just returned a new object.

And no probs - I try avoid extra assignments but it is clearer to read for sure :)

app/core/blocks/quote.py
rzarrouk: Although now you say that remobving it didn't change anything I'm questioning it - if I remove it it will change the structure of the data but maybe it's not strictly necessary if it doesn't change the UI since that's the point. 

I did build this incrementally so could have missed something like this, but was under the impression that it was what was keeping the source and attribute/citation blocks nested in the actual block

jamesbiggs: We can definitely do nested blocks, and collapse them, but I can't find an example (code) where this is done - I've not properly looked into that feature though (it was one of the things I mentioned a few weeks ago that is new and would be cool to add)

This could well be the right way to do it but I don't see a difference when deleting/adding it which is the only reason I questioned it 😆 

The data all stays the same in the API so no problem there either 🙂

jamesbiggs: https://docs.wagtail.org/en/stable/advanced_topics/customization/streamfield_blocks.html#customizing-the-initial-collapsible-state

I think this is what you're after

rzarrouk: Yup, that's the exact example I used- copied and changed it to add source as collapsable initially

app/core/blocks/quote.py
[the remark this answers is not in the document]
rzarrouk: I've left the brackets out of it (there's no way to keep them conditionally) - just replaced with {attr} {sourc}

app/core/blocks/quote.py
rzarrouk: talked about on a call, was thinking about the functionality of the line in the model - meta mention of source removed!

ahosgood (approved):

app/core/blocks/quote.py
rzarrouk: Updated on ticket and in branch 👍

PR #1959, opened 2026-04-23
ED-16: Feature/education skeleton pages

rzarrouk:
Ticket URL: [ED-16]
## - I found a couple more issues neither of us caught initially, fixed [here](https://github.com/nationalarchives/ds-wagtail/pull/1959/commits/29334dd6d4aef9447b7b2e39ec191bd30cd31ce5)

[ED-16]: https://national-archives.atlassian.net/browse/ED-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/education/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+#     EducationSessionsListingPage,
+#     EducationResourcesListingPage,
+# )
+
+# __all__ = [
+#     "EducationPage",
+#     "EducationSessionsListingPage",
+#     "EducationResourcesListingPage",
+# ]
```
jamesbiggs:
```suggestion
```

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        null=True,
+    )
+
+    panels = [
+        FieldPanel("key_stage"),
+        FieldPanel("curriculum_connection_description"),
+    ]
+
+
+class TeachingResourcePage(BasePageWithRequiredIntro):
```
jamesbiggs: Can we add https://github.com/nationalarchives/ds-wagtail/blob/b18d3ee70477eddf2263c57a77d19f22098fdc58/app/core/models/mixins.py#L64-L96 as a Mixin so we get `published_date`?

Conversation w/ Lizzy: https://nationalarchivesuk.slack.com/archives/C0A2BCNV6KF/p1777283842599459

This gives us the "new label" functionality that we use across the site. Sometimes editors like to publish privately, which sets the `first_published_at`, but it could be a month until they publish to the public. If we set the new label off of that, the pages might not get a "new" label 🙁 

`published_date` is also the field we chronologically order (most of) our pages on

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+            serializer=CurriculumConnectionSerializer(many=True),
+        ),
+        APIField("extension_activities"),
+        APIField("background_information"),
+        APIField("further_information_title"),
+        APIField("further_information"),
+    ]
+
+
+class EducationSessionPage(BasePageWithRequiredIntro):
```
jamesbiggs: See same comment about `published_date`

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+
+    class Meta:
+        verbose_name = _("Theme")
+        verbose_name_plural = _("Themes")
+
+    def __str__(self):
+        return self.name
+
+
+class QuestionBlock(blocks.StreamBlock):
```
jamesbiggs: These should blocks probably go in a `blocks.py` file under `education/` 😃

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                    required=False,
+                ),
+            ),
+        ],
+        icon="help",
+    )
+
+
+# same as MediaBlock but without the title
+class SourceMediaBlock(blocks.StructBlock):
```
jamesbiggs: If you want to remove a field, rather than replicating the block line-for-line, you can do this:

```suggestion
from app.media.blocks import MediaBlock

class SourceMediaBlock(MediaBlock):
    title = None
```

Not sure why it won't let me select the whole block! D'oh!

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # TODO: maybe remove cached_property? it will persist and maybe we don't want that
+    @cached_property
+    def latest_education_sessions(self) -> list:
+        upcoming = (
+            EducationSessionPage.objects.live()
+            .public()
+            .filter(start_date__gte=timezone.now())
+            .order_by("start_date")
+        )
```
jamesbiggs: I don't think this will make much difference aside from potentially only showing less than 3 sessions

If we're ordering them by `start_date`, we'll get the most "upcomingest" in cards 1-3 by default 🙂

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+
+    subpage_types = [
+        "education.EducationSessionsListingPage",
+        "education.TeachingResourcesListingPage",
+    ]
+
+    max_count = 1
+
+    # Teaching resources section
+    teaching_resource_listing_page = models.ForeignKey(
```
jamesbiggs:
```suggestion
    teaching_resources_listing_page = models.ForeignKey(
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("teaching resource listing page"),
+        help_text=_(
+            "The teaching resource listing page to display on the Education landing page."
+        ),
+    )
+
+    teaching_resources_teaser = models.TextField(
```
jamesbiggs: I'm not wedded to the name but should aim for consistency between the two (resource and session listing)

However we should use `CharField` in these instances, not `TextField` - `TextField` allows for multi-line inputs, `CharField` is just single line. I'd recommend setting the `max_length` to the same as the default `teaser_text` limit on `BasePage`

```suggestion
    teaching_resources_teaser_override = models.CharField(
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("featured teaching resource"),
+        help_text=_(
+            "Option to add a highlighted teaching resource, particularly for history months etc"
+        ),
+    )
+
+    featured_teaching_resource_teaser_override = models.TextField(
```
jamesbiggs:
```suggestion
    featured_teaching_resource_teaser_override = models.CharField(
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("education sessions listing page"),
+        help_text=_(
+            "The education sessions listing page to display on the Education landing page."
+        ),
+    )
+
+    education_sessions_teaser = models.TextField(
```
jamesbiggs:
```suggestion
    education_sessions_teaser_override = models.CharField(
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        "education.EducationSessionPage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("featured education session"),
+        help_text=_("Page picker to highlight a featured education session"),
+    )
+
+    featured_education_session_teaser_override = models.TextField(
```
jamesbiggs:
```suggestion
    featured_education_session_teaser_override = models.CharField(
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        MultiFieldPanel(
+            [
+                PageChooserPanel("teaching_resource_listing_page"),
+                FieldPanel("teaching_resources_teaser"),
+                MultiFieldPanel(
+                    [
+                        PageChooserPanel("featured_teaching_resource"),
+                        FieldPanel("featured_teaching_resource_teaser_override"),
+                    ],
+                    heading=_("Featured"),
```
jamesbiggs:
```suggestion
                    heading=_("Featured resource"),
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+        MultiFieldPanel(
+            [
+                PageChooserPanel("education_sessions_listing_page"),
+                FieldPanel("education_sessions_teaser"),
+                MultiFieldPanel(
+                    [
+                        PageChooserPanel("featured_education_session"),
+                        FieldPanel("featured_education_session_teaser_override"),
+                    ],
+                    heading=_("Featured"),
```
jamesbiggs:
```suggestion
                    heading=_("Featured session"),
```

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        "wagtailcore.Page",
+        on_delete=models.CASCADE,
+        related_name="+",
+        verbose_name=_("selected page"),
+        help_text=_("Navigation to other sections within Education"),
+    )
+
+    panels = [
+        PageChooserPanel("selected_page"),
+    ]
```
jamesbiggs, line 204: If we want these to only have "education" links, you can specify multiple page types in `PageChooserPanel`, seen here: https://docs.wagtail.org/en/latest/reference/panels.html#pagechooserpanel

At current I can select any page across the site (or other sites). Could be okay, could be bad! Maybe another one for Lizzy

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+    @cached_property
+    def type_label(cls) -> str:
+        return "Education"
+
+    @cached_property
+    def latest_teaching_resources(self) -> list:
+        return list(
+            TeachingResourcePage.objects.live()
+            .public()
+            .order_by("-first_published_at")[:3]
```
jamesbiggs:
```suggestion
            .order_by("-published_date")[:3]
```

app/education/models/landing.py
```diff
@@ -0,0 +1,201 @@
...
+                    heading=_("Featured"),
+                ),
+            ],
+            heading=_("Education sessions"),
+        ),
+        InlinePanel(
+            "education_read_more_links",
+            heading=_("Read more"),
+            help_text=_("Navigation to other sections within Education"),
+        ),
```
jamesbiggs, line 167: Not sure if we want a max limit on here? They could get carried away 😅

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("featured teaching resource"),
+        help_text=_(
+            "Option to add a highlighted teaching resource, particularly for history months etc"
+        ),
+    )
+
+    featured_teaching_resource_teaser_override = models.TextField(
```
jamesbiggs:
```suggestion
    featured_teaching_resource_teaser_override = models.CharField(
```

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+        [("web_archive_promo", APIRichTextBlock(features=["bold", "italic", "link"]))],
+        verbose_name=_("web archive promo text"),
+        help_text=_(
+            "Text block highlighting that old resources can be found in the web archive with a link to the archived old Education site. "
+        ),
+        blank=True,
+        null=True,
+    )
+
+    newsletter_sign_up_text = models.TextField(
```
jamesbiggs: This may need some more thought, I don't think we've got a solution for this yet but we should probably discuss with the team. There was some discussion about a newsletter component, but I don't remember if we agreed anything.

Basically - it's more than just text, but how it functions is a different issue :)

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+        help_text=_(
+            "Text block encouraging users to sign up for the education newsletter."
+        ),
+        blank=True,
+    )
+
+    content_panels = BasePageWithRequiredIntro.content_panels + [
+        MultiFieldPanel(
+            [
+                PageChooserPanel("featured_teaching_resource"),
```
jamesbiggs, line 53: You can use `PageChooserPanel` here or just `FieldPanel`. Personally I default to `FieldPanel` if we're specifying the page model on the `ForeignKey` - they work the exact same in the CMS 🙂

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+        "education.EducationSessionPage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("featured education session"),
+        help_text=_("Page picker to highlight a featured education session"),
+    )
+
+    featured_education_session_teaser_override = models.TextField(
```
jamesbiggs:
```suggestion
    featured_education_session_teaser_override = models.CharField(
```

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+        blank=True,
+    )
+
+    newsletter_sign_up_text = models.TextField(
+        verbose_name=_("newsletter sign up text"),
+        help_text=_(
+            "Text block encouraging users to sign up for the education newsletter."
+        ),
+        blank=True,
+    )
```
jamesbiggs: See comment on newsletters

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+
+    content_panels = BasePageWithRequiredIntro.content_panels + [
+        MultiFieldPanel(
+            [
+                PageChooserPanel("featured_education_session"),
```
jamesbiggs, line 104: See comment on `PageChooserPanel`s

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class EducationSessionsListingPage(BasePageWithRequiredIntro):
+    """
+    A page for displaying education sessions.
+    """
+
+    @cached_property
+    def type_label(cls) -> str:
+        return "Education Session"
```
jamesbiggs: I'm not sure if we want a supertitle on these pages, there's not one on the designs but maybe best to check with AJ, if you leave it blank it will default to `None` 👍

app/education/models/listings.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class TeachingResourcesListingPage(BasePageWithRequiredIntro):
+    """
+    A page for displaying education/teaching resources.
+    """
+
+    @cached_property
+    def type_label(cls) -> str:
+        return "Education Resource"
```
jamesbiggs: I'm not sure if we want a supertitle on these pages, there's not one on the designs but maybe best to check with AJ, if you leave it blank it will default to `None` 👍

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                    verbose_name=_("hero image"),
+                    blank=True,
+                ),
+            )
+        ],
+        null=True,
+        max_num=1,
+    )
+
+    enquiry_question = models.TextField(
```
jamesbiggs:
```suggestion
    enquiry_question = models.CharField(
```

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                ContentImageBlock(
+                    rendition_size="max-900x900",
+                    verbose_name=_("hero image"),
+                    blank=True,
+                ),
+            )
+        ],
+        null=True,
+        max_num=1,
+    )
```
jamesbiggs: We have the `HeroImageMixin` for this one

https://github.com/nationalarchives/ds-wagtail/blob/905dbef9c6b476d593288ebcdc37fe62e6b1acc2/app/core/models/mixins.py#L99-L142

or `RequiredHeroImageMixin` if every page must have a hero (check w/ Lizzy)

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+
+    enquiry_question = models.TextField(
+        verbose_name=_("enquiry question"),
+        blank=True,
+    )
+
+    introduction_text = models.TextField(
+        verbose_name=_("introduction text"),
+        blank=True,
+    )
```
jamesbiggs: We already have intro text from `BasePageWithRequiredIntro`

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+
+    theme = models.ForeignKey(
+        "education.Theme",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("theme"),
+    )
```
jamesbiggs: I think these should be `InlinePanel` as this only allows us to pick one, but there's the need for many to be selected.

In pseudo-ish-code:
```python
class KeyStage(models.Model):
stage = integerfield or charfield (2, 3, 4, etc)
age_range = charfield (5-7, 7-11, etc) # I doubt we'll need to do anything maths-y so it's fine to have it just as text
slug = charfield

@cached_property
def public_key_stage():
return f"Key stage {stage}"

@cached_property
def short_key_stage():
return f"KS{stage}"

@cached_property
def public_age_range():
return f"Ages {age_range}"

class KeyStageTag(models.Model):
page = parentalkey of wagtailcore.Page (relationship "education_keystage_tags" or similar name)
key_stage = foreignkey of KeyStage

class TeachingResourcePage():
in promote_panels: InlinePanel("education_keystage_tags")
```
Similar applies to Topic and Theme 🙂

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+
+    # Body
+    sources_title = models.TextField(
```
jamesbiggs:
```suggestion
    sources_title = models.CharField(
```

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+            (
+                "sources_introduction",
+                APIRichTextBlock(features=["bold", "italic", "link"]),
+            )
+        ],
+        verbose_name=_("sources introduction"),
+        help_text=_("Optional text field to provide an introduction to the sources."),
+        blank=True,
+        null=True,
+    )
```
jamesbiggs: Just an optional `RichTextField` will do here 🙂

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                APIRichTextBlock(features=["bold", "italic", "link", "ol", "ul"]),
+            )
+        ],
+        verbose_name=_("teachers notes"),
+        help_text=_(
+            "A general overview of what the resource contains and how it can be used."
+        ),
+        blank=True,
+        null=True,
+    )
```
jamesbiggs: Depending on how this works, it could be fine with being just a `RichTextField`, however looking at the designs it looks a bit more involved than that. Might be something we need to iron out 🤔

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        help_text=_("Section providing historical context to the teaching resource."),
+        blank=True,
+        null=True,
+    )
+
+    # Further information
+    further_information_title = models.CharField(
+        max_length=255,
+        verbose_name=_("further information title"),
+        help_text=_("Title of the further information section (required)."),
```
jamesbiggs: I don't think we should put `(required)` if it's not required (or even if it is required), if they leave it blank we can just default to "Further Information" (or similar) in the frontend

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        null=True,
+    )
+
+    content_panels = BasePageWithRequiredIntro.content_panels + [
+        FieldPanel("hero_image"),
+        FieldPanel("enquiry_question"),
+        FieldPanel("introduction_text"),
+        FieldPanel("key_stage"),
+        FieldPanel("time_period"),
+        FieldPanel("theme"),
```
jamesbiggs: These are best placed in the `promote_panels`

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+            [
+                InlinePanel(
+                    "curriculum_connections",
+                    label=_("Curriculum connection"),
+                    heading=_("Connections to the curriculum"),
+                    min_num=1,
+                ),
+            ],
+            heading=_("Connections to the curriculum"),
+        ),
```
jamesbiggs: Having `min_num=1` means you _have_ to put one. They might not want to put one on a page, potentially, meaning this breaks that

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                    "sources",
+                    label=_("Source"),
+                    heading=_("Sources"),
+                    min_num=1,
+                ),
+            ],
+            heading=_("Sources"),
+        ),
+        FieldPanel("teachers_notes"),
+        MultiFieldPanel(
```
jamesbiggs: There's a `MultiFieldPanel` with just one `InlinePanel` in it - I have a feeling `teachers_notes` was meant to be in here too 🙂 (I'd also change the name of the `MultiFieldPanel.heading` to cover both, if so)

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+    )
+
+    api_fields = BasePageWithRequiredIntro.api_fields + [
+        APIField("key_stage", serializer=KeyStageSerializer()),
+        APIField("time_period", serializer=TimePeriodSerializer()),
+        APIField("theme", serializer=ThemeSerializer()),
+    ]
+
+
+class EducationReadMoreLink(Orderable):
```
jamesbiggs: Best to move this to `landing.py` as it's only used there

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+    )
+
+    theme = models.ForeignKey(
+        "education.Theme",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        verbose_name=_("theme"),
+    )
```
jamesbiggs: Same as on the resource page

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        "education.TeachingResourcePage",
+        on_delete=models.CASCADE,
+        related_name="sources",
+    )
+
+    source_title = models.TextField(
+        verbose_name=_("source title"),
+        help_text=_("A unique, descriptive title for the source."),
+        blank=True,
+    )
```
jamesbiggs: I think this will come from the selected image data, as we'll be linking into the "image page" - so they should match

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+                "source_image",
+                APIImageChooserBlock(
+                    rendition_size="max-900x900",
+                    verbose_name=_("source image"),
+                    help_text=_("An image for the source."),
+                    blank=True,
+                ),
+            )
+        ],
+    )
```
jamesbiggs: Just a `ForeignKey` to the image model will do here

jamesbiggs: Ah, actually, I've just seen it's a mix of image or media. This might need some further thought/discussion w/ Lizzy/team

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        context["src"] = value["media"].sources[0]["src"]
+        context["type"] = value["media"].sources[0]["type"]
+        return context
+
+    @property
+    def admin_label(self):
+        return self.meta.label
+
+
+class Source(Orderable):
```
jamesbiggs: I've got quite a few questions/concerns about the proposed implementation (design/spec-wise) of this section so I'll ask the team and come back to this one

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+        related_name="curriculum_connections",
+    )
+
+    key_stage = models.ForeignKey(
+        "education.KeyStage",
+        on_delete=models.CASCADE,
+        related_name="+",
+        verbose_name=_("Key stage"),
+        help_text=_("The key stage for this curriculum connection."),
+        limit_choices_to={"slug__in": KEY_STAGE_ALLOWED_SLUGS},
```
jamesbiggs: I don't think the `limit_choices_to` is necessary as we dictate which `KeyStage` instances will exist

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+KEY_STAGE_NAME_CHOICES = [(choice.label, choice.label) for choice in KeyStageChoices]
+
+
+class KeyStage(models.Model):
+    """A model for individual key stage tags"""
+
+    name = models.CharField(
+        max_length=255,
+        verbose_name=_("name"),
+        choices=KEY_STAGE_NAME_CHOICES,
```
jamesbiggs: This won't be needed as `KeyStage` will be set via migrations, so we can just procedurally make these key stages in a migration - no need to create choices etc

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+from ..serializers import (
+    CurriculumConnectionSerializer,
+    KeyStageSerializer,
+    SourceSerializer,
+    ThemeSerializer,
+    TimePeriodSerializer,
+)
+
+
+class KeyStageChoices(models.TextChoices):
```
jamesbiggs: https://github.com/nationalarchives/ds-wagtail/pull/1959/changes#r3147809685

app/education/models/details.py
```diff
@@ -0,0 +1,611 @@
...
+            (
+                "curriculum_connection_description",
+                APIRichTextBlock(features=["bold", "italic", "link", "ul"]),
+            )
+        ],
+        verbose_name=_("curriculum connection description"),
+        help_text=_("Add the curriculum connection description."),
+        blank=True,
+        null=True,
+    )
```
jamesbiggs: This can just be `RichTextField`

app/education/serializers.py
```diff
@@ -0,0 +1,80 @@
+from app.core.serializers import DefaultPageSerializer
+from rest_framework import serializers
+
+
+class _StreamFieldRepresentationMixin:
```
jamesbiggs, line 6: I see what this is doing but not sure it's needed/necessary - but let's do the page content first and then see what needs serialising etc 🙂

jamesbiggs (changes requested):
Hi @rzarrouk, I've left comments for stuff that will need changing - some may need further discussion/review from the team. I'm posting some questions in Slack (mainly for my own understanding, but they'll help inform my review better 😃)

jamesbiggs (changes requested):
Hi @rzarrouk, I've left comments for stuff that will need changing - some may need further discussion/review from the team. I'm posting some questions in Slack (mainly for my own understanding, but they'll help inform my review better 😃)

app/education/models/details.py
jamesbiggs: Questions asked and answered here: https://nationalarchivesuk.slack.com/archives/C0A2BCNV6KF/p1777301440288509

Let's have a meeting and map the requirements out 🙂

app/education/models/details.py
rzarrouk: 🙏

app/education/models/details.py
rzarrouk: that's a D'oh for me! danke

app/education/models/details.py
rzarrouk: sorry, last min addition when re-reading and checking the spec, didn't think!

app/education/models/listings.py
rzarrouk: I'd showed Lizzy everything on a call last week and she was happy with these just as placeholders more than anything - i don't think it's been ironed out unless there was more talk on that since I left.

app/education/models/listings.py
rzarrouk: Undecided what's happening with these so leaving it for now aside from changing to CharField

app/education/models/details.py
rzarrouk: I"ve potentially misunderstood the specs as anything with asterix is required since other things say optional explicitly

Screenshot 2026-05-05 at 13 53 18

app/education/models/details.py
rzarrouk: Lizzy's reply to your query on this says teacher's notes should be decoupled so I've left this out of it - unless I've misunderstood

I'll change the type of panel so it's not MultiField :))

app/education/models/details.py
rzarrouk: Again re. Lizzy's comment seems this is okay as is? 
If I've misunderstood or you think we should have a fallback please do say though

app/education/models/landing.py
rzarrouk: Maybe 3 - will double check with Lizzy

app/education/models/listings.py
rzarrouk: Removed all references to this - spoke on call, it's handled in the frontend

app/education/models/details.py
rzarrouk: Added these but will sort the order of promote panels seems a bit messy to me atm but not a priority and not too difficult to do

app/education/models/landing.py
rzarrouk: Refactored and removed this part, though in my head we need a fallback or a way of it being set more specifically for a date range maybe? sounds overkill but was thinking about and maybe easier to talk through cause I think am being silly and I think I've just overthought it - but then it's like how much is handled or will be by fronend I think is what i need to decouple

app/education/models/landing.py
rzarrouk: I changed this, but I've now removed it because it was duplicatig headings and I think it was ugly?

app/education/models/details.py
rzarrouk: Left this one for now cause I think I wanna sit down and quick chat panels and things so I make sure I have it all straight in my head - but I understand :)

rzarrouk: I think what I'm missing is just being like this is what it is in the backend and this is what it looks like in the UI - I keep mixing bits up

app/education/models/landing.py
rzarrouk: Added mixin so this is available

app/education/models/listings.py
rzarrouk: Changed for consistency - I thought it added readability and if wagtail update it as a thing then maybe nicer functionality? But I suppose it doesn't matter

app/education/models/listings.py
rzarrouk: Removed all references, as mentioned above

app/education/models/listings.py
rzarrouk: Changed as above for consistency - I thought it added readability and if wagtail update it as a thing then maybe nicer functionality? But I suppose it doesn't matter

app/education/models/listings.py
rzarrouk: Removed, easy enough to add if we need one - this was more of a hangover from copying from whatson tbh

app/education/models/listings.py
rzarrouk: Removed, easy enough to add if we need one - this was more of a hangover from copying from whatson tbh

app/education/models/details.py
rzarrouk: I've used RequiredHeroImageMixin for the page causeit was required in spec remade migrations

app/education/models/details.py
rzarrouk: These were honestly just placeholders/starting points because at the time we were just talking about tags - i had spoken to lizzy who had a separate thing in the promote tag for tags too. I think it'd be good to talk through because it's something I thought about about a month ago now and I'd looked through tags and their uses in whatson etc - lizzy had voiced it being better to have things like dropdowns rather than free text to find things and I wanted to talk more about it in general, see what options we had and what standardisation would looklike across the system with the new taxonomies that were being discussed/decided

app/education/models/details.py
rzarrouk: See comment above - would be good to chat about, not something i had clear in my head at the time

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+        APIField("sources_title"),
+        APIField("sources_introduction", serializer=RichTextSerializer()),
+        APIField("teachers_notes"),
+        APIField("sources", serializer=SourceSerializer(many=True)),
+        APIField("curriculum_connection_description", serializer=RichTextSerializer()),
+        APIField("extension_activities"),
+        APIField("background_information"),
+        APIField("further_information_title"),
+        APIField("further_information"),
+    ]
```
jamesbiggs: We're missing some api_fields (and default_api_fields) but we can sort these out/decide on these after finishing the admin 🙂

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+
+
+class EducationSessionPage(BasePageWithRequiredIntro, PublishedDateMixin):
```
jamesbiggs: There are a few instances where you've inherited `BasePage[...]` first - it should come last for method resolution order reasons 🙂

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+        verbose_name=_("sources title"),
+        help_text=_(
+            "Title of the main section of the page. In most cases ‘Investigate the sources’"
+        ),
+        blank=True,
+        max_length=160,
+    )
+
+    sources_introduction = RichTextField(
+        features=["bold", "italic", "link"],
```
jamesbiggs: We can use `INLINE_RICH_TEXT_FEATURES` via `from django.conf import settings` (which we set in `production.py`) though I see `INLINE_RICH_TEXT_FEATURES` has `code` in it now, so we might want to make `DEFAULT_RICH_TEXT_FEATURES` and just have that as `bold, italic, link` and then make `INLINE_RICH_TEXT_FEATURES` extend the above.

We can do that later though, but I think we should remove `code` from the `INLINE_RICH_TEXT_FEATURES` at some point (wider task so don't worry about it here)

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+        blank=True,
+        null=True,
+    )
+
+    # Teacher’s Notes*
+    teachers_notes = StreamField(
+        [
+            (
+                "teachers_notes",
+                APIRichTextBlock(features=["bold", "italic", "link", "ol", "ul"]),
```
jamesbiggs: Here we could use `RESTRICTED_RICH_TEXT_FEATURES` though 🙂 (and any other places we set `features=`)

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+        verbose_name=_("slug"),
+        unique=True,
+    )
+
+    class Meta:
+        verbose_name = _("Theme")
+        verbose_name_plural = _("Themes")
+
+    def __str__(self):
+        return self.name
```
jamesbiggs, line 150: DISCUSS ON CALL: This would be fine if we're just tagging with one theme etc, but we want the ability to do multiple - so a middle model is needed, e.g.
XPage <-> XYTag <-> YModel

This current set up only allows for:
XPage - YModel

app/education/models/details.py
```diff
@@ -0,0 +1,510 @@
...
+from ..serializers import (
+    KeyStageSerializer,
+    SourceSerializer,
+    ThemeSerializer,
+    TimePeriodSerializer,
+)
+
+# TODO: sort panel order on promote tabs
+
+
```
jamesbiggs, line 11: We might want to add a bit more data to the KS/TPs, e.g. `age_range`, `stage`, etc. So we can take different bits of data and use them differently across the site. There are a few examples on the designs where we just display "KS4" or "KS4 (Ages X-Y)" or "Key Stage 4" etc. By having just `name` we can't programmatically generate those examples. Same with TPs - `date_from` and `date_to` might be useful at some point, so probably best to add it now.

app/education/models/details.py
jamesbiggs: Let's discuss this model on a call - it will probably be better as fields vs StreamFields, but will be good to discuss

jamesbiggs (changes requested):
A couple of bits still need ironing out but otherwise looking good! We'll have a call to discuss the bigger bits 😎

app/education/models/details.py
jamesbiggs: Discussed on call: Just one StreamField containing Image, Media and Youtube (sources can have multiple "sources")

app/education/models/details.py
```diff
@@ -0,0 +1,910 @@
...
+        ],
+    )
+
+    source_media_caption = StreamField(
+        [("source_media_caption", APIRichTextBlock(features=["bold", "italic"]))],
+        verbose_name=_("source caption"),
+        help_text=_("If provided, displays directly below the source."),
+        blank=True,
+        null=True,
+    )
```
jamesbiggs: Source(orderable)

source_media = StreamField
([
PICK ONE: [ImageBlock,
MediaBlock,
YoutubeBlock,],
ALWAYS AFTER: CaptionBlock
])

app/education/models/details.py
```diff
@@ -0,0 +1,910 @@
...
+
+    key_stage = models.ForeignKey(
+        "education.KeyStage",
+        on_delete=models.CASCADE,
+        related_name="+",
+        verbose_name=_("Key stage"),
+        help_text=_("The key stage for this curriculum connection."),
+    )
+
+    curriculum_connection_description = RichTextField(
```
jamesbiggs:
```suggestion
    description = RichTextField(
```

app/education/models/listings.py
```diff
@@ -0,0 +1,129 @@
...
+
+    featured_teaching_resource_teaser_override = models.CharField(
+        verbose_name=_("Featured teaching resource teaser text override"),
+        help_text=_("Override text for the featured teaching resource"),
+        blank=True,
+        max_length=160,
+    )
+
+    web_archive_promo = StreamField(
+        [("web_archive_promo", APIRichTextBlock(features=["bold", "italic", "link"]))],
```
jamesbiggs: Hard-code in FE

app/education/models/details.py
rzarrouk: TODO Ask Lizzy

app/education/models/listings.py
rzarrouk: REmoved

app/education/models/details.py
rzarrouk: Changed, have lef tthe other one as explicit which feels weird tbh but since this one was inside the model it makes sense to me.

app/education/models/details.py
rzarrouk: Done this, also did a similar thing for the links to make it a bit cleaner for the user

app/education/models/details.py
rzarrouk: Done this, as above/discussed

app/education/migrations/0002_seed_key_stages.py
```diff
@@ -0,0 +1,113 @@
...
+# Tuple: (slug, name, date_from, date_to, available_for_resources)
+TIME_PERIODS = [
+    ("early-civilisations-pre-900", "Early civilisations (pre 900)", None, date(900, 1, 1), False),
+    ("medieval-900-1485", "Medieval (900-1485)", date(900, 1, 1), date(1485, 12, 31), True),
+    ("early-modern-1485-1750", "Early modern (1485-1750)", date(1485, 1, 1), date(1750, 12, 31), True),
+    ("industrial-revolution-1750-1901", "Industrial revolution (1750-1901)", date(1750, 1, 1), date(1901, 12, 31), True),
+    ("early-twentieth-century-1901-1945", "Early Twentieth Century (1901-1945)", date(1901, 1, 1), date(1945, 12, 31), True),
+    ("post-war-and-modern-1945-present", "Post-War and modern (1945- present)", date(1945, 1, 1), None, True),
+    ("cross-period", "Cross period", None, None, True),
+]
```
jamesbiggs, line 41: Question: Is there a need to have these as dates? Probably simpler to just have them as integers and the "year" is the value (I don't think we'd want to get as granular as days/months but I may be wrong!)

app/education/models/sessions.py
```diff
@@ -0,0 +1,575 @@
...
+        default=False,
+        verbose_name=_("At your school"),
+        help_text=_("Check if this is at a school."),
+    )
+
+    location_name = models.CharField(
+        max_length=255,
+        verbose_name=_("location name"),
+        help_text=_("Venue name"),
+    )
```
jamesbiggs: I think this should have `blank=True` - if it's at TNA/online/at your school, we don't need to write that (unless I'm missing some other functionality)

app/education/models/sessions.py
```diff
@@ -0,0 +1,575 @@
...
+        FieldPanel("postcode"),
+        FieldPanel("duration"),
+    ]
+    
+    class Meta:
+        verbose_name = _("session location")
+        ordering = ["sort_order"]
+    
+    def __str__(self):
+        return self.location_name
```
jamesbiggs: `location_name` may be None/empty (if we make `blank=True`), so fallbacks like
```
if self.location_name:
return location_name
if self.at_tna:
return "At The National Archives
...
```

app/education/models/sessions.py
```diff
@@ -0,0 +1,575 @@
...
+
+
+class EducationSessionPageThemeTag(BaseThemeTag):
+    page = ParentalKey(
+        "education.EducationSessionPage",
+        on_delete=models.CASCADE,
+        related_name="education_theme_tags",
+    )
+
+class SessionLocation(Orderable):
```
jamesbiggs, line 63: Some validation to prevent someone saying a location is online, at TNA, at your school etc all in one is needed 🙂

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+                ),
+            ),
+        ],
+        verbose_name=_("source media"),
+        help_text=_(
+            "Choose one media type for this source. A caption can be added for each."
+        ),
+        blank=True,
+        null=True,
+    )
```
jamesbiggs: When we have big/complex blocks like this, it's better (in my opinion) to have a StreamField that uses a block we construct (usually in blocks.py), similar to this:
https://docs.wagtail.org/en/7.1/reference/streamfield/blocks.html#wagtail.blocks.StreamBlock where it says "As with StructBlock, the list of sub-blocks can also be provided as a subclass of StreamBlock:"

Makes it more readable, and personally, easier to edit/maintain.

```py
class SourceMediaBlock(blocks.StreamBlock):
    source_image = SourceImageWithCaptionBlock()
    source_video = SourceMediaWithCaptionBlock()
    source_youtube = SourceYouTubeWithCaptionBlock()

    class Meta:
        label = "Source media"
```

In this example we also create a `StructBlock` with each field in it, so we can have a caption for each "media"
```py
class SourceImageWithCaptionBlock(blocks.StructBlock):
    source_image = APIImageChooserBlock(
        rendition_size="max-900x900",
        help_text="An image for the source.",
    )
    caption = APIRichTextBlock(features=["bold", "italic"], required=False)
```

Your `source_media` StreamField then becomes:
```py
    source_media = StreamField(
        SourceMediaBlock(),
        verbose_name=_("source media"),
        help_text=_(
            "Choose one media type for this source. A caption can be added for each."
        ),
        blank=True,
        null=True,
    )
```

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+
+    # Source link — choose internal or external
+    source_media_featured_link = StreamField(
```
jamesbiggs: Same as above - best to use a class version of the blocks

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+        blank=True,
+        null=True,
+    )
+
+
+    # Source questions
+    source_question = StreamField(
+        QuestionBlock(
+            verbose_name=("source question"),
+            help_text=("A series of questions relating to each source."),
```
jamesbiggs: If one source can have many questions:
```suggestion
            help_text=("A series of questions relating to the source."),
```

If not, we need to make this so you can only have one question per source (in which case I'd opt for just fields on the model, rather than a QuestionBlock StreamField

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+
+class TeachingResourcePageThemeTag(BaseThemeTag):
+    page = ParentalKey(
+        "education.TeachingResourcePage",
+        on_delete=models.CASCADE,
+        related_name="education_theme_tags",
+    )
+
+
+class Source(Orderable):
```
jamesbiggs, line 76: You've got a few fields on here with `source_`, no need for that (double naming, e.g. `Source.source_title` should just be `Source.title`)🙂

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+    def key_stage(self):
+        return self.key_stages[0] if self.key_stages else None
+
+    @cached_property
+    def time_period(self):
+        return self.time_periods[0] if self.time_periods else None
+
+    @cached_property
+    def theme(self):
+        return self.themes[0] if self.themes else None
```
jamesbiggs: Duplicated code between Resources and Sessions would be best kept in a Mixin, either individual mixins for theme, time period, and keystage - or all in one. I think having separate mixins could be better but I don't mind either way.

app/education/models/resources.py
```diff
@@ -0,0 +1,457 @@
...
+            InlinePanel(
+                "education_time_period_tags",
+                label=_("Time period"),
+                heading=_("Time periods"),
+            ),
+            InlinePanel(
+                "education_theme_tags",
+                label=_("Theme"),
+                heading=_("Themes"),
+            ),
```
jamesbiggs: Same as the properties above, these should go in Mixins and be inherited by Resources and Sessions 🙂

app/education/models/sessions.py
```diff
@@ -0,0 +1,575 @@
...
+                        "Inset text - TODO ask for better decsription of this? where it sits etc"
+                    ),
+                    blank=True,
+                    null=True,
+                )
+            )
+        ],
+        blank=True,
+        null=True
+    )
```
jamesbiggs: I'm not sure on the spec of this but I feel these should all be in one StreamBlock so they can be chosen in any order. If they need to be in a specific order/structure, some of them should be just fields (e.g. partner_logo should be PartnerLogoField)

I think the aim is for them to be in one StreamBlock though. Sorry if I've missed any that should also be in here!

jamesbiggs (changes requested):

app/education/models/details.py
rzarrouk: To confirm - I spoke to Lizzy about this and it's likely they will want to be able to choose other pages accross the site.

app/education/models/landing.py
rzarrouk: Spoke to Lizzy - no limit.

app/education/models/details.py
rzarrouk: Spoke to Lizzy - yes always minimum 1 - if they didn't have a connection to the curriculum then they shouldn't be in education.

app/education/models/details.py
rzarrouk: I think I've done this for the most part - I'd assume better to have a list of features than use restricted minus XYZ for others.

app/education/models/sessions.py
rzarrouk: I've changed how this operates but am using required=False and a validation

app/education/models/sessions.py
rzarrouk: Yep is in new commits

app/education/models/resources.py
rzarrouk: removed

app/education/models/resources.py
rzarrouk: I've copied them into EducationTaxonomyMixin with the promote panels too

app/education/models/resources.py
rzarrouk: As above now in EducationTaxonomyMixin

app/education/models/sessions.py
rzarrouk: Please look again I changed this before - also this is me beoing really silly cause I copied the spec into my editor to write it bit by bit using comments as a checklist and didn't realise till I went back to check

app/education/migrations/0002_seed_key_stages.py
rzarrouk: I should have commented re why but also actually yeah it doesn't need it! This was from what you said and something rose suggested to lizzy  re future use of resources with associated dates - but also i agree tbf, if thinsg will have dates in the future for associated timelines etc then categorising time periods by year shouldn't be affected anyway, so am with ya (Y)

app/education/models/resources.py
rzarrouk: God yeah that's so much nicer, I did it with media cause it was already there and didn't think

app/education/models/resources.py
rzarrouk: Done, same as above

app/education/models/resources.py
rzarrouk: Changed wording

app/education/models/details.py
rzarrouk: re. date from date to was this +rose+lizzy thinking about future stuff but not in spec and dealable with - will resolve comment as other comments do deal with this anyway

app/education/models/details.py
rzarrouk: done

app/education/models/sessions.py
```diff
@@ -0,0 +1,395 @@
...
+                    "price_detail": _(
+                        "Price detail is required when a session price is specified."
+                    )
+                }
+            )
+        if self.price is None and self.price_detail:
+            raise ValidationError(
+                {"price": _("Price is required when price detail is specified.")}
+            )
+
```
jamesbiggs, line 335: Sorry, one last change! Validation that start_date is before end_date
```suggestion
if (
    self.start_date is not None
    and self.end_date is not None
    and self.start_date >= self.end_date
):
    raise ValidationError(
        {"end_date": _("End date must be later than start date.")}
    )
```

jamesbiggs (approved):
🌟 ⭐ 🌟 

Great work @rzarrouk !

Just one change which I noticed last minute, but otherwise approved! 🙂

ahosgood (approved):

jamesbiggs (approved):

PR #1961, opened 2026-04-23
Fix/Add thumbnail to admin_form_fields

rzarrouk:
Ticket URL: https://national-archives.atlassian.net/browse/EDEV-173


ahosgood:
This works in the sense that the thumbnail field is now available in the admin area.

I think we also need to remove "Thumbnail" from the media block (see below)

image

Then we need to expose the same image rendition in the API from the thumbnail added to the media item rather than the media block.

ahosgood:
https://github.com/torchbox/wagtailmedia/issues/196

PR #1972, opened 2026-05-15
Ruff formatting and linting fixes

ahosgood:


ahosgood:
@jamesbiggs - How much trust do you have in our tests?

app/core/management/commands/rename_app.py
ahosgood: Do we still even need this?

config/settings/util/__init__.py
```diff
@@ -7,7 +7,6 @@ def strtobool(val: str) -> bool:
     val = val.lower()
     if val in ("y", "yes", "t", "true", "on", "1"):
         return True
-    elif val in ("n", "no", "f", "false", "off", "0"):
+    if val in ("n", "no", "f", "false", "off", "0"):
         return False
-    else:
-        raise ValueError("invalid truth value %r" % (val,))
+    raise ValueError("invalid truth value %r" % (val,))
```
ahosgood: Consider replacing with TNA Python Utilities https://github.com/nationalarchives/python-utilities/blob/main/tna_utilities/__init__.py

Import would be:

```python
from tna_utilities import strtobool
```

app/core/json_api_client.py
```diff
@@ -21,18 +21,18 @@ class JSONAPIClient:
     A client for interacting with a JSON API, which can be extended for specific APIs.
     """
```
ahosgood, line 22: Consider replacing with TNA Python Utilities `SimpleJsonApiClient`: https://nationalarchives.github.io/python-utilities/api/

rzarrouk:
I'm happy to pass this as is - re. suggested improvements they can be done after the fact.

app/api/urls/pages.py
```diff
@@ -184,7 +183,7 @@ def get_base_queryset(self):
         "depth",
     ]
 
-    def find_object(self, queryset, request):  # noqa: C901
```
rzarrouk, line 187: Thread from slack: 

Are we fine with it removing things like this?

 def find_object(self, queryset, request):  # noqa: C901AJ  [5:00 PM]
We just removed the # noqa: C901 bit
Rachel Zarrouk  [5:00 PM]
My cyclomatic complexity fix is re one of those noqa comments too actually
[5:00 PM]I get that but do we want or have visibility on them as "issues" otherwise?
[5:00 PM]I guess Sentry does flag N+1s
[5:01 PM]Realistically I don't think it will make tonnes of difference, but just thinking
AJ  [5:01 PM]
I expanded the cyclomatic complexity to match the Engineering Handbook.

Originally it was set to 12 everywhere.

Now it's 20 in "normal" mode and 12 in strict mode.
Rachel Zarrouk  [5:01 PM]
Gotcha
AJ  [5:01 PM]
That's why they don't appear as issues any more.
[5:02 PM]Ultimately, I would like to move towards the strict checks but there's no need right now
Rachel Zarrouk  [5:18 PM]
It's doing a lot of setting empty arrays to None then checking if their None and then setting to empty arrays which is a pattern I'm not aware of.. but doesn't seem an issue

Here I always thought the former was clearer since it's explicit but it's instatiated either way so:
                - raise ValueError()
                + raise ValueError

Aside from noqa comments - I can't see anything worth flagging either, no functional difference from what I can see

rzarrouk (approved):
I'm happy to pass this as is - re. suggested improvements they can be done after the fact given there's a more imminent threat/worry re. broken pipelines.

Seems sensible to be using shared utilities where possible so I do support those changes. Unsure about rename_app.py, admittedly haven't looked into it specifically.

Qs re. cyclomatic complexity changes answered in slack - noted in a now resolved conversation.

app/core/management/commands/rename_app.py
jamesbiggs: I guess we can just write a new one if we ever need it again - but I think we are better at naming things for the future now. Happy to remove it if you are 🙂

jamesbiggs (approved):
LGTM! 🚀 I much prefer ruff to isort + black + flake8

app/core/management/commands/rename_app.py
ahosgood: Removed. We shouldn't _really_ be renaming applications because it can cause issues.

If there was a real need, we'd want to do it as a migration.

app/core/json_api_client.py
ahosgood: I will suggest this as a separate task...

PR #1974, opened 2026-05-18
Education skeleton tweaks

jamesbiggs:
I've made this as a PR into the original PR but unsure if that makes it harder to follow than merging into main. Can update if so!

This:
- cleans up the redundant prefixing (e.g. `session_`),
- adds a key details tab
- a few small tweaks for editor readability/consistency (e.g. moving `PublishedDateMixin.promote_panels` to the top of the `promote_panels`
- splits `RelatedPageLinkBase` subclasses into their own models (I like the idea, but it may be over-genericising for little benefit!)
- moves `SessionLocation.details` from a `StreamField` to individual fields. Easier to use in future when we _know_ those fields will exist (even if they're empty), and removes other general complexities of `StreamField` use vs dedicated fields
- rejigs `TimePeriod` representations so `name` is just the name (e.g. Medieval) and the to/from (or not) dates are used for a `display_name` as we have on other models (e.g. `KeyStage`) to prevent inconsistencies between dates/names (and updated 0002 migration to match!)
- moves `question` `StructBlock` into its own class (`SourceQuestionItemBlock`), which is then instantiated in `SourceQuestionBlock(StreamBlock).question`
- applies `ruff` formatting to `feature/education-skeleton-pages` changes


app/education/migrations/0002_seed_key_stage_and_taxonomies.py
```diff
@@ -32,12 +31,12 @@
...
-    ("early-twentieth-century-1901-1945", "Early Twentieth Century (1901-1945)", 1901, 1945, True),
-    ("post-war-and-modern-1945-present", "Post-War and modern (1945- present)", 1945, None, True),
+    ("early-civilisations", "Early civilisations", None, 900, False),
+    ("medieval", "Medieval", 900, 1485, True),
+    ("early-modern", "Early modern", 1485, 1750, True),
+    ("industrial-revolution", "Industrial revolution", 1750, 1901, True),
+    ("early-twentieth-century", "Early Twentieth Century", 1901, 1945, True),
+    ("post-war-and-modern", "Post-War and modern", 1945, None, True),
     ("cross-period", "Cross period", None, None, True),
 ]
```
ahosgood, line 41: _Personally_, I feel this might be too complex.

I don't _think_ we need the year from and to as separate fields, do we?

```suggestion
# Tuple: (slug, name, available_for_resources)
TIME_PERIODS = [
    ("early-civilisations", "Early civilisations (pre 900)", False),
    ("medieval", "Medieval (900-1485)", True),
    ("early-modern", "Early modern (1485-1750)", True),
    ("industrial-revolution", "Industrial revolution (1750-1901)", True),
    ("early-twentieth-century", "Early Twentieth Century (1901-1945)", True),
    ("post-war-and-modern", "Post-War and modern (1945- present)", True),
    ("cross-period", "Cross period", True),
]
```

jamesbiggs: I think we should keep it just to cover our bases for future (if we want to merge with EtC or other tags in future), we also get ordering by date and could be useful for future functionality - I'd rather put it in now rather than try to add it retrospectively later 🤔

rzarrouk: I think the idea behind this was just futureproofing, corroborated by rose+lizzy thinking in office https://github.com/nationalarchives/ds-wagtail/pull/1959#discussion_r3225539131

Edit: discussed at catch-up

ahosgood: I'm happy to follow your twos' train of thought 👍🏼

rzarrouk (approved):
All good to me - called about changes, only Q I had was re. venue details being collapsable with classname=collapsed - but is UI vs consistency so chill if not. Also can see why streamfield approach was changed for readability. 
+ thanks for catching all my typos 👀 lol

PR #1978, opened 2026-05-21
Education page amendments

rzarrouk:
## Migrations added, comments actioned :)

Lizzy's feedback here: 
https://nationalarchivesuk.slack.com/docs/T09LRLGHY/F0B64SCRWL8


app/education/models/sessions.py
```diff
@@ -293,6 +293,7 @@ def type_label(cls) -> str:
     )
 
     curriculum_connection_description = RichTextField(
+        features=["italic", "bold", "ol", "ul", "link"],
```
jamesbiggs: I think just `settings.EXPANDED_RICHTEXT_FEATURES` will do here, unless we really don't want `code` as a setting 🙂

app/education/blocks.py
```diff
@@ -80,6 +81,31 @@ class Meta:
         label = "Source question"
 
 
+class TeachersNotesBlock(blocks.StreamBlock):
+    notes = ParagraphBlock()
+    subheading = SubHeadingBlock()
+    inset_text = InsetTextBlock()
```
jamesbiggs, line 87: This might cause some problems for AJ on the frontend - the names we set here are displayed in the API - so while the frontend might be okay with `paragraph` it will get `notes`, which it has no knowledge of. I'm not sure if this needs to be `paragraph = ParagraphBlock()` or if `notes` is fine... Same goes for `sub_heading` instead of `subheading`.

jamesbiggs (commented):
This all looks good, well done @rzarrouk !! 🎉 

Just a couple of comments (and then the migrations need doing too, of course 😃 )

app/education/blocks.py
ahosgood: Ah, yes please - I have a switch statement based on the `type` and I support `type: "paragraph"` at the moment.

This doesn't change how it appears in the CMS though, right?

rzarrouk: Cool, this was done in the last couple commits :) 👍

jamesbiggs: I don't think so - I think it should still say "paragraph text" or "sub heading" etc

jamesbiggs (approved):
🚀

ahosgood (approved):

PR #1979, opened 2026-05-21
Education page API fields

rzarrouk:
## - i've also still left out the 3 primary tag fields


app/education/models/sessions.py
```diff
@@ -383,15 +383,24 @@ def clean(self):
...
+            APIField("last_published_at"),
+        ]
+        + [
+            # TODO: primary tags?
+            # APIField("key_stage", serializer=KeyStageSerializer()),
+            # APIField("time_period", serializer=TimePeriodSerializer()),
+            # APIField("theme", serializer=ThemeSerializer()),
+            APIField("key_stages", serializer=KeyStageSerializer(many=True)),
+            APIField("time_periods", serializer=TimePeriodSerializer(many=True)),
+            APIField("themes", serializer=ThemeSerializer(many=True)),
```
jamesbiggs, line 420: These may need some more thought/input from the wider team

Image

If I upload them in a specific order, should end users see them in that order? In the CMS we can use it as a weighting, but would end users understand that? Or should they be alphabetical/chronological, regardless of the weighting/ordering in the CMS... 🤔

app/education/models/sessions.py
```diff
@@ -383,15 +383,24 @@ def clean(self):
...
-        APIField("time_periods", serializer=TimePeriodSerializer(many=True)),
-        APIField("themes", serializer=ThemeSerializer(many=True)),
-    ]
+    api_fields = (
+        BasePageWithRequiredIntro.api_fields
+        + RequiredHeroImageMixin.api_fields
+        + [
+            PublishedDateMixin.get_published_date_apifield(),
+            PublishedDateMixin.get_is_newly_published_apifield(),
+            APIField("last_published_at"),
```
jamesbiggs: No need for `last_published_at` - it's already in the `meta` fields 🙂

app/education/models/sessions.py
```diff
@@ -383,15 +383,24 @@ def clean(self):
...
-        # APIField("time_period", serializer=TimePeriodSerializer()),
-        # APIField("theme", serializer=ThemeSerializer()),
-        APIField("key_stages", serializer=KeyStageSerializer(many=True)),
-        APIField("time_periods", serializer=TimePeriodSerializer(many=True)),
-        APIField("themes", serializer=ThemeSerializer(many=True)),
-    ]
+    api_fields = (
+        BasePageWithRequiredIntro.api_fields
+        + RequiredHeroImageMixin.api_fields
+        + [
```
jamesbiggs, line 392: You're missing most of the session fields here (content and details)

app/education/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from app.core.serializers import DefaultPageSerializer
+from app.core.serializers import DefaultPageSerializer, RichTextSerializer
 
 
 class _StreamFieldRepresentationMixin:
```
jamesbiggs: I had meant to review this mixin in the first PR, I like the idea but it's probably better to use `wagtail.api.v2.serializers.StreamField`. You'd end up with something like:
```py
class SourceSerializer(serializers.Serializer):
    title = serializers.CharField()
    media = StreamField()
    featured_link = StreamField()
    description = RichTextSerializer()
    question = StreamField()
```
instead. `StreamField` essentially does the same as the mixin here as well:
```py
def to_representation(self, value):
        return value.stream_block.get_api_representation(value, self.context)
```

It can also be removed from `CurriculumConnectionSerializer` as it's unused

app/education/models/resources.py
```diff
@@ -307,22 +307,32 @@ class TeachingResourcePage(
...
+    api_fields = (
+        BasePageWithRequiredIntro.api_fields
+        + RequiredHeroImageMixin.api_fields
+        + ContentWarningMixin.api_fields
+        + [
+            PublishedDateMixin.get_published_date_apifield(),
+            PublishedDateMixin.get_is_newly_published_apifield(),
+            APIField("last_published_at"),
+        ]
+        + [
```
jamesbiggs, line 319: Missing `curriculum_connections`

jamesbiggs (changes requested):
Nice work @rzarrouk - just a few bits missing and needing tweaking 🙂

app/education/models/sessions.py
rzarrouk: stashed!

app/education/models/sessions.py
jamesbiggs: Just checked in with Lizzy, these should be ordered alphabetical (themes), chronological (time periods), and numerical (key stages) 🙂

jamesbiggs (approved):
Approved but see comment r.e. ordering of tags 😃

app/education/models/sessions.py
rzarrouk: Ah okay, so we aren't doing primary tags? I should remove a couple more things if that's the case.

rzarrouk: I'll do this in the amendments branch :)) (Y) Thanks for asking 

https://github.com/nationalarchives/ds-wagtail/pull/1978/changes/1d388e3c84b200962862e99b837c03568efd93d9

PR #1982, opened 2026-05-26
Remove `MediaBlock` inheritance from Education `YouTubeBlock`

jamesbiggs:
Noticed that the fields from `MediaBlock` were being displayed on Education's `YoutubeBlockWithCaption` - this was because `YoutubeBlockWithCaption` inherits `MediaBlock`. This PR removes that so that it inherits `StructBlock` instead.


rzarrouk:
Spoke about this on slack - better to add to the amendments branch to avoid extra migrations, I've made the change here: https://github.com/nationalarchives/ds-wagtail/pull/1978/changes/d9a6383448b4fdd6f2f651c38f4ab9d69d589c0e

PR #1984, opened 2026-05-29
Add type_labels to resources and sessions.

rzarrouk:
I wasn't sure if you also wanted Education as an umbrella category?

### DIFF
--- a/app/education/models/resources.py
+++ b/app/education/models/resources.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.db import models
+from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from modelcluster.fields import ParentalKey
 from wagtail.admin.panels import (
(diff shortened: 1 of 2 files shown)


app/education/models/resources.py
```diff
@@ -171,6 +172,10 @@ class TeachingResourcePage(
 ):
     """A page to display a teaching resource"""
 
+    @cached_property
+    def type_label(cls) -> str:
+        return "Teaching Resource"
```
ahosgood:
```suggestion
        return "Teaching resource"
```

app/education/models/sessions.py
```diff
@@ -249,6 +250,10 @@ class EducationSessionPage(
 ):
     """A page to display an education session"""
 
+    @cached_property
+    def type_label(cls) -> str:
+        return "Education Session"
```
ahosgood:
```suggestion
        return "Education session"
```

ahosgood (approved):

PR #1985, opened 2026-06-01
Remove gettext_lazy

jamesbiggs:
Fix for: https://the-national-archives.sentry.io/issues/7513606707/?alert_rule_id=11003752&alert_type=issue&notification_uuid=e92a674d-17fa-4be8-a0b8-c4a7d86f69e7&project=6377027

We don't actually need `gettext_lazy` as it's only used internally. Translation isn't really a need, and to be honest, in-browser translation will do the trick if it's _really_ needed...

I think I'll still raise this with Wagtail as I feel this is breaking at Wagtail level. But this fix works for us

### DIFF
--- a/app/articles/models.py
+++ b/app/articles/models.py
@@ -125,9 +125,7 @@ class ArticleIndexPage(BasePageWithRequiredIntro):
         blank=True,
         on_delete=models.SET_NULL,
         related_name="+",
-        help_text=(
-            "Select a page to display in the featured area. This can be an Article, Focused Article or Record Article."
-        ),
+        help_text="Select a page to display in the featured area. This can be an Article, Focused Article or Record Article.",
         verbose_name="featured article",
     )
 
@@ -578,9 +576,7 @@ class RecordArticlePage(
         blank=True,
         on_delete=models.SET_NULL,
         related_name="+",
-        help_text=(
-            "Select a page to display in the featured area. This can be an Article, Focused Article or Record Article."
-        ),
+        help_text="Select a page to display in the featured area. This can be an Article, Focused Article or Record Article.",
         verbose_name="featured article",
     )
 
--- a/app/articles/models.py
+++ b/app/articles/models.py
@@ -128,7 +128,7 @@ class ArticleIndexPage(BasePageWithRequiredIntro):
         help_text=(
             "Select a page to display in the featured area. This can be an Article, Focused Article or Record Article."
         ),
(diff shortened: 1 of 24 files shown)


app/articles/models.py
```diff
@@ -126,10 +125,10 @@ class ArticleIndexPage(BasePageWithRequiredIntro):
         blank=True,
         on_delete=models.SET_NULL,
         related_name="+",
-        help_text=_(
+        help_text=(
             "Select a page to display in the featured area. This can be an Article, Focused Article or Record Article."
         ),
-        verbose_name=_("featured article"),
+        verbose_name=("featured article"),
```
ahosgood:
```suggestion
        verbose_name="featured article",
```

app/ciim/blocks.py
```diff
@@ -29,11 +28,11 @@ class Meta:
 
 
 class RecordLinkBlock(blocks.StructBlock):
-    record = RecordChooserBlock(label=_("Record"))
-    descriptive_title = blocks.CharBlock(label=_("Descriptive title"), max_length=255)
-    record_dates = blocks.CharBlock(label=_("Date(s)"), max_length=100)
+    record = RecordChooserBlock(label=("Record"))
```
ahosgood:
```suggestion
    record = RecordChooserBlock(label="Record")
```

app/articles/models.py
ahosgood: I wonder how many of these instances are still left in?

jamesbiggs: Serves me right for trying to be clever with some regex... 🥲 

Removed the all of the brackets now 🙂

app/education/models/sessions.py
```diff
@@ -323,22 +312,22 @@ def clean(self):
...
+                {"price": ("Price is required when price detail is specified.")}
             )
         if (
             self.start_date is not None
             and self.end_date is not None
             and self.start_date >= self.end_date
         ):
             raise ValidationError(
-                {"end_date": _("End date must be later than start date.")}
+                {"end_date": ("End date must be later than start date.")}
```
ahosgood:
```suggestion
                {"end_date": "End date must be later than start date."}
```

ahosgood: There may still be one or two more...

ahosgood (approved):

ahosgood (approved):

PR #1990, opened 2026-06-03
EDEV-181: Feature/update json api

rzarrouk:
https://national-archives.atlassian.net/browse/EDEV-181


app/ciim/client.py
```diff
@@ -13,16 +12,18 @@
...
-            params = {}
-        super().__init__(api_url, params=params)
-        self.add_parameter("filter", "@datatype.base:record")
+    def __init__(
+        self, api_url: str = settings.ROSETTA_API_URL, default_params: dict = None
+    ):
+        if default_params is None:
+            default_params = {}
+        super().__init__(api_url, default_params=default_params)
+        self.add_default_parameter("filter", "@datatype.base:record")
```
ahosgood: Should we do something like this?

```suggestion
    def __init__(self, default_params: dict = None):
        if default_params is None:
            default_params = {}
        api_url = settings.ROSETTA_API_URL
        if not api_url:
            raise ValueError("ROSETTA_API_URL is not defined")
        super().__init__(api_url, default_params=default_params)
        self.add_default_parameter("filter", "@datatype.base:record")
```

app/ciim/client.py
ahosgood, line 45: `self.params` doesn't exist any more

config/settings/production.py
```diff
@@ -303,7 +303,7 @@
 WAGTAILIMAGES_IMAGE_MODEL = "images.CustomImage"
 
 # CIIM API Client
-ROSETTA_API_URL = os.getenv("ROSETTA_API_URL")
+ROSETTA_API_URL = os.getenv("ROSETTA_API_URL", "")
```
ahosgood: I don't _think_ the application should work without this setting. This was a quick and dirty was to ensure it was being set as an env var. There may be a better way to declare this though...

app/ciim/client.py
rzarrouk: This is a confusing naming thing - JSONSimpleAPI still stores default_params on self.params (python-utilities/tna_utilities/api.py line 51) so it fails if this is changed: 

```
class SimpleJsonApiClient:
...
        if default_params is None:
            default_params = {}
        self.params: dict = default_params.copy()


```

app/ciim/client.py
rzarrouk: I've done this but left api_url in the init params and defaulted it to None if there's nothing there.

config/settings/production.py
rzarrouk: We could do it in settings the same way that SECRET_KEY is done but might then be tooo enforced?

ROSETTA_API_URL = os.getenv("ROSETTA_API_URL", "")
if not ROSETTA_API_URL:
    raise ImproperlyConfigured(
        "ROSETTA_API_URL environment variable must be set and cannot be empty."
    )
    
could also set the default to the rosetta test url?

app/ciim/client.py
ahosgood: Ah yeah...it's in the parent class 🤦🏼‍♂️

config/settings/production.py
ahosgood: I think that might be a good idea for these kind of variables. We know we need them to be set and there's no real default we can fall back to. 👍🏼

rzarrouk: No probs, on it 👍

rzarrouk: Added and tested, seems to work fine with the extra docker add

config/settings/test.py
```diff
@@ -29,10 +29,10 @@
     }
 }
 
-STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
-
 ROSETTA_API_URL = "http://rosetta.test/data"
 
+STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
```
jamesbiggs, line 32: What does the move do?

rzarrouk: This was unintentional

rzarrouk: changed

jamesbiggs (dismissed):
LGTM! 😃

docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
       - WAGTAILAPI_BASE_URL=http://host.docker.internal:8000
       - WAGTAILAPI_MEDIA_BASE_URL=https://localhost
       - REDIS_URL=redis://redis:6379/0
+      - ROSETTA_API_URL=http://rosetta.test/data
```
ahosgood: Would it still work if we removed this definition and added the value into the test config?

ahosgood (approved):

PR #2000, opened 2026-06-09
Simplify LinkedPageSerializer output

ahosgood:


app/education/blocks.py
```diff
@@ -18,34 +18,10 @@
...
-
-
-class SourceYouTubeWithCaptionBlock(blocks.StructBlock):
-    youtube = YouTubeBlock()
-    caption = APIRichTextBlock(features=["bold", "italic"], required=False)
-
-    class Meta:
-        label = "YouTube"
-
-
```
jamesbiggs, line 44: These were needed so each source could have a caption after it

jamesbiggs: Just seen the comment/message on Slack... Ignore :)

PR #2005, opened 2026-06-09
Feature/further education tweaks

rzarrouk:
btw for this I've left editing the validation to treat 0 as None the other way round (so someone can write 0 and still explain if they want)

related to this highlighted bit:
Screenshot 2026-06-09 at 18 41 35


app/education/models/sessions.py
```diff
@@ -311,7 +310,7 @@ def type_label(self) -> str:
     def clean(self):
         super().clean()
 
-        if self.price is not None and not self.price_detail:
+        if self.price not in (None, 0) and not self.price_detail:
```
ahosgood: Is this more Pythonic?

```suggestion
        if self.price and not self.price_detail:
```

rzarrouk: Ah yes?!

ahosgood (approved):

PR #2006, opened 2026-06-10
Rename YouTube block field to 'youtube_video' and update description section key to 'content_section'

ahosgood:


rzarrouk (approved):
I could pull this into my PR here too potentially?

https://github.com/nationalarchives/ds-wagtail/pull/2005

 Depends if it's blocking blocking you - but the changes in mine are also simple enough too

PR #2011, opened 2026-06-12
Feature/wagtail edu blog additions

rzarrouk:
Minor changes and the addition of sub sub headings to blog posts


ahosgood (changes requested):
Could you please integrate https://github.com/nationalarchives/ds-wagtail/pull/2010 into this PR?

app/blog/blocks.py
```diff
@@ -37,6 +38,7 @@ class SectionContentBlock(blocks.StreamBlock):
     quote = QuoteBlock()
     record_links = RecordLinksBlock()
     sub_heading = SubHeadingBlock()
+    sub_sub_heading = SubSubHeadingBlock()
```
jamesbiggs, line 41: You'll also need to add `sub_sub_heading = None` in `BlogPostPageStreamBlock` so that we can't have `sub_sub_headings` before the `heading` from `section` 🙂

jamesbiggs (dismissed):

app/blog/blocks.py
rzarrouk: Ah thank you :)

ahosgood (approved):

PR #2018, opened 2026-06-17
EDEV-187: Feature/video chapter hh:mm:ss

rzarrouk:
[EDEV-187]

[EDEV-187]: https://national-archives.atlassian.net/browse/EDEV-187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/media/models.py
```diff
@@ -2,28 +2,51 @@
...
+        required=True,
+        default="00:00:00",
+        label="Chapter time",
+        help_text="Enter chapter time as HH:MM:SS.",
+        validators=[
+            RegexValidator(
+                regex=r"^\d{1,3}:[0-5]\d:[0-5]\d$",
+                message="Enter chapter time as HH:MM:SS.",
+            )
+        ],
```
jamesbiggs: I entered this:

Image

And then upon saving it got translated to this:

Image

Which then fails if I try to save again as it breaks the `RegexValidator` rules.

I think the logic would probably be best in a custom subclass of `CharBlock` - something like `HourTimeBlock` (or a better name 😆). Then you can have `time = HourTimeBlock(...)` and that moves the logic to the block rather than the parent block.

I think you'll also want to use `get_form_state` to have different values in the frontend (form) and the backend (DB) of the CMS

app/media/models.py
```diff
@@ -162,6 +185,31 @@ def chapters_file_full_url(self):
...
+            sort_key = parse_chapter_time_to_seconds(time_str)
+            chapter_pairs.append(
+                (
+                    sort_key,
+                    {
+                        "time": time_str,
+                        "heading": chapter.value["heading"],
+                        "transcript": expand_db_html(
+                            chapter.value["transcript"].source
+                        ),
```
jamesbiggs: This should probably be `"time": sort_key,` as the current block returns an `int` of the chapter's seconds

I think the logic above can also be improved/simplified a bit e.g. `value = chapter.value` to prevent frequent calls to `chapter.value`

jamesbiggs (changes requested):

jamesbiggs (changes requested):
I think we shouldn't be clearing the field (resets to 0 / 00:00:00) as often editors can miss the error message and click "save" again. This would mean if I entered `00:62:5123` the CMS would throw an error and set it back to `00:00:00` which, if I save again, is valid - and would pass. Meaning my chapter I meant to set for `00:xx:yy` is now set for `00:00:00`. I think this would be best done in `to_python` and `get_form_state` 😃 

As well as this, Copilot's comment about updating `api/urls/media.py` is right - as we're saving `chapter.time` as a `str` now, when we use the `CustomMediaItemSerializer` it checks for the `chapter.value["time"]` which will now be a `str xx:yy:zz` rather than `int 1234` so just updating that `CustomMediaItemSerializer` to use `api_chapters` instead should do the trick 👌

rzarrouk:
@jamesbiggs 
Thoughts on this? A copilot suggestion - it'd silently fail but does stop bad data causing API issues
Screenshot 2026-07-17 at 11 53 30

jamesbiggs (changes requested):
The field is still being set to 00:00:00 if an editor enters an invalid value:

Image

And on the Copilot suggestion of `if sort_key is None: continue` I think I would leave that out as `time` is required so there should always be a value 🙂

jamesbiggs (approved):
I was wrong on my last comment 😆 It works! 🚀

PR #2026, opened 2026-06-30
Additional image format generation

jamesbiggs:
This PR updates the functionality of the `image_generator` helper function (and subsequent calls from serializers) to allow for `additional_rendition_specs`. This is a `dict`, such as `{"square": "fill-900x900", "wide": "fill-1200x600"}`. This will then return the original image's rendition in `jpeg`, `webp`, and any other formats (e.g. `png`) as well as the additional renditions, prefixed with the key in the dictionary.

image

https://github.com/nationalarchives/ds-frontend/pull/105


app/api/tests/expected_results/author.json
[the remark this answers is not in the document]
jamesbiggs: #wiz_ignore test fixture, not real PII

app/core/models/mixins.py
```diff
@@ -329,11 +328,6 @@ class Meta:
     ]
 
     # API field building blocks for flexible composition
-    _teaser_image_square_api_field = APIField(
-        "teaser_image_square",
```
ahosgood, line 333: We still need this image in some form - all the teaser images on https://www.nationalarchives.gov.uk/people/ are square (as well as the images on the actual profile pages).

What I mean is that we don't **_always_** need a square teaser image like I think you have done - only for people pages.

jamesbiggs: Ah okay, no worries - that's a straightforward change

Which page types need square images? (and any other formats, if at all?)

ahosgood: Just the people index and the person profile pages need square images right now.

jamesbiggs: @ahosgood we had/have square teaser images on every page and I think we still need them e.g. on education: 
image

As for the other images that have changed, I've created a list of all of the images that are affected:
```
BasePage
teaser_image_square -> teaser_image.square_jpeg/webp

Pages with HeroImage
hero_image_small -> hero_image.small_jpeg/webp

ExplorerIndexPage
stories_hero_image_small -> stories_hero_image.small_jpeg/webp

ShopCollectionBlock
background_image_small -> background_image.small_jpeg/webp

PeoplePage
image_small -> image.small_jpeg/webp

EventPage
speakers.image_small -> speaker.image.small_jpeg/webp
```

ahosgood: Those aren't square although they may look it.

They are the 3:2 images but cropped to a set width and the height just matches the height of the content. For responsive purposes, they should remain 3:2.

jamesbiggs: I've removed all "square" teaser images from every page aside from `PersonPages` 👍

ahosgood (approved):

PR #2031, opened 2026-07-09
Media page URLs

jamesbiggs:
This PR adds links in the image and media editing pages to the "live" media pages.

The base URL is set by `MEDIA_PAGE_URL`, and the individual routes are set in `context_processors.py`. TODO added for when audio is correctly displayed under `url.com/audio/` - it is currently `url.com/video/`

image

image

image

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -241,6 +241,7 @@ STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 
 MEDIA_ROOT = "/media"
 MEDIA_URL = "media/"
+MEDIA_PAGE_URL = os.getenv("MEDIA_PAGE_URL", WAGTAILADMIN_BASE_URL)
 
 WAGTAILMEDIA = {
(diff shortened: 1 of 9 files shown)


config/settings/production.py
```diff
@@ -241,6 +241,7 @@
 
 MEDIA_ROOT = "/media"
 MEDIA_URL = "media/"
+MEDIA_PAGE_URL = os.getenv("MEDIA_PAGE_URL", WAGTAILADMIN_BASE_URL)
```
ahosgood: `WAGTAILADMIN_BASE_URL` gives http://localhost:8080 - I don't think we can use this as it's the URL for Wagtail, not the frontend.

I wonder if this should fall back to `WAGTAILAPI_MEDIA_BASE_URL` instead?

ahosgood:
I think we need to document more of these env vars...

jamesbiggs:
@ahosgood added docs similar to RAMSR with the "environment variables" section 👍

ahosgood (approved):

PR #2033, opened 2026-07-13
Feature/education api regions

rzarrouk:
https://ds-education-prototype-9a0169fe9b4e.herokuapp.com/locations

- Remove custom from the list of the search_filters location in http://localhost:8000/api/v2/pages/695/
- Add another entry into search_filters of regions which returns all the regions used on published, future and current sessions
- Allow me to pass a query string of ?region=south-west&region=north-east etc.
- Follow the logic from the linked prototype at the top of this message chain

https://national-archives.atlassian.net/wiki/spaces/WM/pages/1942323242/Listing+API+and+filters

### DIFF
--- a/app/api/tests/test_education_listings.py
+++ b/app/api/tests/test_education_listings.py
@@ -174,6 +174,7 @@ class EducationListingsAPITest(WagtailPageTestCase):
         location_types,
         start_date=None,
         end_date=None,
+        region=None,
     ):
         page = EducationSessionPage(
             title=title,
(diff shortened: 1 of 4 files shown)


jamesbiggs (changes requested):
If I go to http://localhost:8000/api/v2/education/sessions/?region=south_west&region=south_east&location=national_archives I would expect to see sessions at TNA and in the South East + South West (school + custom) but currently it only shows TNA.

If I make it http://localhost:8000/api/v2/education/sessions/?region=south_west&region=south_east&location=national_archives&location=your_school then I get the South West + South East but only for "at your school" (so no custom venues)

app/api/tests/test_education_listings.py
```diff
@@ -354,6 +358,54 @@ def test_sessions_endpoint_supports_location_taxonomy_search_and_pagination(self
...
+            title="Mouse at TNA",
+            key_stages=[self.ks1],
+            time_periods=[self.early_modern],
+            themes=[self.medicine_welfare],
+            location_types=[SessionLocation.LocationType.NATIONAL_ARCHIVES],
+            start_date=future_date,
+        )
+
+        response = self.request_api(
+            "/api/v2/education/sessions/?region=north_east&region=south_west"
```
ahosgood, line 397: Also a test for this but with "At your school" selected as a location to show the custom location won't show up.

jamesbiggs (approved):
🚀

PR #2069, opened 2026-08-03
Image "alternative format"

jamesbiggs:
This adds the ability to add a file (currently limited to csv, xls and xlsx) to an image for an "alternative format" such as "transcript with tables".

There is a dropdown that will be expanded when we require more alternative versions, but for now it is just one option.


app/images/models.py
```diff
@@ -81,6 +86,23 @@ class CustomImage(ClusterableModel, AbstractImage):
         help_text="An optional English / Modern English translation of the transcription.",
     )
 
+    alternative_format_heading = models.CharField(
+        verbose_name="alternative format heading",
+        max_length=30,
+        choices=AlternativeFormatHeadingChoices.choices,
+        default=AlternativeFormatHeadingChoices.TRANSCRIPT_WITH_TABLES,
```
ahosgood: I don't know about setting a default - I would prefer that it be a conscious decision.

rzarrouk: I've made the change here, so it's a conscious choice - just doing some testing

EDIT: all tests fine 👍

rzarrouk:
@ahosgood API shape is currently the below - I've updated it so file is an object and metadata is a bit more consistent with image related files e.g. renditions: 28f6c7f5 (also added file_type which is probs overkill but seems to test fine..)

"alternative_format": {
  "heading": "Transcript with tables",
  "file": "/media/images/alternative_formats/table.csv"
}

rzarrouk (approved):
I've made a couple of changes incl. the one suggested by AJ - tempted to remove file_type from the API cause it isn't as robust a check and realistically why and what would we need it for (I added it too early this morning to be thinking as straight...) but am happy with it otherwise

rzarrouk:
Screenshot 2026-08-11 at 17 02 29

PR #2076, opened 2026-08-06
Feature/2fa recovery codes

rzarrouk:
## Lots of line changes but dw cause there are lots of tests, templates and additional logging****

I got a bit stuck here with the login part because there were issues with race conditions/concurrent requests and had to take a caching approach and then added a nonce (thanks to copilot tbh) and I'm not sure that it's alltogether a cohesive approach/design cause bits have been tacked on -but it does all work

Files to actually focus on:
- middleware.py
- admin_views.py
- wagtail_hooks.py
- list/manage_2fa_devices.py
- 2fa-management.md

What's functionally been added: 
- adding 2fa recovery codes on login
- ability to list users without 2fa/recovery codes
- ability to reset recovery codes (with an email to the user to let them know)
- lots of tests for all of it 

****Looks v long but some chunky files that are fairly simple:
- manage 2fa devices and list 2fa files are mostly logging and some refactoring to avoid loads of repetition
- lots of test files and additional tests added for emails too
- a template recreated (otp_v6) this is purely so that users read they can use recovery codes as well as one time passwords
- some refactor of auth.py email sends (all tested on dev)


app/core/forms/auth.py
```diff
@@ -51,3 +51,64 @@ def save(self, *args, **kwargs):
...
+    subject,
+    plain_tpl,
+    html_tpl=None,
+    ctx=None,
+    execute=False,
+):
+    ctx = ctx or {}
+    plain = ""
+    html = None
+    # Render plain text with Django loader to avoid Jinja parsing Django tags
```
ahosgood, line 67: What incompatibility are we dealing with here where we have to use Django over Jinja?

app/core/admin_views.py
```diff
@@ -134,3 +142,54 @@ def tree_explorer_view(request):
...
+    if used_cache and nonce:
+        try:
+            cache.delete(get_recovery_codes_cache_key(request.user, nonce))
+            logger.info(
+                "Cleared recovery codes cache for user %s and nonce %s",
+                getattr(request.user, "pk", "<unknown>"),
+                nonce,
+            )
+        except Exception as e:
+            logger.warning(
```
ahosgood:
```suggestion
            logger.error(
```

app/core/middleware.py
```diff
@@ -0,0 +1,130 @@
...
+def get_recovery_codes_cache_key(user, nonce=None):
+    if nonce:
+        return f"{RECOVERY_CODES_CACHE_NAMESPACE}:u{user.pk}:n{nonce}"
+    return f"{RECOVERY_CODES_CACHE_NAMESPACE}:u{user.pk}"
+
+
+def create_static_device_with_tokens(
+    user,
+    count=10,
+    length=10,
```
ahosgood:
```suggestion
    length=16,
```

